PR #1, opened 2019-05-17
Init gaia

alessio:


sabau:
~Have you already been able to populate CCI environment with all the proper credentials?~ Apparently yes

PR #40, opened 2019-06-10
Update to SDK's latest master

alessio:
- [ ] Wrote tests
- [ ] Added a relevant changelog entry: `clog add [section] [stanza] [message]`

______

For Admin Use:
- Reviewers Assigned

### DIFF
--- a/.pending/breaking/gaiacli/40-rest-server-s---c
+++ b/.pending/breaking/gaiacli/40-rest-server-s---c
@@ -1,0 +1,1 @@
+#40 rest-server's --cors option is now gone.
--- a/app/params.go
+++ b/app/params.go
@@ -1,0 +1,23 @@
+package app
+
+// Simulation parameter constants
+const (
+	StakePerAccount                                    = "stake_per_account"
+	InitiallyBondedValidators                          = "initially_bonded_validators"
+	OpWeightDeductFee                                  = "op_weight_deduct_fee"
+	OpWeightMsgSend                                    = "op_weight_msg_send"
+	OpWeightSingleInputMsgMultiSend                    = "op_weight_single_input_msg_multisend"
+	OpWeightMsgSetWithdrawAddress                      = "op_weight_msg_set_withdraw_address"
+	OpWeightMsgWithdrawDelegationReward                = "op_weight_msg_withdraw_delegation_reward"
+	OpWeightMsgWithdrawValidatorCommission             = "op_weight_msg_withdraw_validator_commission"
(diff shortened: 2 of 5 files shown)


alexanderbez (approved):
ACK

.pending/breaking/gaiacli/40-rest-server-s---c
```diff
@@ -0,0 +1 @@
+#40 rest-server's --cors option is now gone.
```
alexanderbez, line 1: Let's remove this entirely. This raises the question of whether or not we should duplicate change log entries (I assume the SDK already has this in its pending log).

alessio: Well, it's a change in gaiacli `rest-server` interface, isn't it? I think it's worth mentioning the removal of a flag. I suspect I may have gotten wrong the sdk's changelog entry anyway :-/

alessio: I'll rewrite SDK's pending entry once this is merged https://github.com/cosmos/cosmos-sdk/commit/8cc708b67cf8cf06484bb3571eeead0103f560a8

rigelrozanski (approved):

.pending/breaking/gaiacli/40-rest-server-s---c
alexanderbez: I don't think we should duplicate change logs...Gaia should ultimately have a pending long entry to an updated SDK version which contains all the updates. Otherwise, we duplicate everything. Also, this PR is much more than just removing a CLI flag.

alessio: The removal of the `--cors` flag is an ABI change in `gaiacli` and affect usage of the built binary. It generally seems reasonable to mention changes to a program in the changelog it is shipped with.

alexanderbez: True, but what about once the SDK is released and we have a single PR in Gaia to update the SDK version. This SDK version changes many things (CLI, binaries, APIs, etc...). Do we include all of these in Gaia's pending log too? If so, that means we're gonna have to either:

1. Copy/paste the entire SDK change log over
2. Read change-by-change and cherry-pick

I think (2) is better, but very burdensome. Do you agree? Maybe we have no choice :-/

alessio: Oh I see now - 2. is suboptimal, whilst 1. is horrible. Let's go for 2. :+1:

PR #58, opened 2019-06-27
Restore contract tests

### LINKED ISSUE #57: CI: Fix contract_tests
## Summary of Bug

The `contract_tests` CI target is currently broken. As such, it's currently disabled.

The `contract_tests` CI target executes `make setup-contract-tests-data`:

```shell
setup-contract-tests-data:
	echo 'Prepare data for the contract tests'
	rm -rf /tmp/contract_tests ; \
	mkdir /tmp/contract_tests ; \
	cp "${GOPATH}/pkg/mod/${SDK_PACK}/client/lcd/swagger-ui/swagger.yaml" /tmp/contract_tests/swagger.yaml ; \
	./build/gaiad init --home /tmp/contract_tests/.gaiad --chain-id lcd contract-tests ; \
	tar -xzf lcd_test/testdata/state.tar.gz -C /tmp/contract_tests/
```

The `tar -xzf lcd_test/testdata/state.tar.gz -C /tmp/contract_tests/` command is completely overriding the `.gaiad` directory which is wrong.

I'm not sure where `state.tar.gz` comes from or how it's configured, but we should not be dumping static configuration here.

Example with this pipeline enabled: https://circleci.com/gh/cosmos/gaia/2543. In this case, Tendermint's configuration structure has changed and the correct configuration was replaced by whatever was in `state.tar.gz` which was old/deprecated.

## Version

0432ac5

/cc @alessio @sabau 

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged

sabau:
To facilitate upgrades, instead of having to update the tests we can just replace the parts of `config.toml` that we want to change:
- replace config.toml overriding with sed on top of the original toml file

Closes: https://github.com/cosmos/gaia/issues/57

- Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/develop/CONTRIBUTING.md#pr-targeting))

- [ ] Wrote tests
- [ ] Added a relevant changelog entry: `clog add [section] [stanza] [message]`

______

For Admin Use:
- Reviewers Assigned


Makefile
```diff
@@ -175,6 +175,15 @@ setup-contract-tests-data:
 	./build/gaiad init --home /tmp/contract_tests/.gaiad --chain-id lcd contract-tests ; \
 	tar -xzf lcd_test/testdata/state.tar.gz -C /tmp/contract_tests/
 
+	# speed up tests by reducing timeouts
```
alexanderbez: How do these additions fix the original problem? Also, if this is find-and-replace, it seems pretty hacky and fragile.

alexanderbez (changes requested):

Makefile
alessio: I agree with @alexanderbez here - wouldn't it be better to just provide a `config.toml`?

alexanderbez:
Another thing I'd like to note actually. I spent about 20 minutes debugging why the aforementioned PR was failing on this CI step. As a result, I still have no idea what this CI pipeline does or what its purpose is? Can you please add some documentation to the makefile targets, the CI configuration and perhaps elsewhere too?

Also, still not clear on where `state.tar.gz` comes from??

Makefile
sabau: We had a `config.toml` inside `state.tar.gz`, I just removed it in order to avoid this problem to happen again.
The pipeline was not working because of a single value change: `db_backend` changed from `leveldb` to `goleveldb` and when the script was replacing the generated `config.toml` with the one inside `state.tar.gz` this change got overridden.

sabau:
Ok I'm working on documenting it

`state.tar.gz` was just generated locally, it contains the minimal set of info I could find in order to have predictable output, so two folders:
- `.gaiacli` addresses that I can use to submit proposals, voting, make transactions etc, and they have to be always the same in order to check automatically against `swagger.yaml`
- `.gaiad` validators and genesis info, to have a predictable output

With those, I can be sure that the outcome of the tests will have the right hashes, addresses without changing them in the swagger file. If the swagger file would have been dynamic and generated on demand all this could happen without any stored state.

>  still have no idea what this CI pipeline does or what its purpose is

The goal is to make sure what we document is what we have implemented, but to do so automatically we need to have a predictable output

alexanderbez:
Great, thanks @sabau. Still not sure on how `lcd_test/testdata/state.tar.gz` is generated. You state that you removed all unnecessary files, but is this documented anywhere? How can I verify how this file is created?

sabau:
>  Still not sure on how `lcd_test/testdata/state.tar.gz` is generated

> How can I verify how this file is created?

I can add a script on how to replicate it, it's basically a sequence of few commands to add a validator and gentxs to give stakes to some addresses. Those addresses are the ones used later to setup the state by doing a governance, deposit on it, vote, send tokens and unbond,  and what the tests will check: accounts, transactions with sender/receiver filters, governance, deposits, rewards etc will happen on top of those addresses

> You state that you removed all unnecessary files, but is this documented anywhere?

I removed only `config.toml`, I think decompressing it would be the easiest way to check what's inside, if you think it makes sense I can add the list of files i.e. in the makefile when the tar extraction happens

I guess the name `state` is misleading, I could split it into two files: `gaiacli.tar.gz` and `config.tar.gz`

alexanderbez:
Yes, this absolutely needs to be in the repo. Also, we have a `testnet` command that runs a local network. Would this be an option to use instead?

sabau:
Just pushed the script to replicate a similar state.tar.gz, although it can't be used dynamically yet, as the addresses will change and they are defined in the swagger file into `cosmos-sdk`. Once the swagger will be generated dynamically from the annotations then probably will be feasible to feed it with custom addresses etc.

> Also, we have a testnet command that runs a local network. Would this be an option to use instead?

I'm not sure it's convenient, basically it will allow me only to remove the target `run-lcd-contract-tests` but at the price of making the testnet more complex. It seems that command runs on top of this image: https://hub.docker.com/r/tendermint/gaiadnode but was updated 4 months ago last time

sabau:
Nice, while simplifying this tool I found a little discrepancy in the `supply` endpoints. 
This is should patch it, then we have to update the SDK reference in here https://github.com/cosmos/cosmos-sdk/pull/4698

@alexanderbez in the meanwhile I removed everything static from `.gaiad` now I only keep the three addresses, so just the `.gaiacli` folder comes compressed.

app/app.go
```diff
@@ -149,15 +149,20 @@ func NewGaiaApp(logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest b
 	crisisSubspace := app.paramsKeeper.Subspace(crisis.DefaultParamspace)
 
 	// account permissions
-	basicModuleAccs := []string{auth.FeeCollectorName, distr.ModuleName}
-	minterModuleAccs := []string{mint.ModuleName}
-	burnerModuleAccs := []string{staking.BondedPoolName, staking.NotBondedPoolName, gov.ModuleName}
+	maccPerms := map[string][]string{
```
sabau: Can you please double-check @fedekunze I had to change it when I updated the reference SDK master

fedekunze (approved):
ACK

fedekunze (changes requested):
After talking to @alexanderbez we decided on blocking this until we write documentation on what's the goal of the contract tests and how this operates

fedekunze:
i.e this needs an ADR

PR #141, opened 2019-09-25
Upgrade golangci-lint to latest release (1.19.1)

alessio:
- [ ] Wrote tests

______

For Admin Use:
- Reviewers Assigned


alexanderbez (changes requested):
Instead of pinning to a specific version, why not just use `go get ...@latest`?

alessio:
> Instead of pinning to a specific version, why not just use go get ...@latest?

Thanks for the question @alexanderbez, I apologise for not providing a rationale beforehand. The reason behind this approach is that `go get`ting `golangci-lint` would cause `go.{mod,sum}` to be polluted with dependencies that are not necessary to build `gaia`.

alexanderbez:
> > Instead of pinning to a specific version, why not just use go get ...@latest?
> 
> Thanks for the question @alexanderbez, I apologise for not providing a rationale beforehand. The reason behind this approach is that `go get`ting `golangci-lint` would cause `go.{mod,sum}` to be polluted with dependencies that are not necessary to build `gaia`.

Correct sir, hence for `runsim` we `(cd /tmp && go get ...)`. We should do the same here.

alessio:
I'm unsure. `$TMPDIR` is defined in Make's manual though it's on the makefile to ensure it exists beforehand. I've updated the installer script to automatically detect and download `golangci-lint`'s latest release instead. Please have a look @alexanderbez

alexanderbez (approved):
testedACK

PR #150, opened 2019-10-03
Example type dumping from amino

jackzampolin:
This is the result of a conversation with @aaronc and @hschoenburg about amino types in JS

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/viper v1.4.0
 	github.com/stretchr/testify v1.4.0
-	github.com/tendermint/go-amino v0.15.0
+	github.com/tendermint/go-amino v0.15.1-0.20191003185523-ddfc7fca5e86
 	github.com/tendermint/tendermint v0.32.5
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/viper v1.4.0
 	github.com/stretchr/testify v1.4.0
-	github.com/tendermint/go-amino v0.15.0
+	github.com/tendermint/go-amino v0.15.1-0.20191003185523-ddfc7fca5e86
```
hschoenburg, line 22: Just pointing out this is the key to accessing the typeInfo's of amino

alexanderbez: Is this officially released?

alexanderbez (changes requested):

go.mod
hschoenburg: its just a demo jacked whipped up for use internally so we can use it in building out the typescript libs

hschoenburg: https://github.com/cosmos/gaia/pull/150/files

aaronc:
Has there been any follow up on this? Is it being actively worked on somewhere or in the backlog?

PR #174, opened 2019-10-31
IBC demo fixes

cwgoes:
Updating as I test.

______

For Admin Use:
- Reviewers Assigned


cwgoes:
The proof is `nil` for some reason.

cwgoes:
Fixed that; `flagProve` needed to be true.

cwgoes:
Tried `flagProve = true`, that doesn't seem to be the problem.

cwgoes:
We cannot use queriers for any IBC state queries, since we always need proofs.

We also need to prove the thing actually in the store - binary-encoded, not JSON-encoded.

PR #280, opened 2020-02-04
Makefile: remove golangci-lint installation

erikgrinaker:
## Description

`Makefile` has a hardcoded hash for the `golangci-lint` install script, but dynamically fetches the latest version. This breaks whenever a new version is released, since the hardcoded hash no longer matches. Since the user can easily install golangci-lint using their preferred source (e.g. `brew`), let's drop the automatic installation.

______

For contributor use:

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/master/CONTRIBUTING.md#pr-targeting))
- [x] ~Linked to Github issue with discussion and accepted design OR link to spec that describes this work.~
- [x] ~Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).~
- [x] ~Wrote unit and integration [tests](https://github.com/cosmos/gaia/blob/master/CONTRIBUTING.md#testing)~
- [x] ~Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)~
- [x] ~Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).~
- [x] ~Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`~
- [x] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/master/README.md#merging-a-pr))


alexanderbez (commented):
Thanks @erikgrinaker! Indeed this is an issue and something I think I brought up with @alessio. I would still prefer to keep pinning to the latest version. Any chance we can just remove the hash check?

erikgrinaker:
I guess the idea is to avoid arbitrary code execution in case `install.sh` gets compromised. But you could argue that it's just as easy for someone to compromise the actual release binary and checksum. Your call.

alexanderbez:
Well for me personally, I keep my local golangci-lint updated to the latest version periodically and it's annoying to continue having it downgraded.

Matter of fact, I would even suggest we remove this target/step entirely. We don't use it in CI and the user can dl how they wish (e.g. brew).

erikgrinaker:
Agreed, updated the PR to remove the `golangci-lint` installation.

alexanderbez (approved):
ACK

PR #305, opened 2020-03-02
Update docs website

fadeev:
* ~Copied REST API docs from [cosmos-sdk](https://github.com/cosmos/cosmos-sdk/tree/master/client/lcd/swagger-ui) into `docs/rest` directory.~
* Updated the theme to the latest version
* Replaced Riot with Discord in the footer.

https://github.com/allinbits/design/issues/220


alessio (approved):

gamarin2:
@fadeev the `swagger.yaml` file should be taken from SDK `v0.37.x` since that's what's being used on mainnet atm

fadeev:
@gamarin2 replaced swagger.yaml with the one from [v0.37.7](https://raw.githubusercontent.com/cosmos/cosmos-sdk/release/v0.37.7/client/lcd/swagger-ui/swagger.yaml). Changed host to `api.cosmos.network`

alexanderbez (dismissed):
I would advise _against_ this approach for the following reasons:

- We now have duplicate docs and twice the maintenance burden. Not only do we have to be sure to keep the SDK REST documentation up to date, but now also Gaia.
- The next major upgrade will incorporate Protobuf and most likely gRPC which has the option for Swagger docs. This would make these changes completely obsolete.

For an immediate solution and given the above, I would highly recommend continuing to use the SDK's version of the REST documentation but point to a specific release instead of always building off of `master`.

gamarin2:
I understand your points @alexanderbez, but the question is about long-term strategy for Gaia swagger docs. 

Gaia and SDK repo are already separated, and swagger docs will diverge between the two (it is even unclear if swagger docs will remain in the SDK repo if/when modules are taken out of the repo). Does it not make more sense for the Hub's swagger doc to live with the code for the Hub, i.e. in the `gaia` repo? 

Maybe for now we can work on a script to clone `swagger.yaml` from whichever SDK branch specified in `gaia` 's `go.mod`?

alexanderbez:
> Gaia and SDK repo are already separated, and swagger docs will diverge between the two (it is even unclear if swagger docs will remain in the SDK repo if/when modules are taken out of the repo). 

Why would modules be taken out of the SDK? And when would the docs diverge? The docs would only diverge if Gaia introduces its own unique module(s). With gRPC I don't think this should be a concern. 

> Does it not make more sense for the Hub's swagger doc to live with the code for the Hub, i.e. in the gaia repo?

The problem is that's not where the code is. The code resides in the SDK. Gaia just pulls stuff together to compose an app. But again, Gaia could very well decide to have its own module(s) in addition to what the SDK defines.

gamarin2:
> Why would modules be taken out of the SDK? And when would the docs diverge? The docs would only diverge if Gaia introduces its own unique module(s). 

Well it depends wether we want to have a repo specifically for modules someday, or keep them in the SDK repo. In any case, I think as SDK adoption increases, we will have more and more modules that will be available. When we have a strategy for outside contributors to include their module in the set (whether by opening PR directly on the SDK repo or on another module-specific repo), then SDK modules and Gaia modules will diverge. I believe it's inevitable that the SDK module ecosystem will expand beyond what Gaia uses.

> The problem is that's not where the code is. The code resides in the SDK. Gaia just pulls stuff together to compose an app. But again, Gaia could very well decide to have its own module(s) in addition to what the SDK defines.

Or the SDK could have more modules than what Gaia actually uses, which is very likely. I think if we're talking about Gaia swagger docs, then they need to be built from what the code in the gaia repo actually uses.

alexanderbez:
> Well it depends wether we want to have a repo specifically for modules someday, or keep them in the SDK repo.

I've expressed my disinterest in removing the core modules from the SDK as I believe that is what a bulk of the SDK is.

> In any case, I think as SDK adoption increases, we will have more and more modules that will be available. When we have a strategy for outside contributors to include their module in the set (whether by opening PR directly on the SDK repo or on another module-specific repo), then SDK modules and Gaia modules will diverge.

Not the case for various reasons:

- We will not have modules being merged into the SDK. The SDK will only maintain a "core" set of modules -- ones the team has domain knowledge and resources to maintain.
- The entire point of the module registry (i.e. npmjs.org for go-based SDK modules) is to avoid having this module maintenance headache. Developers will have modules in their _own_ repos!
- They won't diverge as the documentation comes with the code (I think). If the SDK has X modules, and Gaia only needs Y, it'll only get Y's docs.

The point is from my original post. This would be futile and removed in a couple of weeks anyway as we drive towards protobuf-based client code. Maintaining a (duplicate) `swagger.yaml` is just not the way to go.

fadeev:
I've removed `rest` dir for now. Can we merge this, so that we have the latest version of the website online? We don't have preview set up, but it's the same version of the website as https://docs.staging-tendermint.com/master/

fadeev:
@gamarin2 @alexanderbez @hschoenburg @fedekunze can you please review this RP? Thanks!

docs/.vuepress/config.js
```diff
@@ -144,12 +141,12 @@ module.exports = {
...
-    // [
-    //   "@vuepress/google-analytics",
-    //   {
-    //     ga: "UA-51029217-12"
-    //   }
-    // ],
+    [
+      "@vuepress/google-analytics",
+      {
+        ga: "UA-51029217-12"
```
tac0turtle, line 147: is this optional for the user?

fadeev: No, it is not. Do we want to make it optional?

tac0turtle: I would prefer it, but what do others think?

tac0turtle:
>@fadeev the swagger.yaml file should be taken from SDK v0.37.x since that's what's being used on mainnet atm

I think for the next release it should be the version used on the hub, but then master should reflect the master branch of the sdk. 

Maybe next release is done with the swagger file, and then we remove it afterward, that way if the docs are built off a release then the swagger files will be there. With the next major release, there will be auto-generated swagger files so we won't need it anyways

hschoenburg (approved):
pretty minimal changes at gthis point

gamarin2 (approved):

fadeev:
* Fixed some string being treated as HTML-tags and not rendered in code blocks
* Replaced GoZ banner image

PR #336, opened 2020-03-25
Add testing dockerfile to ibc-alpha

jackzampolin:
This adds a `Dockerfile` for testing multiple chains. This is to be used in the relayer integration tests.


jackzampolin:
Block timeouts currently set at ~500ms. Could turn up more, but getting non deterministic errors like:

```
    TestBasicTransfer: relayer_test.go:58: err: verify from #17 to #18 failed: invalid header: new header has a time from the future 2020-03-26 21:57:16.711688 +0000 UTC (now: 2020-03-26 14:57:06.240194 -0700 PDT m=+8.036503258; max clock drift: 10s)
    TestBasicTransfer: relayer_test.go:62: [ibc0] too many (act(0) != exp(1)) channels returned
```

jackzampolin:
Going to go ahead and merge! Its ready to go.

PR #344, opened 2020-04-01
Query updates for IBC queriers

jackzampolin:


jackzampolin:
Blocked on https://github.com/cosmos/cosmos-sdk/pull/5901

PR #386, opened 2020-05-02
GameOfZones Phase 1: Fix for capability initialization on app restart

jackzampolin:
Update to @AdityaSripal cap init fix 

supersedes #385


fedekunze:
@jackzampolin should I close this now?

PR #408, opened 2020-05-26
docker: add user 

tac0turtle:
## Description

This PR adds a `USER` to the docker image. This is done for security reasons:

```
When a Dockerfile doesn’t specify a USER, it defaults to executing the container using 
the root user. In practice, there are very few reasons why the container should have 
root privileges. Docker defaults to running containers using the root user. When that 
namespace is then mapped to the root user in the running container, it means that the 
container potentially has root access on the Docker host. Having an application on the 
container run with the root user further broadens the attack surface and enables an 
easy path to privilege escalation if the application itself is vulnerable to exploitation.
```


alessio (approved):
Great work indeed

PR #435, opened 2020-07-06
migrating some things from master to main

tac0turtle:
part 1 of migrating things from main to master. 

the reason for these migrations is because we reverted main to latest release


Dockerfile
```diff
@@ -21,9 +21,17 @@ RUN apk add --no-cache $PACKAGES && \
 # Final image
 FROM alpine:edge
 
+ENV GAIA /gaia
+
 # Install ca-certificates
 RUN apk add --update ca-certificates
-WORKDIR /root
+
+RUN addgroup gaiauser && \
```
alessio: Why not just `gaia` instead of `gaiauser`? Not a big deal anyway - it's just conventional on UNIX systems to name daemon's system users after the daemon itself.

alessio (approved):

Dockerfile
tac0turtle: will make the change.

.github/workflows/test.yml
```diff
@@ -0,0 +1,70 @@
...
+    if: "!startsWith(github.ref, 'refs/tags/') && github.ref != 'refs/heads/master'"
+
+  test-coverage-upload:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: technote-space/get-diff-action@v1
+        with:
+          SUFFIX_FILTER: |
+            .go
```
pekis1, line 23: #478

PR #459, opened 2020-08-27
Add Crypto.com to list of supported wallets

dogemos:
Add Crypto.com to list of wallets that support Cosmos Hub


tac0turtle (approved):
YASSS

PR #465, opened 2020-08-31
Create isa-os-matrix.yml

### LINKED ISSUE #466: Binaries for every push
Because of #464 which builds a raspberry pi image for Gaia, it was necessary to build binaries with each push, so that the pi image would be several gigabytes smaller. 

These binaries are also likely to come in handy for folks who do not have a go dev env set up.  Now, in order to test your modified gaia code, all that's needed is to download one or more of the binaries produced by the matrix :).  

#465 adds matrix builds.

faddat:
so far just go versions and oses.  Soon add ARM64, x86_64, armv7h, mips, RISCV

## Description

Closes: #466 

- Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/develop/CONTRIBUTING.md#pr-targeting))

- [x] Linked to github-issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Reviewed `Files changed` in the github PR explorer


faddat (commented):
lint.yml -> golangci-lint.yml

isa-os-matrix.yml gives us binaries with every push onto every branch that can be used to feed `gaia-pi`

faddat:
FYI: 

If maintainers think we need a matrix build that provides binaries for every push (I could add tests, too) -- then this is probably great for that.  

However, I am changing the Pi image builder so that it builds only ARM64 binaries and then copies them into the image.  

So basically please let me know if  we:

* want matrix builds as seen here
* want tests added to the matrix
* want to drop this PR

👍 
Cheers!

faddat:
This was needlessly broad.  I am moving development of the pi image builder out of gaia so that it can include a matrix that builds a number of production cosmos-sdk chains.

PR #525, opened 2021-01-09
Update README for v3.0.0

jackzampolin:


LICENSE
```diff
@@ -2,7 +2,7 @@ Cosmos-SDK
 License: Apache2.0
 
                                  Apache License
-                           Version 2.0, January 2004
+                           Version 2.0, January 2020
```
shahankhatch: This year should not change, since the year is based on the official approval date by the ASF.

alessio: This text should be the same as https://www.apache.org/licenses/LICENSE-2.0

```suggestion
                           Version 2.0, January 2004
```

shahankhatch (approved):
lgtm

alessio (approved):
Changes to LICENSE should be reverted. Otherwise it looks good!

LICENSE
jackzampolin: cool

PR #597, opened 2021-01-28
Remove circleci

### LINKED ISSUE #593: Prevent Circle CI actions from taking place on update of md files
## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue

shahankhatch:
Closes: #593


fedekunze (approved):
utACK

fedekunze:
actually I think the `localnet` check needs to be migrated to GH actions

alessio:
@fedekunze is right, we should add something like the following:

```yaml
  liveness-test:
    runs-on: ubuntu-latest
    timeout-minutes: 10
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-go@v2.1.3
        with:
          go-version: 1.15
      - uses: technote-space/get-diff-action@v4
        id: git_diff
        with:
          PATTERNS: |
            **/**.go
            go.mod
            go.sum
      - name: start localnet
        run: |
          make clean build-simd-linux localnet-start
        if: env.GIT_DIFF
      - name: test liveness
        run: |
          ./contrib/localnet_liveness.sh 100 5 50 localhost
        if: env.GIT_DIFF
```

I invoke our Github Actions Overlord @marbar3778 :pray:

.github/workflows/sims.yml
```diff
@@ -121,3 +121,32 @@ jobs:
...
+  test-liveness:
+    runs-on: ubuntu-latest
+    needs: Build
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2.1.3
+        with:
+          go-version: 1.15
+#      - uses: technote-space/get-diff-action@v4
```
tac0turtle: should this be commented?

tac0turtle (approved):

.github/workflows/sims.yml
shahankhatch: I believe so, the localnet action didn't execute when it was there. I believe it was there in order to cache the simulation binary, which doesn't seem to apply for the localnet sim. This is a good reminder though to re-enable the file filter before committing this PR.

shahankhatch: No, it shouldnt be. I had commented it just to trigger the action and test. Since it appears it's fine I will uncomment, commit, and merge the PR.
Thanks @marbar3778

PR #834, opened 2021-04-27
Prepare gaia alpha release for cosmoshub-4.6 upgrade

### LINKED ISSUE #829: Create branch that imports closest SDK v0.43 branch/tag
## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue

shahankhatch:
Closes #829 

This branch is for the DEX and the upcoming Hub upgrade

Updates deps to Cosmos SDK v0.43-alpha1 and ibc-go v1.0.0-alpha1


shahankhatch:
I have two places in the code commented with `// TODO: cosmoshub-4.6` where I'm not sure if the right API was used to update breaking changes.

Note that these breaking changes don't look like they were included in the SDK changelog.

app/pubkey_replacement.go
```diff
@@ -68,7 +68,9 @@ func loadKeydataFromFile(clientCtx client.Context, replacementrJSON string, genD
 
 			toReplaceValConsAddress, _ := val.GetConsAddr()
 
-			consPubKey, err := sdk.GetPubKeyFromBech32(sdk.Bech32PubKeyTypeConsPub, replacement.ConsensusPubkey)
+			// TODO: cosmoshub-4.6
```
shahankhatch, line 71: Disable linter for use of deprecated api

zmanian:
Wouldn't this be gaia 5.0?

colin-axner:
There's a regression in this branch when compared against v4.2.1. The `--home` flag doesn't work. It only partially applies the new home directory, so some files are placed in the right location and others are placed in `.gaia`. Came across this bug since relayers use this functionality in there scripts which break with this branch

Verified by installing this and v4.2.1 and running `gaia init --home <different director>`

Haven't looked into where the bug is exactly

Quickly looking at the diffs, looks like the issue should be in the SDK, but I was unable to reproduce using simd from v0.43.0-alpha1

zmanian:
@colin-axner I have a lead on the root cause for this bug. I ran into it too. It's missing changes in `root.go`. I'll try have a patch soon after i have get out of back to back calls.

zmanian:
Here is a fix for the home flag issue.
https://github.com/cosmos/gaia/pull/852

shahankhatch:
Seems like non-deterministic sim is failing with 

```
panic: type_url /cosmos.authz.v1beta1.Msg/GrantAuthorization has not been registered yet. Before calling RegisterService, you must register all interfaces by calling the `RegisterInterfaces` method on module.BasicManager. Each module should call `msgservice.RegisterMsgServiceDesc` inside its `RegisterInterfaces` method with the `_Msg_serviceDesc` generated by proto-gen [recovered]
	panic: type_url /cosmos.authz.v1beta1.Msg/GrantAuthorization has not been registered yet. Before calling RegisterService, you must register all interfaces by calling the `RegisterInterfaces` method on module.BasicManager. Each module should call `msgservice.RegisterMsgServiceDesc` inside its `RegisterInterfaces` method with the `_Msg_serviceDesc` generated by proto-gen
```

I will look into this.

colin-axner:
> Seems like non-deterministic sim is failing with
> 
> ```
> panic: type_url /cosmos.authz.v1beta1.Msg/GrantAuthorization has not been registered yet. Before calling RegisterService, you must register all interfaces by calling the `RegisterInterfaces` method on module.BasicManager. Each module should call `msgservice.RegisterMsgServiceDesc` inside its `RegisterInterfaces` method with the `_Msg_serviceDesc` generated by proto-gen [recovered]
> 	panic: type_url /cosmos.authz.v1beta1.Msg/GrantAuthorization has not been registered yet. Before calling RegisterService, you must register all interfaces by calling the `RegisterInterfaces` method on module.BasicManager. Each module should call `msgservice.RegisterMsgServiceDesc` inside its `RegisterInterfaces` method with the `_Msg_serviceDesc` generated by proto-gen
> ```
> 
> I will look into this.

Maybe try bumping to beta1? I'd expect this to be an error in the SDK, but a lot has changed since their alpha1 release

shahankhatch:
Thanks @colin-axner I suspect it's because feegrant module wasn't included in sim setup within the gaia app. I added it so hopefully that'll do the trick.

shahankhatch:
Apart from adding the code that was missing, I updated to sdk beta1 and ibc-go alpha2, and the sims seem to be running. Thanks @colin-axner for the suggestion!

app/app.go
```diff
@@ -277,7 +291,7 @@ func NewGaiaApp(
 		AddRoute(paramproposal.RouterKey, params.NewParamChangeProposalHandler(app.ParamsKeeper)).
 		AddRoute(distrtypes.RouterKey, distr.NewCommunityPoolSpendProposalHandler(app.DistrKeeper)).
 		AddRoute(upgradetypes.RouterKey, upgrade.NewSoftwareUpgradeProposalHandler(app.UpgradeKeeper)).
-		AddRoute(ibchost.RouterKey, ibcclient.NewClientUpdateProposalHandler(app.IBCKeeper.ClientKeeper))
+		AddRoute(ibchost.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper))
```
colin-axner: This needs to be `ibcclienttypes.RouterKey` not `ibchost.RouterKey`. See the NOTE in the [migration docs](https://github.com/cosmos/ibc-go/blob/main/docs/migrations/ibc-migration-043.md#proposal-handler-registration)

shahankhatch: Thanks! updated and followed the migration guide for to integrate upgrade proposals.

colin-axner:
The migrate command should be removed or updated. I'm in favor of updating it

I view the plan for this upgrade as the following:
- Plan A - x/upgrade + upgrade migrations (probably via cosmovisor)
- Plan B - validators use flag `unsafe-skip-upgrades`
- Plan C - before Plan A, validators would have exported genesis. If > 1/3 validators accidentally get corrupt databases due to a failure in Plan A, Plan B may no longer be viable. In such a situation, the chain should restart via the exported genesis

My main concern is validators not doing a back up before upgrade migrations and upgrade migrations failing for some unforeseen reason . Does cosmovisor do backups before restarting? I think Plan C is super rare, but if Plan A and Plan B fail, it'd be a real mess to get the network back up

shahankhatch:
Closing this as the liquidity module has been merged into master through another branch #858

colin-axner:
this branch was unrelated to the liquidity module and was preparation for SDK v0.43. Is there a new branch that will be opened?

PR #861, opened 2021-05-21
develop simple tooling for local testnets without virtualization

### LINKED ISSUE #860: Develop simple tooling for local testnets without virtualization

shahankhatch:
Closes #860


alessio (approved):

denniswon:
@shahankhatch The testnet is failing at the initial peer connection step ` DBG no available peers to send a PEX request to (retrying)` when running the script on macos. Do you have an idea what would be the cause? thanks

PR #869, opened 2021-06-01
Create interchain security lightpaper DRAFT

### LINKED ISSUE #659: RFC Interchain Staking Light Paper
## Summary
Create a RFC for Cross Chain Validation to track discussion that should result in an eventual ADR

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue

okwme:
Closes: #659

## Description

This is a draft paper describing Interchain Security and Cross Chain Validation.

It's unclear if this is the right location for this document. A shorter and higher level document is needed to function as more of a summary in order to get early user feedback. That could double as marketing material in a blog post.

This document is meant to act as high level in that it doesn't go into implementation detail but describes enough of each moving part so that the entire configuration should be understandable.

It is still in draft stage so help with edits / wording is extremely encouraged.

______

For contributor use:

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/main/README.md#merging-a-pr))


docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+At a very high level, Interchain Staking is the ability for staking tokens that have been delegated to validators on a parent chain to inform the composition of a validator set on a child chain. Inter-Blockchain Communication is utilized to relay updates of validator stake delegations from the parent chain to the child chain so that the child chain will have an up-to-date model of which validators can produce blocks on the child chain. The inclusion of parent chain validators is opt-in. The parent chain will honor any proof of validator misbehaviour produced by the child chain as evidence that results in slashing the stake of misbehaving validators on the parent chain. In this way the security gained from the value of the stake locked on the parent chain will be shared with the child chain.
```
AdityaSripal: Is opt-in a necessary MVP requirement?

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+The Cost of Corruption calculation doesn't account for the availability of any specific token but it does give a very rough estimate for how secure a chain is. It's important that the total value locked (TVL) on a chain remains less than the Cost of Corruption, otherwise the chain should be considered insecure. Since the ability of a chain to serve a valuable purpose is often dependent on the TVL it can handle, it's important to find ways to increase the Cost of Corruption for chains in the Cosmos ecosystem. One method of doing this is to lend the value of the Cosmos ATOM to the Cost of Corruption of any chain. This becomes possible with Interchain Staking as the ATOM, which already has a sizeable market cap, can increase the Cost of Corruption a new chain.
```
AdityaSripal:
```suggestion
The Cost of Corruption calculation doesn't account for the availability of any specific token but it does give a very rough estimate for how secure a chain is. It's important that the total value locked (TVL) on a chain remains less than the Cost of Corruption, otherwise the chain should be considered insecure. Since the ability of a chain to serve a valuable purpose is often dependent on the TVL it can handle, it's important to find ways to increase the Cost of Corruption for chains in the Cosmos ecosystem. One method of doing this is to lend the value of the Cosmos ATOM to the Cost of Corruption of any chain. This becomes possible with Interchain Staking as the ATOM, which already has a sizeable market cap, can increase the Cost of Corruption of a new chain.
```

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Regardless of the exact reason, one must assume that each validator has considered the work it takes to run a validator for the child chain in order to produce blocks on that network, and considered the risk of having their parent chain stake (ATOMs) slashed should they do a poor job validating on that network. In order to opt-in, the validator must use their validator key to submit a transaction to the Chain Registry with the intention of being included in a set of validators of a relevant child chain.
+
+The final piece of information relevant to the Chain Registry is the Time Til Launch (TTL) for each child chain. The TTL designates the point of no return for Validators to join or leave the Chain Registry. At that point the list of validators contained within the registry for a specific network can be exported to an initial genesis file for that network. If this network is transitioning to use Interchain Staking from a sovereign validator set, this will be seen as a network upgrade where the validator set gets completely redefined.
```
AdityaSripal: Is this going to be a predefined time, will it be triggerable by some third party? Who sets the TTL, is it modifiable?

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Interchain Staking should increase the diversity of the validator ecosystem by lowering the barrier to running a profitable validator business. This will go far in creating a healthy ecosystem of diverse validators that will result in anti-fragile and robustly operated networks. In order to make it possible for the top 125 validators to remain eligible as block producers for the Cosmos Hub while increasing the number of eligible validators for child chains, the Staking Module needs to stop forcing validators to undelegate when they leave the top set of 125 validators. This will result in a longer list of validators with ATOM delegations that are not participating in block production on the parent chain (Cosmos Hub). These extra validators will however be eligible to produce blocks on child chains and use their delegated ATOMs to earn rewards on the child chains as well as risk their parent chain ATOMs to slashable events should they misbehave on child chains.
```
AdityaSripal: Should they still undelegate if they are validating 0 child chains

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+The optionality around complex delegations would eventually increase the possible diversity of validator operations. However due to the complexity of such delegations, this functionality is assumed to be part of a V2 of Interchain Staking.
+
+
+### Epoch Staking
+
+The current Staking Module of the Cosmos SDK is moving towards Epoch based staking. This means that instead of validator set delegation amounts being calculated on a per block basis, they will be calculated over some group of time (or blocks) called an Epoch. This will decrease the number of times staking is calculated and generally decrease the complexity involved in staking. The additional complexity of chain relevant stake calculations will similarly benefit from a general simplification of stake calculations.
```
AdityaSripal: It will also require less packets to be sent between the chains which is a big bonus

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+The rate at which these validator set updates are relayed is a function of safety. At one extreme you could imagine the validator set being collated and relayed with every single block that is produced on the parent chain. This would ensure that the child chain has an absolutely accurate representation of validator weights at every potential state update within the parent chain. However, since delegations are subject to unbonding periods it is possible to approach state updates more conservatively. At another extreme, one may reason that if there is no active stake unbonding happening on the parent chain, it may be assumed that a validator set update will not be possible within a maximum duration of that unbonding period and so a validator set update can wait until just before that moment. Based on the possibility of instant redelegations this assumption may need to be further adjusted.
```
AdityaSripal:
```suggestion
The rate at which these validator set updates are relayed is a function of safety. At one extreme you could imagine the validator set being collated and relayed with every single epoch that is produced on the parent chain. This would ensure that the child chain has an absolutely accurate representation of validator weights at every potential state update within the parent chain. However, since delegations are subject to unbonding periods it is possible to approach state updates more conservatively. At another extreme, one may reason that if there is no active stake unbonding happening on the parent chain, it may be assumed that a validator set update will not be possible within a maximum duration of that unbonding period and so a validator set update can wait until just before that moment. Based on the possibility of instant redelegations this assumption may need to be further adjusted. 
```

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Either way comes with the question of whether it should be the parent chain that is verifying the evidence of slashable events, or simply trusting the child chain's commands to slash validators at the level of the parent. While there may be the ability to both, trusting the child chain makes the job of the parent chain much easier. In that scenario the parent chain doesn't need to have knowledge of the logic contained within the child chain that may be necessary for determining whether a slashable offense took place. Trusting the child chain to enforce its own rules puts a lot of trust in the child chain to not abuse the position and submit a wave of slashable commands that may reduce a validator to nothing. However this is ultimately the responsibility of the validator to ensure that the slashable risk of validating on a child chain is worth the expected rewards. This includes determining whether the state machine of the child chain includes logic that the validator is comfortable with.
```
AdityaSripal: Note this means that a validator's stake is slashable by the weakest validator set it participates in. This is an important point to discuss and decide on as it really changes the calculus for validators on whether they want to participate in Interchain staking

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Limits on a child chain's ability to slash a parent chain validator may also be imposed at the parent chain level. For instance it is possible to make it a required parameter of a child chain within the Chain Registry to provide a maximum slashing rate. This would ensure that a validator knows even if they violate all the rules of a child chain, they can still limit the damage to their stake by some amount. It is also possible to enforce a parameter at the level of the parent chain that prevents a validator from joining any combination of child chain validator sets that results in a combined slashable rate over some threshold. For instance if there are 3 child chains that each have maximum slashable rates of 33% over X blocks, and the parent chain has a safety limit for validators that require them to stay below a total of 90% slashable events over X blocks, no validator would be allowed to join the Chain Registry for all 3 of those child chains. This parameter on the parent chain should be a governance parameter that can be adjusted by token staked voting to reflect the risk appetite of the actual validator set of the parent chain.
```
AdityaSripal: Mitigation helps, though its important to note that it's not just the validator violating rules. A 2/3 cabal of validators on child chain can arbitrarily slash any other child chains validator's parent stake up to the maximum slashable amount

AdityaSripal (commented):
Great work! Here are my initial comments

docs/interchain-staking.md
milosevic: Isn't TVl always smaller than Cost of Corruption as the latter is the fraction of the former?

okwme: TVL can include external value as well. For instance the Gravity DEX may have more value in the tokens inside liquidity pools than the market cap of the ATOM. That would be bad because the cost of attacking the network (acquiring the ATOMs) may be less than the amount of tokens that could be stolen from the liquidity pools.

docs/interchain-staking.md
okwme: This was written before our discussion last week about a V0 MVP. I have some thoughts on this that I'd like to discuss today which woudl warrant a larger revision of the paper with more explicit V0, V1 and V2 distinctions

docs/interchain-staking.md
okwme: these are all similar questions to more general registry management. I'd prefer to keep those details vague here and link to a more detailed description of the registry from @gamarin2 when it's ready.

docs/interchain-staking.md
okwme: I think no, it will take time to accumulate enough delegators to operate in different capacities—on the cosmos hub itself, or as a potential validator on a child chain. If a validator is able to get delegations without yet producing blocks I think it would be a testament to their potential rather than a technical mistake that should be corrected. Certainly if they never produce blocks on any networks the delegators should consider redelegating.

docs/interchain-staking.md
okwme:
```suggestion
The current Staking Module of the Cosmos SDK is moving towards Epoch based staking. This means that instead of validator set delegation amounts being calculated on a per block basis, they will be calculated over some group of time (or blocks) called an Epoch. This will decrease the number of times staking is calculated and generally decrease the complexity involved in staking. The additional complexity of chain relevant stake calculations will similarly benefit from a general simplification of stake calculations, and it will require less packets to be sent between the chains.
```

docs/interchain-staking.md
milosevic: Note that it is not only about trusting state machine, as evidence can be blocked at the tendermint level (by censorship). It might be very hard for validator to compute the risk just by looking at validator set.

okwme: this has always been my understanding altho putting limits to the amount a validator can be slashed or forcing them into jail could be remedies for runaway slashing events?

docs/interchain-staking.md
okwme: very interesting....
this attack would be one child chain targeting another (possibly competitive) child chain by punishing its own shared validators? pretty nasty behaviour, i am inclined to ask why that validator got involved in such sketchy chains in the first place but I guess it could be a lot less about the chain than the other validators...

this would be more likely on child chains that have low cost of corruption due to low participation from parent validator sets? Even one or a small number of large validators could join the child chain just to get rid of general validator competition.

this makes me more inclined to prioritize the V0 version of full validator sets only...

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+The Cost of Corruption calculation doesn't account for the availability of any specific token but it does give a very rough estimate for how secure a chain is. It's important that the total value locked (TVL) on a chain remains less than the Cost of Corruption, otherwise the chain should be considered insecure. Since the ability of a chain to serve a valuable purpose is often dependent on the TVL it can handle, it's important to find ways to increase the Cost of Corruption for chains in the Cosmos ecosystem. One method of doing this is to lend the value of the Cosmos ATOM to the Cost of Corruption of any chain. This becomes possible with Interchain Staking as the ATOM, which already has a sizeable market cap, can increase the Cost of Corruption of a new chain.
```
milosevic: Total value locked term could be confusing as locking a value has different meaning in Tendermint. Maybe total value secured is more appropriate.

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Instead of distributing rewards on the child chain, the rewards collected for each block produced should be transferred to the parent chain at some interval. It may be a regular interval or a user initiated IBC packet, but would be able be similar to an ics-20 token transfer from the distribution module account of the child chain to the distribution module account of the parent chain. The difference between a standard ics-20 transfer being that the parent chain distribution module account needs to know which chain the rewards came from and over which period they were collected. This information will be necessary to perform the parent chain distribution module's responsibility of distributed rewards to delegators and validators on a per chain basis informed by the Chain Registry.
```
crodriguezvega: Typo: _... but would `be able` be similar to an ics-20 token ..._

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+
+# Conclusion
+
+Interchian staking is a new shared security primitive that has implications for the security and scalability of single blockchains like the Cosmos Hub as well as the potential to dramatically lower the barrier to running secure public blockchains for new applications. It could be thought of as a competitive configuration to sharding and put the Cosmos Hub on par with Eth 2.0 or Polkadot in terms of their security offerings to applications included in their environment. The design philosophy around the Cosmos ecosystem has always prioritized autonomy and sovereignty over guarantees around security—"Bring your own security" is a term that has been used in the past. The design of Interchain Staking discussed here tries to incorporate the Cosmos design philosophy of autonomy and sovereignty with the offering of shared security. Even within the balance between those considerations there is a spectrum of possibilities and it might be the case that multiple versions and configurations are implemented in parallel in order to satisfy as many needs as possible.
```
crodriguezvega: Typo: _`Interchian` staking ..._

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Initially it is up to validators to evaluate whether participation in a child chain validator set is worth the extra work and risk it imposes. Subsequently it is up to a delegator to decide whether the choices of their validator are aligned with their own. However it may be the case that a delegator agrees with how a validator operates on the parent chain and wishes to stay delegated ther, but disagrees with the choice to participate (or not) on a child chain.
```
crodriguezvega: Typo: _... stay delegated `ther`, but disagrees ..._

docs/interchain-staking.md
```diff
@@ -0,0 +1,178 @@
...
+
+Consideration should be made to the incentives around submitting evidence in order to ensure that punishable offenses do not go unnoticed. This is a similar issue to Relayer Incentivization, where currently it costs some fee to relay IBC packets but there is no reward or payment possible as part of the core IBC logic. As a result IBC packets are currently relayed in an altruistic manner. It's important to ensure that for integrity of the operation of the child and parent chain that slashable offenses are always submitted. It may be that the slashable amounts of tokens are used as rewards for submitting the evidence (assuming it's possible to ensure it is not the culprit submitting their own evidence and regaining their stake). Maybe some flat fee to at least pay for the transactions are required, although this may become redundant with the addition of any Relayer incentivizations into core IBC.
```
crodriguezvega: Maybe this is already covered in relayer incentivization, but I guess there should also be some incentive for relayers to relay the messages that carry the validator set updates from the parent to the child chain, right?

milosevic (approved):

docs/interchain-staking.md
okwme:
```suggestion
At a very high level, Interchain Staking is the ability for staking tokens that have been delegated to validators on a parent chain to inform the composition of a validator set on a child chain. Inter-Blockchain Communication is utilized to relay updates of validator stake delegations from the parent chain to the child chain so that the child chain will have an up-to-date model of which validators can produce blocks on the child chain. The inclusion of parent chain validators can be mandatory or opt-in depending on the requirements of the child chain. The parent chain will honor any proof of validator misbehavior produced by the child chain as evidence that results in slashing the stake of misbehaving validators on the parent chain. In this way the security gained from the value of the stake locked on the parent chain will be shared with the child chain.
```

PR #942, opened 2021-08-16
Bump Cosmos SDK and ibc-go

alexanderbez:
- Add `x/feegrant` and `x/authz` modules
- Bump SDK version to v0.44.0
- Bump ibc-go version to v1.1.0

TODO:

- [x] Update module path to `v6`
- [x] Verify the upgrade name: https://github.com/cosmos/gaia/pull/942/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R546
- [x] Verify the new IBC params: https://github.com/cosmos/gaia/pull/942/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7R549
- [x] Change `github.com/tendermint/liquidity v1.2.6-0.20210513094606-6cd272e3814d` to the upcoming release version.
- [x] change dependency ibc-go from v1.0.0 to v1.1.0
______

For contributor use:

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/main/README.md#merging-a-pr))


cmd/gaiad/cmd/root.go
```diff
@@ -53,8 +56,11 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 		Use:   "gaiad",
 		Short: "Stargate Cosmos Hub App",
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
-			initClientCtx = client.ReadHomeFlag(initClientCtx, cmd)
+			// set the default command outputs
+			cmd.SetOut(cmd.OutOrStdout())
+			cmd.SetErr(cmd.ErrOrStderr())
```
alexanderbez: I don't think we need this?

cmd/gaiad/cmd/root.go
```diff
@@ -64,7 +70,7 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 				return err
 			}
 
-			return server.InterceptConfigsPreRunHandler(cmd)
+			return server.InterceptConfigsPreRunHandler(cmd,"", appconfig.AppConfig{})  //todo: check if right
```
alexanderbez: This is correct. Gaia doesn't have a custom config.
```suggestion
			return server.InterceptConfigsPreRunHandler(cmd, "", nil)
```

cmd/gaiad/cmd/root.go
```diff
@@ -74,10 +80,10 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 }
 
 func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
-	authclient.Codec = encodingConfig.Marshaler
+	//authclient.Codec = encodingConfig.Marshaler
```
alexanderbez: remove :)
```suggestion
```

cmd/gaiad/cmd/root.go
```diff
@@ -122,6 +128,8 @@ func queryCommand() *cobra.Command {
 		rpc.BlockCommand(),
 		authcmd.QueryTxsByEventsCmd(),
 		authcmd.QueryTxCmd(),
+		feegrantcli.GetQueryCmd(),
+		authzcli.GetQueryCmd(),
```
alexanderbez: Do we need to explicitly add these? The module manager in app.go takes care of this for us.

cmd/gaiad/cmd/root.go
```diff
@@ -149,6 +157,9 @@ func txCommand() *cobra.Command {
 		authcmd.GetBroadcastCommand(),
 		authcmd.GetEncodeCommand(),
 		authcmd.GetDecodeCommand(),
+		vestingcli.GetTxCmd(),
+		feegrantcli.GetTxCmd(),
+		authzcli.GetTxCmd(),
```
alexanderbez: ditto. I think we don't need to call these.

cmd/gaiad/cmd/root.go
okwme: let's remove these @yaruwangway

cmd/gaiad/cmd/root.go
okwme: let's remove these too @yaruwangway

cmd/gaiad/cmd/root.go
okwme: let's remove these @yaruwangway

cmd/gaiad/cmd/root.go
```diff
@@ -205,22 +225,43 @@ func newApp(logger log.Logger, db dbm.DB, traceStore io.Writer, appOpts serverty
...
-	encCfg.Marshaler = codec.NewProtoCodec(encCfg.InterfaceRegistry)
-	var gaiaApp *gaia.GaiaApp
-	if height != -1 {
-		gaiaApp = gaia.NewGaiaApp(logger, db, traceStore, false, map[int64]bool{}, "", cast.ToUint(appOpts.Get(server.FlagInvCheckPeriod)), encCfg, appOpts)
+	var loadLatest bool
+	if height == -1 {
+		loadLatest = true
+	}
+
+	umeeApp := gaia.NewGaiaApp(
```
okwme: did this come from the umee code @alexanderbez ?
we should use `gaiaApp` instead of `umeeApp` @yaruwangway

go.mod
```diff
@@ -1,16 +1,18 @@
...
+	github.com/cosmos/cosmos-sdk v0.43.0
+	github.com/cosmos/ibc-go v1.0.0
 	github.com/gorilla/mux v1.8.0
-	github.com/gravity-devs/liquidity v1.2.9
 	github.com/pkg/errors v0.9.1
 	github.com/rakyll/statik v0.1.7
 	github.com/spf13/cast v1.3.1
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.7.0
+	github.com/tendermint/liquidity v1.2.6-0.20210513094606-6cd272e3814d
```
okwme: Hyung thinks they should be able to have a new release for us by the end of the week.

okwme:
https://github.com/cosmos/ibc-go/releases/tag/v1.0.1

colin-axner (commented):
first pass. The IBC changes looked good to me

dongsam:
~Hi, Liquidity version [v1.3.0](https://github.com/Gravity-Devs/liquidity/releases/tag/v1.3.0) based on sdk v0.43.0 has been released, 
and a separate PR https://github.com/cosmos/gaia/pull/956 is ready to replace liquidity version to v1.3.0 for branch [start-upgrade](https://github.com/cosmos/gaia/tree/start-upgrade) of this PR https://github.com/cosmos/gaia/pull/942~

Since the Cosmos-sdk version of this PR was updated based on v0.44.0, the liquidity was also bumped to v1.4.0 release based on sdk v0.44.0 on #956 PR
As was the latest version with Gaia, import to `gravity-devs/liquidity`, not to `tendermint/liquidity`.

app/app.go
```diff
@@ -115,11 +122,18 @@ var (
...
+			upgradeclient.ProposalHandler,
+			upgradeclient.CancelProposalHandler,
+			ibcclientclient.UpdateClientProposalHandler,
+			ibcclientclient.UpgradeProposalHandler,
 		),
 		params.AppModuleBasic{},
 		crisis.AppModuleBasic{},
 		slashing.AppModuleBasic{},
+		feegrantmodule.AppModuleBasic{},
+		authzmodule.AppModuleBasic{},
```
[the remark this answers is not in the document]
okwme: yup! i think authz is especially important while feegrant is a nice to have. They will be explicitly outlined in the proposal as added features. Would def be interested to hear about any reason you think there is to not include them tho!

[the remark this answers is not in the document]
colin-axner: I support adding them! :+1: Good to ensure someone from the SDK reviews their integration though

app/ante_handler.go
```diff
@@ -0,0 +1,64 @@
...
+        authante.NewValidateMemoDecorator(options.AccountKeeper),
+        authante.NewConsumeGasForTxSizeDecorator(options.AccountKeeper),
+        authante.NewDeductFeeDecorator(options.AccountKeeper, options.BankKeeper, options.FeegrantKeeper),
+        authante.NewSetPubKeyDecorator(options.AccountKeeper), // SetPubKeyDecorator must be called before all signature verification decorators
+        authante.NewValidateSigCountDecorator(options.AccountKeeper),
+        authante.NewSigGasConsumeDecorator(options.AccountKeeper, sigGasConsumer),
+        authante.NewSigVerificationDecorator(options.AccountKeeper, options.SignModeHandler),
+        authante.NewIncrementSequenceDecorator(options.AccountKeeper),
+        // todo check ibcante is at the right order
+        ibcante.NewAnteDecorator(options.Channelkeeper),
```
colin-axner: This looks correct to me. @AdityaSripal ?

app/pubkey_replacement.go
```diff
@@ -69,7 +69,7 @@ func loadKeydataFromFile(clientCtx client.Context, replacementrJSON string, genD
 
 			toReplaceValConsAddress, _ := val.GetConsAddr()
 
-			consPubKey, err := sdk.GetPubKeyFromBech32(sdk.Bech32PubKeyTypeConsPub, replacement.ConsensusPubkey)
+			consPubKey, err := legacybech32.UnmarshalPubKey(legacybech32.ConsPK, replacement.ConsensusPubkey)  //todo check here
```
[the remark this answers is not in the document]
alexanderbez: https://github.com/cosmos/cosmos-sdk/pull/9827

app/ante_handler.go
```diff
@@ -0,0 +1,64 @@
+package gaia
+
+import (
```
alexanderbez, line 3: Why do we have/need a custom ante-handler file/setup?

app/app.go
```diff
@@ -407,32 +530,65 @@ func NewGaiaApp(
...
-	app.UpgradeKeeper.SetUpgradeHandler("Gravity-DEX",
-		func(ctx sdk.Context, plan upgradetypes.Plan) {
-			var genState liquiditytypes.GenesisState
-			genState.Params = liquiditytypes.DefaultParams()
-			genState.Params.PoolCreationFee = sdk.NewCoins(sdk.NewCoin("uatom", sdk.NewInt(40000000)))
-			app.LiquidityKeeper.InitGenesis(ctx, genState)
-		})
+
+	app.UpgradeKeeper.SetUpgradeHandler(
+		"vega",
```
alexanderbez: nit: let's put all usage of `"vega"` as a `const`

app/app.go
```diff
@@ -407,32 +530,65 @@ func NewGaiaApp(
...
+		func(ctx sdk.Context, _ upgradetypes.Plan, _ module.VersionMap) (module.VersionMap, error) {
+			// TODO: Verify these parameters are correct and intended.
+			app.IBCKeeper.ConnectionKeeper.SetParams(ctx, ibcconnectiontypes.DefaultParams())
+
+			fromVM := map[string]uint64{
```
alexanderbez: nit: avoid using direct string values here. Either use the `module.Name` or see if we can just iterate over `mm`

app/ante_handler.go
colin-axner: @AdityaSripal? I think the answer is yes if you want the IBC ante handler which discards redundant relay txs. From my understanding the ante handler isn't extensible without redefining it

AdityaSripal: Yes because we're adding an ante decorator (for IBC) that's not part of the SDK default list

app/ante_handler.go
AdityaSripal: Yes that is correct, order here is not important but better to just keep it last imo

AdityaSripal (commented):
Antehandler changes look good!

cmd/gaiad/cmd/root.go
alexanderbez: Yes :p

app/ante_handler.go
alexanderbez: But this:

> From my understanding the ante handler isn't extensible without redefining it

Do we really have to redefine everything? Can't we just add/replace what we want?

yaruwangway: another way to do it maybe to wrap the anteHandler the ` func NewAnteHandler` returns.
so the `func NewAnteHandler ` still returns the default antehandler, which is `sdk.ChainAnteDecorators(anteDecorators...), `, we just "chain" this again with the ibc decorator to do a final ibc antehandler check. since ibc handler's order in the chain does not matter.

app/app.go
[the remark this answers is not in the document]
okwme: asked @amaurym to take a look!

app/pubkey_replacement.go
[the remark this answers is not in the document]
okwme: this seems verified at this point but the issue you linked @alexanderbez seems additional? was there something else there you wanted to ensure?

app/pubkey_replacement.go
```diff
@@ -69,7 +69,7 @@ func loadKeydataFromFile(clientCtx client.Context, replacementrJSON string, genD
 
 			toReplaceValConsAddress, _ := val.GetConsAddr()
 
-			consPubKey, err := sdk.GetPubKeyFromBech32(sdk.Bech32PubKeyTypeConsPub, replacement.ConsensusPubkey)
+			consPubKey, err := legacybech32.UnmarshalPubKey(legacybech32.ConsPK, replacement.ConsensusPubkey) //todo check here
```
amaury1093: lgtm

app/ante_handler.go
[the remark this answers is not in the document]
okwme: just to confirm, this order actually doesn't matter @AdityaSripal ? by putting it at the bottom of the list it will be the first to be returned, correct?

app/ante_handler.go
```diff
@@ -0,0 +1,64 @@
...
+func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
+    if options.AccountKeeper == nil {
+        return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "account keeper is required for ante builder")
+    }
+
+    if options.BankKeeper == nil {
+        return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "bank keeper is required for ante builder")
+    }
+
+    //todo Check each field in channelKeeper is not nil ?
```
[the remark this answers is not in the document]
okwme: lets check for each of them, i've run into an error before when this wasn't checked.

app/app.go
```diff
@@ -407,32 +531,50 @@ func NewGaiaApp(
...
+
+	app.UpgradeKeeper.SetUpgradeHandler(
+		upgradeName,
+		func(ctx sdk.Context, _ upgradetypes.Plan, _ module.VersionMap) (module.VersionMap, error) {
+			// TODO: Verify these parameters are correct and intended.
+			app.IBCKeeper.ConnectionKeeper.SetParams(ctx, ibcconnectiontypes.DefaultParams())
+
+			fromVM := make(map[string]uint64)
+			for moduleName := range app.mm.Modules {
+				fromVM[moduleName] = 1
```
amaury1093, line 562: In 0.44, x/auth's ConsensusVersion is `2`, which means this upgrade will run x/auth's 1->2 migration (the one with the fix for tracking vesting accounts delegations). If I'm not mistaken, the Hub already included that fix, right? If so, then we should set:

```go
fromVM["auth"] = auth.AppModule{}.ConsensusVersion() // Will skip x/auth migrations
```

amaury1093: Also, for authz and feegrant, setting to `1` means skipping their InitGenesis (see https://docs.cosmos.network/master/core/upgrade.html#overwriting-genesis-functions).

I believe both their DefaultGenesis is empty state, so the current code works, but imo it still feels clearer to explicitly set `fromVM["authz"] = 0, fromVM["feegrant"] = 0`.

amaury1093 (commented):
authz and feegrant integration lgtm 👍

app/app.go
alexanderbez: Updated.

app/ante_handler.go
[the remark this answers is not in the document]
alexanderbez: Order _does_ matter, but IBC at the end is OK.

app/pubkey_replacement.go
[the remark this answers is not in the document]
alexanderbez: That PR I linked was to address Colin's question as to why we made the change :)

These files will be removed anyway.

alexanderbez:
I think we can merge this once CI passes. If there are future versions of the SDK and/or ibc-go that we want to use, we can simply create PRs to bump `go.mod` in a future PR.

app/app.go
```diff
@@ -407,32 +531,54 @@ func NewGaiaApp(
...
```
amaury1093, line 562: (didn't get an answer for https://github.com/cosmos/gaia/pull/942#discussion_r709064892) Did the hub ever integrate the vesting tracking delegation issue? (https://github.com/cosmos/cosmos-sdk/pull/8865)? If yes, we would also need `fromVM[moduleName] = 2`

amaury1093 (approved):
lgtm!

app/app.go
alexanderbez: AFAIK the vesting issue still exists on the Hub and was never fixed because that fix is included in 0.44.x. Is that correct @zmanian?

amaury1093: OK, in this case we should revert 3f7c39babbac08788a43fd5b731bd46c2d39c8eb

app/pubkey_replacement.go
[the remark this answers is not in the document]
colin-axner: sweet, seems like this has been verified. Resolving

app/app.go
alexanderbez: @amaurym we have the following as of right now:

```go
			fromVM := make(map[string]uint64)
			for moduleName := range app.mm.Modules {
				fromVM[moduleName] = 1
			}

			// override versions for _new_ modules as to not skip InitGenesis
			fromVM[authz.ModuleName] = 0
			fromVM[feegrant.ModuleName] = 0
```

Do we need to explicitly set the version of `auth`?

amaury1093: > @amaurym we have the following as of right now:

Looking at the files, I see: https://github.com/cosmos/gaia/blob/f42f5023af99367d2bff60edd2a9e40ad9396df5/app/app.go#L568-L569

which we need to revert (commit 3f7c39b)

okwme: @yaruwangway has done this

okwme (approved):

PR #979, opened 2021-09-20
gaia v6.0.0-rc2

yaruwangway:
Closes: #969 #991

## Description
This PR is for releasing gaia v6.0.0-rc2. The following changes are included:
- [x] Change Go version to 1.17
- [x] Add IBC Router to Hub if proposal_id 56 passes the voting
- [x] [fix: testnet cli command update genesis supply](https://github.com/cosmos/gaia/pull/991)
- [x] [Remove legacy migration code](https://github.com/cosmos/gaia/issues/969#issue-997234532)
- [x] Bump ibc-go to [v1.2.1](https://github.com/cosmos/ibc-go/releases/tag/v1.2.1)

______

For contributor use:

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Added appropriate labels to PR (ex. `WIP`, `R4R`, `docs`, etc)
- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/main/README.md#merging-a-pr))

PR #983, opened 2021-09-21
Add router module to Vega upgrade

jackzampolin:


app/app.go
```diff
@@ -99,6 +99,10 @@ import (
 	dbm "github.com/tendermint/tm-db"
 
 	gaiaappparams "github.com/cosmos/gaia/v6/app/params"
+	router "github.com/strangelove-ventures/packet-forward-middleware/router"
```
yaruwangway, line 102: no need to write "router" if the alias is the same name as the package

yaruwangway (commented):
Hi Jack, it is great that we can include router module to gaia vega-rc2 release!
Please confirm if the following code change need to be added:
- add `routertypes.ModuleName` in `app.mm.SetOrderBeginBlockers` and `app.mm.SetOrderInitGenesis`
-  add `fromVM[routertypes.ModuleName] = 0` if the router module do not want to skip init genesis. like this:https://github.com/cosmos/gaia/blob/06d6f8f29d989da7a5275e2a664d7ca7bf1992d8/app/app.go#L565
- add the router module store in the store upgrade. like here: https://github.com/cosmos/gaia/blob/06d6f8f29d989da7a5275e2a664d7ca7bf1992d8/app/app.go#L579

alexanderbez (changes requested):
This needs to go into `main` then backported into the release branch. Otherwise, LGTM.

jackzampolin:
Ok so this is good, but I just want to make sure to test the e2e with this build. Should have this done by eow

jackzampolin:
Will rebase against `main`

okwme:
this work was cherry-picked by @yaruwangway and added to the `vega-rc2` branch as part of https://github.com/cosmos/gaia/pull/979. Going to close this PR unless there's another reason to keep it open.

PR #1001, opened 2021-09-29
Release/v5.0.6

okwme:
Not sure why the content of `v5.0.6` doesn't already exist inside `v5.0.x` but this would make it an up to date target for PR #999


go.mod
```diff
@@ -3,18 +3,20 @@ module github.com/cosmos/gaia/v5
...
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.7.0
-	github.com/tendermint/tendermint v0.34.11
+	github.com/tendermint/tendermint v0.34.13
 	github.com/tendermint/tm-db v0.6.4
 )
 
 replace google.golang.org/grpc => google.golang.org/grpc v1.33.2
 
+replace github.com/tendermint/tendermint => github.com/tendermint/tendermint v0.34.13
```
charleenfei, line 20: this line probably redundant if you are bumping versions above

yaruwangway:
> Not sure why the content of `v5.0.6` doesn't already exist inside `v5.0.x` but this would make it an up to date target for PR #999

I think it is because v5.0.4, 5.0.5 and 5.0.6 are checkout their own branches off branch release/v5.0.x?

yaruwangway (changes requested):
I think the `app.go` also changed.  in the `setOrderBeginBlockers`, `slashingingtypes.ModuleName` is added. 
 
we also need to merge release/5.0.5, release/5.0.4 into release/5.0.x.

go.mod
colin-axner: I believe it is overwriting the tm version the SDK uses

okwme: yup this is to force the SDK to use that version of tendermint too

yaruwangway (approved):

PR #1085, opened 2021-11-30
Add Permissioned CosmWASM to hub roadmap

jackzampolin:
🙏

### DIFF
--- a/docs/roadmap/cosmos-hub-roadmap-2.0.md
+++ b/docs/roadmap/cosmos-hub-roadmap-2.0.md
@@ -49,6 +49,7 @@ The upgrades aim to add features such as liquidity, economic security, usability
   - Frees secure and low-risk delegations for use in other parts of the Cosmos ecosystem
   - Features include enabling transfer of rewards and voting rights
 - Items still being discussed:
+  - Governance permissioned CosmWASM instance on the hub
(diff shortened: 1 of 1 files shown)


tac0turtle (approved):
Lgtm

zmanian:
I'd like to get some record of Lido's use case for cosmwasm on the Hub somewhere.

jackzampolin:
I also think it would be cool to have DAO DAO on the hub to make it the best place to manage daos in the interchain

okwme (approved):

ethanfrey:
I would love to support this from Confio's side.
We should have a Cosmos SDK 0.44 compatible wasmd 1.0 release within a month.
Which is my prerequisite to recommend it to go to the hub for security/stability.
(Welcome more auditors/eyes on the Go side)

This was the original intention of the Cosmos Hub Community Spend proposal in summer 2020, and very happy to complete that work.

vshvsh:
> I'd like to get some record of Lido's use case for cosmwasm on the Hub somewhere.

Lido does want to have an ecosystem-aligned place to deploy its smart contracts and Hub looks like the best fit. Will have more color on that later when we're out of the upgrade procedure for Terra liquid staking which consumes a lot of focus.

Issue #977, opened 2021-09-18
Clean up Gaia repo


okwme:
## Summary

Gaia is full of stale issues and out of date docs. Let's make it a better place to live.

## Proposal

* close stale issues
* solve real issues
* update all docs
* add governance to this repo
* add mainnet to this repo
* update docs build system
* add some flair

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

faddat:
YES!

dcwalk:
Where can we propose emoji flair for a specific area? 😂 

The existing ones from discord++:
⚛️
💫 - ecosystem
🤝 - ibc
🔭 - status
🔨 - cosmos sdk
🧰 - cosmos sdk modules
🍩 - hub (!!)

candidate Governance emoji?:
📜 🗳️ ⚖️ 🧭

dcwalk:
I can take on closing out stale issues, addressing current ones, and updating parts of docs for gov- related topics! I've reviewed/tagged `docs` and `governance` issues in this repo. Post-gov repo merge I can transfer remaining issues from the governance repo and we'll have all documented gov issues in one place :sparkles:  I have ideas for refining gov docs once they land that I'm eager to take on as well.

Happy to join a doc / issue review call if needed!   

Is it okay to edit your top comment @okwme to add in tasks / checklist ? Or better to break out bite-sized issues and reference this top one? Still learning norms around github collaboration here :smile:

dcwalk:
Udpate/Smaller issues from governance (docs) side:
- [x] add governance to this repo
- [x] Small gov docs cleanups / updates:
   - https://github.com/cosmos/gaia/issues/1042 (more details on params keys + json validator approach)
   - https://github.com/cosmos/gaia/issues/1035 
- [ ] Bigger (proposed) updates:
   - https://github.com/cosmos/gaia/issues/1055

nooomski:
Closed by bot, opening again.

PR #1090, opened 2021-12-03
docs: Set up locales for existing translations

dcwalk:
This PR arises from https://github.com/cosmos/gaia/issues/977: 
- moves the existing translations into their own folder
- sets up locales for  `es` `en-US` `ko` `zh-CN`
- adds a bit of detail to the DOCS_README (borrowed from cosmos-sdk docs): https://github.com/cosmos/cosmos-sdk/blob/master/docs/DOCS_README.md


docs/es/README.md
```diff
@@ -1,22 +1,16 @@
-<!--
-parent:
-  order: false
-layout: home
--->
-
```
okwme, line 6: @dcwalk did you confirm this doesen't break the vue build? i had thought these comments were necessary for some part of the build process

dcwalk: Confirmed just now by building, a index is generated in with the page contents:
Screenshot_2021-12-09_09-40-32 

I removed frontmatter here to match zh and ko READMEs, but happy to add to all if preference is for that. 

If you are referring to the "order: false" line (which is mentioned in [cosmos theme readme](https://github.com/cosmos/vuepress-theme-cosmos#file-configuration)), it doesn't seem like it is needed when setting up translated folders, none of them are added to the English version navigation.

okwme: 👍

okwme (approved):

okwme:
ah looks like there's some conflicts, can you resolve those and i'll merge?

dcwalk:
resolved @okwme!

docs/DOCS_README.md
```diff
@@ -7,6 +7,13 @@ parent:
 
 If you want to open a PR on Gaia to update the documentation, please follow the guidelines in the [`CONTRIBUTING.md`](https://github.com/cosmos/gaia/tree/main/CONTRIBUTING.md)
 
+## Internationalization
+
+- Translations for documentation live in a `docs/<locale>/` folder, where `<locale>` is the language code for a specific language. For example, `zh` for Chinese, `ko` for Korean, `es` for Spanish, etc.
```
yaruwangway, line 12: `docs/<locale>/` ->  `docs/translations/<locale>/`

dcwalk: This PR removes the `/translations/` subfolder as that actually doesn't allow translated text to render on built docs!

yaruwangway: ha ,ok!

yaruwangway (approved):

dcwalk:
This is out of date with main again, I'll ff and resolve the conflict then merge!

Issue #1134, opened 2021-12-28
Default to Rocksdb


faddat:
## Summary

The `whisky on the rocks` series of patches powers nearly every ibc relay operator's nodes on busy networks like Gaia, Terra, Osmosis and Juno.
 
## Problem Definition

We should move sober gaia to a rocksdb by default build and release process so that she:

* syncs about 3x faster
	* save operator time
	* save operator money
	* grow network of operators due to lower time burden

## Proposal

I'm proposing that we use the following replace line:

https://github.com/notional-labs/osmosis/blob/54c0b7e2c1dd01dd46e7430c155aba9f78db6303/go.mod#L126

in go.mod until either:

* tecbot accepts a pr fixing his gorocksdb lib

or

* we change the module path for the cosmos gorocksdb lib and update tm-db so that it imports the cosmos gorocksdb lib.

The performance improvements to many aspects of operating gaia are quite incredible with this one little fix.  

In the next 2-4 weeks, Notional should have a quicksync.io style server set up at an unlimited 10gbps provider in the united states.  If anyone in the community is aware of hosting centers with true 10gbps connections in asia, preferably Singapore or Hong Kong, we'd glady provide direct asia coverage, too. 

This service will provide:

* Archive
* Default
* Fully Pruned

snapshots and should make it much simpler for operators to use rocksdb.  Since rocksdb also helps with read performance, analytics on Gaia can be improved, too. 
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

faddat:
https://github.com/tendermint/tm-db/issues/211

tac0turtle:
sync is faster but how is overall size of state compared? how is the speed when caught up compared to other databases.

PR #1135, opened 2021-12-28
Fix rocksdb build tag usage

faddat:
this replaces the obsolete tecbot/gorocksdb library with the cosnmosified version.  

Alternative to this is an update + release to tm-db.  Maybe just a release.


zmanian:
I don't understand the purpose of this PR?

zmanian:
It looks like adding an unused dependency to go.mod

faddat:
it is true!  It looks that way.  However, without that....



```bash
go install -tags rocksdb ./...
```

does not work because tecbot/rocksdb is quite out of date and imported by tm-db v0.6.6.

Also, I am making several of these currently and firing them off undocumented like:

```bash
gh pr create
```

faddat:
https://github.com/tendermint/tm-db/blob/06fdc12b3cc0ff8eb87ec0cd0b5daa9cb5091d2b/go.mod#L41

... is better, I'll amend

faddat:
Probably would also help if I were to say:

Works towards: #1134

tac0turtle (approved):
utACK, did test this? I know cosmos is on a newer version of rocks.

faddat:
I've tested it.  I think all it sidesteps is the outdated version in the newest release of tm-db.

So could be solved with a new release of tm-db or also using the latest commit hash of tm-db.

Adding one more bit now, just using v1.1.1 of gorocksdb from cosmos.  mildly cleaner.

okwme:
Feels dangerous to be overwriting upstream dependencies. As you mentioned the correct way would be for tm-db to update and to incorporate those updates. Is there any reason why this can't wait for then?

tac0turtle:
tm-db has merged the work just not cut a release. Not sure when that will happen, that repo is deprecated more or less

faddat:
Marko, tm-db is deprecated?

Can you tell me more?

Thanks!

@okwme -- this has been used on several chains successfully.  It dramatically improves sync speed and query rate.  For gaia, we can hit the trifecta-- state sync with rocksdb.   And then it utterly flies.

For me, my specific use case for this has been ibc relaying.  The difference is pretty extreme.

zmanian:
I'm feeling more in favor of merging this based on other usage of rocksdb I see in the ecosystem.

okwme (approved):

zmanian (approved):

okwme (approved):

PR #1204, opened 2022-01-20
Add & style new tabs component for docs

glnro:
This pr enables markdown tabbing for docs.

Ex:
Screen Shot 2022-01-27 at 12 33 07 PM
Screen Shot 2022-01-27 at 12 33 00 PM

PR #1217, opened 2022-01-27
Add tabbing for markdown docs

glnro:


nooomski (approved):
Tested ACK as per the now closed #1204

PR #1219, opened 2022-01-28
 v07-Theta prepare

### LINKED ISSUE #1215: Theta Release Meta Issue
Theta release meta issue for tracking all issues to make the release. Release should include:
* IBC v3 (confirm timelines, will need to work off of alpha for the time being)
  * Interchain Accounts (has its own meta issue)
* Cosmos SDK v0.45
  * confirm there are no breaking changes
* Liquidity module update
* Routing Module Update
  * strangelove said they'd begin on this
  * should confirm as well which version of IBC to target

yaruwangway:
Closes: #1215

- [x] bump cosmos-sdk to v0.45.1
- [x] bump ibc-go to v3
- [x] liquidity 1.5.0
- [x] check orders  in `SetOrder{BeginBlocker, EndBlocker, InitGenesis, ExportGenesis}` and check all modules are included
- [x] add migration logs when upgrade
- [x] test migration logs: "adding a new module: interchainaccounts" is in the log.
- [x] bump liquidity module
- [x] bump router module
- [ ] changelog
- [ ] check replace in go.mod
- [x] gaia module v6 -> v7
- [x] add Interchain Accounts


app/app.go
```diff
@@ -463,24 +465,48 @@ func NewGaiaApp(
...
+		distrtypes.ModuleName,
+		slashingtypes.ModuleName,
+		minttypes.ModuleName,
+		genutiltypes.ModuleName,
+		evidencetypes.ModuleName,
+		authz.ModuleName,
+		feegrant.ModuleName,
+		paramstypes.ModuleName,
+		upgradetypes.ModuleName,
+		vestingtypes.ModuleName,
```
yaruwangway, line 517: Hi @robert-zaremba, @alexanderbez , does the order of`moduleNames` in `SetOrderBeginBlockers`, `SetOrderEndBlockers ` and `SetOrderInitGenesis ` functions matter ?

should `auth` comes after `bank` and `staking` in `SetOrderInitGenesis ` function ?
Thank you!

app/app.go
```diff
@@ -401,16 +402,18 @@ func NewGaiaApp(
...
 		app.BankKeeper,
 		scopedTransferKeeper,
 	)
 	transferModule := transfer.NewAppModule(app.TransferKeeper)
+	ibcmodule := transfer.NewIBCModule(app.TransferKeeper)
 
 	app.RouterKeeper = routerkeeper.NewKeeper(appCodec, keys[routertypes.StoreKey], app.GetSubspace(routertypes.ModuleName), app.TransferKeeper, app.DistrKeeper)
 
-	routerModule := router.NewAppModule(app.RouterKeeper, transferModule)
+	routerModule := router.NewAppModule(app.RouterKeeper, ibcmodule)
```
okwme: why did this change?

yaruwangway: before, the `transferModule` satisfy all the interface of `transfermodule` and `ibcmodule`, now these two are separate

yaruwangway:
The migration log is not show during migrate due to the modules are the same version as before upgrade.

okwme:
missing interchain accounts from IBC. Can take example of what's needed from our previous integration: 
https://github.com/cosmos/gaia/blob/ica-acct-auth/app/app.go

yaruwangway:
@okwme last time you mentioned router module (packet-forward-middleware) is under some tests? how is the result ?

yaruwangway:
I think we should not merge this to main, because if we have a new release process, we might want to rebase this branch or move it to other branches before merge to main.

app/app.go
```diff
@@ -145,19 +154,26 @@ var (
...
 		minttypes.ModuleName:           {authtypes.Minter},
 		stakingtypes.BondedPoolName:    {authtypes.Burner, authtypes.Staking},
 		stakingtypes.NotBondedPoolName: {authtypes.Burner, authtypes.Staking},
 		govtypes.ModuleName:            {authtypes.Burner},
 		liquiditytypes.ModuleName:      {authtypes.Minter, authtypes.Burner},
 		ibctransfertypes.ModuleName:    {authtypes.Minter, authtypes.Burner},
 	}
+
+	// module accounts that are allowed to receive tokens
+	allowedReceivingModAcc = map[string]bool{
```
yaruwangway: before, we block all module account addresses to receive tokens by `MsgSend`, but interchain account
 do not block distribution module (which scaffold from starport), should we keep blocking all the address or make distribution an exception ? 

https://github.com/cosmos/interchain-accounts/blob/7a4ade49c4ddc74b44e23c44906be371d00a322b/app/app.go#L547
https://github.com/cosmos/interchain-accounts/blob/7a4ade49c4ddc74b44e23c44906be371d00a322b/app/app.go#L168

okwme: we actually allow all modules in gaia https://github.com/cosmos/gaia/blob/v6.0.3/app/app.go#L653-L661

app/app.go
okwme: is this still an open question?

amaury1093: > does the order ofmoduleNames in SetOrderBeginBlockers, SetOrderEndBlockers and SetOrderInitGenesis functions matter ?

Yes. Unfortunately the easiest way currently to know the order is to follow the comments.

amaury1093 (approved):
SDK code lgtm (Set*Order functions, upgrade handler)

app/app.go
```diff
@@ -566,31 +652,21 @@ func NewGaiaApp(
 
 	app.UpgradeKeeper.SetUpgradeHandler(
 		upgradeName,
-		func(ctx sdk.Context, _ upgradetypes.Plan, _ module.VersionMap) (module.VersionMap, error) {
+		func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
 			app.IBCKeeper.ConnectionKeeper.SetParams(ctx, ibcconnectiontypes.DefaultParams())
```
crodriguezvega: You can actually delete this line because this run already in a previous upgrade, right?

app/app.go
```diff
@@ -600,9 +676,7 @@ func NewGaiaApp(
 	}
 
 	if upgradeInfo.Name == upgradeName && !app.UpgradeKeeper.IsSkipHeight(upgradeInfo.Height) {
-		storeUpgrades := store.StoreUpgrades{
-			Added: []string{authz.ModuleName, feegrant.ModuleName, routertypes.ModuleName},
-		}
+		storeUpgrades := store.StoreUpgrades{}
```
crodriguezvega: For ibc-go v3 upgrade you need to have here the following:

```
storeUpgrades := store.StoreUpgrades{
	Added: []string{icatypes.ModuleName},
}
```

To ensures that the new module's stores are added to the multistore before the migrations begin.

crodriguezvega (changes requested):

app/app.go
```diff
@@ -193,17 +209,23 @@ type GaiaApp struct { // nolint: golint
...
+	RouterKeeper        routerkeeper.Keeper
 
 	// make scoped keepers public for test purposes
-	ScopedIBCKeeper      capabilitykeeper.ScopedKeeper
-	ScopedTransferKeeper capabilitykeeper.ScopedKeeper
+	ScopedIBCKeeper           capabilitykeeper.ScopedKeeper
+	ScopedTransferKeeper      capabilitykeeper.ScopedKeeper
+	ScopedICAControllerKeeper capabilitykeeper.ScopedKeeper
+	ScopedICAHostKeeper       capabilitykeeper.ScopedKeeper
+	ScopedInterTxKeeper       capabilitykeeper.ScopedKeeper
```
damiannolan: You can remove this keeper. This comes from the `intertx` demo module for interchain-accounts.

app/app.go
```diff
@@ -566,31 +650,20 @@ func NewGaiaApp(
...
-			newVM, err := app.mm.RunMigrations(ctx, app.configurator, fromVM)
-			if err != nil {
-				return nil, err
-			}
-			// now update auth version back to 1, to make the second RunMigrations includes only auth
-			newVM[authtypes.ModuleName] = 1
-
-			// RunMigrations twice is just a way to make auth module's migrates after staking
-			return app.mm.RunMigrations(ctx, app.configurator, newVM)
+		func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```
crodriguezvega, line 631: When testing the upgrade of ibc-go from v2 to v3 with gaia, we found an issue and we have fixed it now in ibc-go v3.0.0-rc1. It's documented [here](https://github.com/cosmos/ibc-go/blob/v3.0.0-rc1/docs/migrations/v2-to-v3.md#upgrade-proposal). You will need to update your dependency. 
If you have any questions about it, let me know.

app/app.go
colin-axner: Won't allowing the distribution module to receive tokens break [invariants](https://github.com/cosmos/cosmos-sdk/blob/26d573e3dab889bb5527a454b1a6437a8760c263/x/distribution/keeper/invariants.go#L138)? While not well documented, this was the intention of the blocked addresses, to prevent invariants from being broken

app/app.go
```diff
@@ -401,19 +418,46 @@ func NewGaiaApp(
...
 		app.GetSubspace(ibctransfertypes.ModuleName),
 		app.IBCKeeper.ChannelKeeper,
+		app.IBCKeeper.ChannelKeeper,
 		&app.IBCKeeper.PortKeeper,
 		app.AccountKeeper,
 		app.BankKeeper,
 		scopedTransferKeeper,
 	)
 	transferModule := transfer.NewAppModule(app.TransferKeeper)
+	ibcmodule := transfer.NewIBCModule(app.TransferKeeper)
```
crodriguezvega: Maybe for more clarity you could name this `transferIBCModule`.

app/app.go
```diff
@@ -401,19 +418,46 @@ func NewGaiaApp(
...
+		appCodec, keys[icahosttypes.StoreKey],
+		app.GetSubspace(icahosttypes.SubModuleName),
+		app.IBCKeeper.ChannelKeeper,
+		&app.IBCKeeper.PortKeeper,
+		app.AccountKeeper,
+		scopedICAHostKeeper,
+		app.MsgServiceRouter(),
+	)
+	icaModule := ica.NewAppModule(&app.ICAControllerKeeper, &app.ICAHostKeeper)
+	icaControllerIBCModule := icacontroller.NewIBCModule(app.ICAControllerKeeper, ibcmodule)
```
crodriguezvega: You're using here the IBC transfer module as the ICA authentication module... It's a bit strange, but I guess you need to pass something here to make the compiler happy. Just keep in mind that you need to update this when you have a proper authentication module.

crodriguezvega (approved):
Great work, @yaruwangway! 🚀

go.mod
```diff
@@ -1,19 +1,19 @@
...
-	github.com/gravity-devs/liquidity v1.4.2
+	github.com/gravity-devs/liquidity v1.5.0
 	github.com/rakyll/statik v0.1.7
 	github.com/spf13/cast v1.4.1
 	github.com/spf13/cobra v1.3.0
-	github.com/strangelove-ventures/packet-forward-middleware v1.0.1
+	github.com/strangelove-ventures/packet-forward-middleware/v2 v2.1.1
 	github.com/stretchr/testify v1.7.0
 	github.com/tendermint/tendermint v0.34.14
-	github.com/tendermint/tm-db v0.6.6
```
mmulji-ic, line 16: @tychoish Any idea why this was reverted?

app/app.go
```diff
@@ -566,31 +628,51 @@ func NewGaiaApp(
...
+			// create ICS27 Controller submodule params
+			controllerParams := icacontrollertypes.Params{}
+			// create ICS27 Host submodule params
+			hostParams := icahosttypes.Params{
+				HostEnabled: true,
+				AllowMessages: []string{
+					"/cosmos.authz.v1beta1.MsgExec",
+					"/cosmos.authz.v1beta1.MsgGrant",
+					"/cosmos.authz.v1beta1.MsgRevoke",
+					"/cosmos.bank.v1beta1.MsgSend",
```
crodriguezvega, line 643: Just suggestion, so feel free to ignore it... Instead of hardcoding these strings you could do something like this: for example, for the `MsgSend` of bank, you can do `sdk.MsgTypeURL(&banktypes.MsgSend{})`, which should return the same string. This way you don't take the risk that you may have a typo in any of this harcoded strings.

glnro: +1 we should try to reduce any hardcoding wherever we can

PR #1230, opened 2022-02-04
app.toml template gRPCWeb port conflict fix

RaspiRepo:
Bug fix/update in contrib/testnets/test_platform

- ```The minimum-gas-prices config in app.toml is set to the empty string. This defaults to 0 in the current version, but will error in the next version (SDK v0.45)```
change minimum-gas-prices = "" to   minimum-gas-prices = "0.25stake" 

- added  gRPC Web Configuration to resolve port conflict 9091 for 4 node test


okwme (approved):
thanks!

PR #1235, opened 2022-02-04
Testnet docs feedback

glnro:
This PR improves visibility for flags and port values when joining the testnet.


okwme (approved):
nice catches 🙏 ty for update

PR #1249, opened 2022-02-09
Initialize code analysis

glnro:
This PR includes the template github generates for their static code analysis tool. Github's code scanning tool analyzes the the github repository to identify security vulnerabilities and errors. This check runs as a github action, and as a team we can discuss refining parameters such as how often the job runs and on which steps it's triggered. Github also offers credential scanning, but this would have to be setup by a repository admin.


faddat:
Seems prudent

faddat (approved):
generally speaking, I try to use tooling like this wherever possible, to catch things that we might not otherwise catch.

mmulji-ic (approved):

PR #1250, opened 2022-02-10
update deposit requirement to 64 atom

hxrts:
changed following this proposal https://www.mintscan.io/cosmos/proposals/47


okwme:
nice thank you @hxrts !

PR #1245, opened 2022-02-09
Fix Broken Links

nooomski:
Some of the errors in the [Check Markdown Links](https://github.com/cosmos/gaia/actions/runs/1815731321) action are fixed in this PR. Some remain unfixed and need to be ignored.

@mmulji-ic as far as I can tell you haven't yet changed the config to retry on 429s, right? If you're still planning to do that, I would recommend adding the following folders to the `ignore` list in the [config file](https://github.com/tcort/markdown-link-check#config-file-format):

```
docs/migration
docs/governance/proposals
docs/governance/param-changes
docs/zh
docs/es
docs/ko
```

These are historical folders not in use anymore, or translations that are outdated and currently not displayed on the website. Additionally, the following two individual files can also be ignored:

```
docs/roadmap/README.md
docs/governance/state-of-cosmos-governance-2021.md
```

FYI, I could not reproduce the 403 errors in `docs/resources/ledger.md`, so I'm not entirely sure why the check fails here.

`docs/hub-tutorials/deploy-testnet.html` is too big to fix, and needs a complete rewrite. There is a new issue for that [here](https://github.com/cosmos/gaia/issues/1244). In the interim I would suggest also placing this one on the ignore list.

@mmulji-ic feel free to add your config file to my branch so we can merge this in one PR.

PR #1254, opened 2022-02-10
DOCS: Fix params broken links

nooomski:
#1245 got merged a little tooooo quickly. Was under the incorrect impression the `docs/governance/param-changes` folder was not being used so didn't fix those links. This should solve those.

@mmulji-ic , keep in mind not to put the `docs/governance/param-changes` on the ignore list in the config. Also let me know if you need me to just get started on it.


nooomski:
Ehhh, why is it showing 24 files changed? Should be ~4 or 5. Didn't #1245 already get merged?

nooomski:
Added ignore tags to each archived / historical file that the link checker should skip. Should be good to test @mmulji-ic

okwme (approved):
looks like it's passing 🎉
why do these historical markdowns need to be skipped? they have broken links that should be kept for historical reasons?

nooomski:
@okwme almost all the docs where the link checker is skipped are currently offline, mostly because they're outdated. 
The old proposals definitely have old links that we don't want to update because the proposals are historical. 
There's also a lot of translated content that is currently so outdated it's not online (afaik).

An overview of folders I've skipped:
```
docs/migration
docs/governance/proposals
docs/zh
docs/es
docs/ko
```

And individual files: 
```
docs/governance/state-of-cosmos-governance-2021.md (historical)
docs/hub-tutorials/deploy-testnet.html (this one is outdated and needs a complete revisit as per #1244 )
```

Two links in `docs/resources/ledger.md` were working fine but kept getting marked as 403 to which I could find no workaround, so those two links contain tags so they get skipped. Not the cleanest solution unfortunately.

PR #1278, opened 2022-02-17
R4R: Update reproducible build command

okwme:
to use latest reproducible build docker image from https://github.com/tendermint/images which is published on the tendermintdev dockerhub account

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ build-reproducible: go.sum
         --env VERSION=$(VERSION) \
         --env COMMIT=$(COMMIT) \
         --env LEDGER_ENABLED=$(LEDGER_ENABLED) \
-        --name latest-build cosmossdk/rbuilder:latest
+        --name latest-build tendermintdev/rbuilder:latest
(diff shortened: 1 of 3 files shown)


okwme:
the old version uses go v1.15.x which freaks out when it sees the `retract` line in `go.mod`.
I think this is also why LGTM stopped working

okwme:
I had to use that image to build the images for the release in `v6.0.2`. It should be used going forward for running
`make distclean build-reproducible`

glnro:
Interesting...was this changed recently or always like this?

okwme:
not sure when it got updated but we didn't follow suit ;_;

okwme:
new docker image is here: https://github.com/tendermint/images/tree/master/rbuilder

okwme:
old one was here: https://hub.docker.com/layers/cosmossdk/rbuilder/latest/images/sha256-0870f37eb666e4c4be9fd22892a32c70247b177008fff06f27b2fb708fe95c34?context=explore

okwme:
updated docs, R4R @mmulji-ic

PR #1288, opened 2022-02-22
docs: gaia version to latest 6.0.3

pratikbin:


okwme:
thanks @pratikbalar !
the testnet is actually correct in using the RC since there was a minor breaking change in bumping the version of IBC before we released v6.0.0.

Also appreciate keeping our docs up to date but I'm hesitant to point people to v6.0.3 since it's the same as v6.0.0. I don't know if you followed the bothed releses of v6.0.1 and v6.0.2 last weekend but there was a regression discovered so we rolled back to v6.0.0 under a new release. Once v6.0.4 is released with non state-breaking changes I'd like to highlight and promote that release, but at this point there is no point in pushing v6.0.3.

Thank you again for the PR and keeping an eye on improving our repo though! I hope to see more in the future even tho we should close this one : )

PR #1291, opened 2022-02-22
WIP: Update README.md so ADRs show up correctly

okwme:
didn't realize that all folders in docs get added to https://hub.cosmos.network automatically.
The ADR section doesn't look bad but the first one is missing the word "Accounts" from "Interchain Accounts in the title. (https://hub.cosmos.network/main/readiness/adr-001-interchain-accounts.html).
I thought we should add correct headers as well with the 
```
<--
    parent
    order
       false
-->
```
kind of stuff


nooomski:
This should render the correct title and hide the template as well. Can someone review? @glnro ? @tokamak-git ?

PR #1305, opened 2022-03-01
update tendermint and tm-db. remove gorocksdb replace

faddat:
tm-db has been upgraded and now uses cosmos/gorocksdb by default.

tendermint has been updated to v0.34.16

:)


okwme:
thank you @faddat !
with the upgrade coming in less than a week these updates would be better targeted at the `theta-main` branch. Could you retarget to that branch for a v7.0.1 release?

alexanderbez:
replaced by: https://github.com/cosmos/gaia/pull/1406

Thanks @faddat

Issue #1263, opened 2022-02-15
Setup Cosmos Hub container registry


mmulji-ic:
Where to host ?
Github registry / Docker Hub registry?

pratikbin:
I would consider docker hub, cause then people that are using organization wide docker registry cache can benefit from it.

You can also host it on both.

tokamak-git:
Hey folks, have configured ghcr: `GitHub container registry` and have been able to push an image to it aswell under the cosmos branding.

Shall close this if no further investigation/modification is required.

mmulji-ic:
@tokamak-git @okwme - any idea who owns this repo? https://hub.docker.com/u/interchain

bert2002:
Are you going to publish more updated docker images in the future?

PR #1320, opened 2022-03-03
dockerfile: refactor image

pratikbin:
- Alpine version based on ARG
- Remove `workdir`, As default is `/root`.
- No cache apk package
- Move few lines for slightly better caching


okwme:
related to #1263 cc @tokamak-git

contrib/Dockerfile.test
```diff
@@ -18,17 +18,16 @@ RUN apk add --no-cache $PACKAGES && \
   make install
 
 # Final image
-FROM alpine:edge
+FROM alpine:latest
 
 # Install ca-certificates
-RUN apk add --update ca-certificates
-WORKDIR /root
```
tokamak-git, line 25: do wonder if its a good idea to remove the `WORKDIR` it may be relevant to the script being executed.

Could you verify there is no reference to `$WORKDIR` please?

contrib/Dockerfile.test
```diff
@@ -18,17 +18,16 @@ RUN apk add --no-cache $PACKAGES && \
   make install
 
 # Final image
-FROM alpine:edge
```
tokamak-git, line 21: Hey @pratikbin makes sense for us to refresh these. Not a big fan of `edge` or `latest` though. it may be a good idea to lock these down to a specific version, WDYT?

tokamak-git (changes requested):

contrib/Dockerfile.test
pratikbin: That's okay too if you want to manually upgrade alpine versions. I'll revert it back.

contrib/Dockerfile.test
pratikbin:
```diff
- WORKDIR /root
```

In baseimage which is alpine, they specify `WORKDIR`, so we don't have to specify it here.

```diff
- RUN apk add --update ca-certificates
```

Kindly keep this

tokamak-git (approved):
LGTM, thanks @pratikbin

okwme:
what's the status here? is this the correct PR or is work happening somewhere else?
R4R? to be merged?

tokamak-git:
@okwme good to merge, shall do a check it doesn't break anything the docker branch im working on and then merge this if ok.

tokamak-git:
Hey @pratikbin this change is no longer relavent. refer #1399

PR #1321, opened 2022-03-03
ci: refactor test pipeline

pratikbin:
- Use path based pipeline trigger to remove `technote-space/get-diff-action` action, That will save us pipeline triggers as well as time
- Add go mod caching

Also planning to rolling out this to all respective pipelines


pratikbin:
This will save us CI runs because we're handling CI trigger on file changes at jobs level.

tokamak-git (approved):
nice catch. looks good to me

PR #1271, opened 2022-02-16
WIP: update release note

yaruwangway:
- [ ] backport release 
       relevant issue: https://github.com/cosmos/gaia/issues/1122
       if let bot do the backport release ?
- [ ] major release
- [ ] point release
- [ ] release candidate release

PR #1323, opened 2022-03-03
merge Rc0/v6.0.4 into release/v6.0.x

okwme:
This is the PR that confirms everything that is going into the release.
After merged a new `release/v6.0.4` branch is created out of `v6.0.x`...


.github/CODEOWNERS
```diff
@@ -1,7 +1,7 @@
 # CODEOWNERS: https://help.github.com/articles/about-codeowners/
 
 # Primary repo maintainers
-*       @alessio @alexanderbez @jgimeno @shahankhatch @zmanian @yaruwangway
+*       @okwme @alexanderbez @zmanian @yaruwangway @jgimeno @crodriguezvega @cmwaters @jackzampolin @mmulji-ic @glnro @tokamak-git
```
okwme:
```suggestion
*      @ alessio @okwme @jgimeno @alexanderbez @zmanian @yaruwangway @jgimeno @crodriguezvega @cmwaters @jackzampolin @mmulji-ic @glnro @tokamak-git
```

okwme:
ok so i rebased main.
we typically don't do this, i believe the reasoning is that we want to be very clear about what is going into a release, and if we have a ton of commits (like above) and a ton of docs files (which are basically the only thing that doesn't go into a release), then it may make it difficult to review the PR.

The problem with never rebasing is that our releases don't have up to date docs. This isn't a problem if you only care about people looking at the live main docs. However it means that it's not possible to version control docs, since switching to v4.0.x of gaia doesn't mean you'll be able to read docs for v4.0.x. This is a problem for anyone who wants to set up a historical node.

so there is this one big messy PR that rebases as part of a release.
this branch would afterwards be merged into v6.0.x which is then merged into main.
at that point we'll have what we're looking for as a clean starting point for our new release process going forward (#1271).

yaruwangway:
> This is the PR that confirms everything that is going into the release. After merged a new `release/v6.0.4` branch is created out of `v6.0.x`...

Hi in the pr title saying merge  `Rc0/v6.0.4` ? just confirm this merge is for `release/v6.0.4` to `release/v6.0.x` ?

okwme:
good point @yaruwangway just updated :+1:

okwme:
closing to do a docs rebase later

PR #1332, opened 2022-03-14
Update contrib/statesync.bash Bash script

ademarco:
The `contrib/statesync.bash` Bash script contained outdated RPC endpoints and seeds. The current PR fixes this, while also improving its maintainability by:

1. Fetching genesis from https://github.com/cosmos/mainnet/raw/master/genesis.cosmoshub-4.json.gz, as per documentation
2. Fetching seeds by parsing https://raw.githubusercontent.com/cosmos/chain-registry/master/cosmoshub/chain.json via `jq`

### DIFF
--- a/contrib/statesync.bash
+++ b/contrib/statesync.bash
@@ -2,38 +2,43 @@
 # microtick and bitcanna contributed significantly here.
 set -uxe
 
-# set environment variables
+# Set Golang environment variables.
 export GOPATH=~/go
 export PATH=$PATH:~/go/bin
 
-
-# Install Gaia
+# Install Gaia.
 make install
 
-# MAKE HOME FOLDER AND GET GENESIS
+# Initialize chain.
 gaiad init test 
-wget -O ~/.gaia/config/genesis.json https://cloudflare-ipfs.com/ipfs/Qmc54DreioPpPDUdJW6bBTYUKepmcPsscfqsfFcFmTaVig
 
+# Prepare genesis file for cosmoshub-4.
+wget https://github.com/cosmos/mainnet/raw/master/genesis.cosmoshub-4.json.gz
+gzip -d genesis.cosmoshub-4.json.gz
+mv genesis.cosmoshub-4.json $HOME/.gaia/config/genesis.json
+
+# Set minimum gas price.
+sed -i'' 's/minimum-gas-prices = ""/minimum-gas-prices = "0.0025uatom"/' $HOME/.gaia/config/app.toml
+
+# Get "trust_hash" and "trust_height".
 INTERVAL=1000
-
-# GET TRUST HASH AND TRUST HEIGHT
-
-LATEST_HEIGHT=$(curl -s https://cosmoshub-4.technofractal.com/block | jq -r .result.block.header.height);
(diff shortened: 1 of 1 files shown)


alexanderbez (approved):

contrib/statesync.bash
```diff
@@ -2,38 +2,43 @@
...
+# Install Gaia.
 make install
 
-# MAKE HOME FOLDER AND GET GENESIS
+# Initialize chain.
 gaiad init test 
-wget -O ~/.gaia/config/genesis.json https://cloudflare-ipfs.com/ipfs/Qmc54DreioPpPDUdJW6bBTYUKepmcPsscfqsfFcFmTaVig
 
-INTERVAL=1000
+# Prepare genesis file for cosmoshub-4.
```
glnro: If you pass --chain-id cosmoshub-4 on `init` then downloading the genesis can be eliminated

okwme: nice catch @glnro , this is because with state sync there's nothing about needing initial state to be accurate within the genesis file, right?

I haven't confirmed this personally, could I get one more ACK that this change could go in from @alexanderbez or @marbar3778 ?

contrib/statesync.bash
```diff
@@ -2,38 +2,43 @@
...
 export PATH=$PATH:~/go/bin
 
-
-# Install Gaia
+# Install Gaia.
 make install
 
-# MAKE HOME FOLDER AND GET GENESIS
+# Initialize chain.
 gaiad init test
```
okwme:
```suggestion
gaiad init test --chain-id cosmoshub-4
```

as per @glnro 's suggestion. Would suggest committing this after confirmed from a second party

contrib/statesync.bash
okwme:
```suggestion
```
ibid

contrib/statesync.bash
```diff
@@ -2,38 +2,43 @@
...
 make install
 
-# MAKE HOME FOLDER AND GET GENESIS
+# Initialize chain.
 gaiad init test 
-wget -O ~/.gaia/config/genesis.json https://cloudflare-ipfs.com/ipfs/Qmc54DreioPpPDUdJW6bBTYUKepmcPsscfqsfFcFmTaVig
 
-INTERVAL=1000
+# Prepare genesis file for cosmoshub-4.
+wget https://github.com/cosmos/mainnet/raw/master/genesis.cosmoshub-4.json.gz
```
okwme:
```suggestion
```
ibid

contrib/statesync.bash
```diff
@@ -2,38 +2,43 @@
...
+gzip -d genesis.cosmoshub-4.json.gz
```
okwme:
```suggestion
```
ibid

contrib/statesync.bash
```diff
@@ -2,38 +2,43 @@
...
+mv genesis.cosmoshub-4.json $HOME/.gaia/config/genesis.json
```
okwme:
```suggestion
```
ibid

contrib/statesync.bash
ademarco: Thank you for the suggestion @glnro! I've tested it locally and it works well. I've pushed a fix that removes the explicit genesis file downloading.

okwme:
thanks @ademarco !

PR #1343, opened 2022-03-17
feat: bypass min fee check for IBC relayer txs

alexanderbez:
Bypass validator min fee check during `CheckTx` iff:

- Contains `MsgRecvPacket`, `MsgAcknowledgement`, and `MsgUpdateClient` only
- Given the above is true, the total gas is less than some constant

### DIFF
--- a/ante/fee.go
+++ b/ante/fee.go
@@ -1,0 +1,77 @@
+package ante
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	ibctransfertypes "github.com/cosmos/ibc-go/v2/modules/apps/transfer/types"
+	ibcclienttypes "github.com/cosmos/ibc-go/v2/modules/core/02-client/types"
+)
+
+const maxIBCRelayerMsgGasUsage = uint64(200_000)
+
+// MempoolFeeDecorator will check if the transaction's fee is at least as large
+// as the local validator's minimum gasFee (defined in validator config).
+//
+// If fee is too low, decorator returns error and tx is rejected from mempool.
+// Note this only applies when ctx.CheckTx = true. If fee is high enough or not
+// CheckTx, then call next AnteHandler.
+//
+// CONTRACT: Tx must implement FeeTx to use MempoolFeeDecorator
+type MempoolFeeDecorator struct{}
+
+func NewMempoolFeeDecorator() MempoolFeeDecorator {
+	return MempoolFeeDecorator{}
+}
+
+func (mfd MempoolFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
+	feeTx, ok := tx.(sdk.FeeTx)
+	if !ok {
+		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+
+	feeCoins := feeTx.GetFee()
+	gas := feeTx.GetGas()
+	msgs := feeTx.GetMsgs()
+
+	// Only check for minimum fees if the execution mode is CheckTx and the tx does
+	// not contain IBC relayer messages. If the tx does contain IBC relayer messages,
+	// it's total gas must be less than or equal to a constant, otherwise minimum
+	// fees are checked.
+	if ctx.IsCheckTx() && !simulate && !(isIBCRelayerTx(msgs) && gas <= uint64(len(msgs))*maxIBCRelayerMsgGasUsage) {
+		minGasPrices := ctx.MinGasPrices()
+		if !minGasPrices.IsZero() {
+			requiredFees := make(sdk.Coins, len(minGasPrices))
+
+			// Determine the required fees by multiplying each required minimum gas
+			// price by the gas limit, where fee = ceil(minGasPrice * gasLimit).
+			glDec := sdk.NewDec(int64(gas))
+			for i, gp := range minGasPrices {
+				fee := gp.Amount.Mul(glDec)
+				requiredFees[i] = sdk.NewCoin(gp.Denom, fee.Ceil().RoundInt())
+			}
+
+			if !feeCoins.IsAnyGTE(requiredFees) {
+				return ctx, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
+			}
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func isIBCRelayerTx(msgs []sdk.Msg) bool {
+	for _, msg := range msgs {
+		switch msg.(type) {
+		case *ibctransfertypes.MsgTransfer:
+			continue
+
+		case *ibcclienttypes.MsgUpdateClient:
+			continue
+
+		default:
+			return false
+		}
+	}
+
+	return true
+}
--- a/ante/fee.go
+++ b/ante/fee.go
@@ -3,7 +3,6 @@ package ante
 import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
-	ibctransfertypes "github.com/cosmos/ibc-go/v2/modules/apps/transfer/types"
 	ibcclienttypes "github.com/cosmos/ibc-go/v2/modules/core/02-client/types"
 	ibcchanneltypes "github.com/cosmos/ibc-go/v2/modules/core/04-channel/types"
 )
@@ -63,7 +62,7 @@ func (mfd MempoolFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate b
 func isIBCRelayerTx(msgs []sdk.Msg) bool {
 	for _, msg := range msgs {
 		switch msg.(type) {
-		case *ibctransfertypes.MsgTransfer:
+		case *ibcchanneltypes.MsgRecvPacket:
 			continue
 
 		case *ibcchanneltypes.MsgAcknowledgement:
--- a/ante/fee.go
+++ b/ante/fee.go
@@ -5,6 +5,7 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	ibctransfertypes "github.com/cosmos/ibc-go/v2/modules/apps/transfer/types"
 	ibcclienttypes "github.com/cosmos/ibc-go/v2/modules/core/02-client/types"
+	ibcchanneltypes "github.com/cosmos/ibc-go/v2/modules/core/04-channel/types"
 )
 
 const maxIBCRelayerMsgGasUsage = uint64(200_000)
@@ -63,6 +64,9 @@ func isIBCRelayerTx(msgs []sdk.Msg) bool {
 	for _, msg := range msgs {
 		switch msg.(type) {
 		case *ibctransfertypes.MsgTransfer:
+			continue
+
+		case *ibcchanneltypes.MsgAcknowledgement:
 			continue
 
 		case *ibcclienttypes.MsgUpdateClient:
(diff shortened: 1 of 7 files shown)


tac0turtle (approved):

jackzampolin:
Well, do we want to waive user fees? I personally think users should still pay fees on the `MsgTransfer` and also only whitelisting txs with `MsgUpdateClient` would exclude the case where a transaction is only made up of `MsgRecvPacket` (relying on an earlier `MsgUpdateClient`) and could increase fees paid by relayers.

ante/fee.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func isIBCRelayerTx(msgs []sdk.Msg) bool {
+	for _, msg := range msgs {
+		switch msg.(type) {
+		case *ibctransfertypes.MsgTransfer:
```
adizere: Are acknowledgements covered here?

alexanderbez: No, they're not. I'll add that case.

alexanderbez:
> Well, do we want to waive user fees? I personally think users should still pay fees on the `MsgTransfer` and also only whitelisting txs with `MsgUpdateClient` would exclude the case where a transaction is only made up of `MsgRecvPacket` (relying on an earlier `MsgUpdateClient`) and could increase fees paid by relayers.

We're not waiving fees here. We're bypassing min-fee checks. The user/relayer still pays fees (whatever they supplied)

ante/fee.go
alexanderbez: Added `MsgAcknowledgement`

charleenfei:
what about MsgRecvPacket and MsgAcknowledgement?

anilcse:
> We're not waiving fees here. We're bypassing min-fee checks. The user/relayer still pays fees (whatever they supplied)

I guess we don't need to bypass fees for `MsgTransfer` but would need to bypass fees for `MsgRecvPacket`, `MsgUpdateClient` and `MsgAcknowledgement`. We still expect fees from users but just don't charge relayers. wdyt?

alexanderbez:
> > We're not waiving fees here. We're bypassing min-fee checks. The user/relayer still pays fees (whatever they supplied)
> 
> I guess we don't need to bypass fees for `MsgTransfer` but would need to bypass fees for `MsgRecvPacket`, `MsgUpdateClient` and `MsgAcknowledgement`. We still expect fees from users but just don't charge relayers. wdyt?

So fees are still charged, but we're just bypassing the local validator min fee check. So for these particular txs, relayers can pay what they'd like.

alexanderbez:
Updated the message types 👍

anilcse (approved):
Lgtm

alexanderbez:
Added tests :)

okwme:
To confirm the mechanics here:
- transactions that contain those msg types can be set to a fee below a validator's min fee and it will not be automatically rejected by that validator for inclusion in the block?

Is the minfee set as a min gas price or a fee? I see min fee everywhere but each tx has a different fee based on gas*gasPrice right? Some transactions need more gas so would have a higher minfee. Or is minfee supposed to incorporate a global min for all msg types?

Worst case scenarios are:
- anyone can spam the network with these message types and fill blocks while paying no fees. Validators will have mo ability to block these spam messages because there is no way for them to filter without custom code? does this still require forking tendermint?

Osmosis allows 0 fees atm but haven't seen any adverse effects right? Only they're still just encouraging 0 fee validators so there's an instant and validator by validator fix that allows them to stop accepting 0fees if there was an attack. This change would make it hard / impossible to block on the hub?

What was wrong with the problem with the feegrant solution? Concern that relayers don't want to receive hub funding for regulatory reasons? Is this still a valid concern?

tac0turtle:
> What was wrong with the problem with the feegrant solution? Concern that relayers don't want to receive hub funding for regulatory reasons? Is this still a valid concern?

There isn't anything wrong with fee grant, but the treasury may want to fund use of the protocol, not so much simple messages for client updates and so on. Making these free would make the use of something like fee grant useful for user interactions, but relayer and ibc protocol specific messages stay free for the relayers.

alexanderbez:
> transactions that contain those msg types can be set to a fee below a validator's min fee and it will not be automatically rejected by that validator for inclusion in the block?

Correct.

> Is the minfee set as a min gas price or a fee? I see min fee everywhere but each tx has a different fee based on gas*gasPrice right? Some transactions need more gas so would have a higher minfee. Or is minfee supposed to incorporate a global min for all msg types?

Validators set a min gas **price**. Note, we bypass the min gas price check if the total gas is under some constant, otherwise, the check still holds.

> anyone can spam the network with these message types and fill blocks while paying no fees. 

You can already do this ;-)

> What was wrong with the problem with the feegrant solution? 

Freegrant isn't a solution here per-se, it really has nothing to do with fees directly. It has to do with the fact that the price of ATOM fluctuates and validators set min fee gas prices and relayers relaying IBC txs to the Hub do not meet these min gas prices. So either we:

1. Tell relayers to set the fees much higher than average to account for the price fluctuations, OR
2. Build an oracle into the relayer(s), i.e. have the relayer check for prices OR
3. Bypass min fee checks (this PR)

robert-zaremba:
I like @anilCSE  suggestion to only include `MsgRecvPacket`, `MsgUpdateClient` and `MsgAcknowledgement`.
Why do you want to bypass fee check for `MsgTransfer`?

robert-zaremba (commented):
* let's update the changelog
* shall we implement this in Cosmos SDK?

alexanderbez:
> I like @anilCSE suggestion to only include `MsgRecvPacket`, `MsgUpdateClient` and `MsgAcknowledgement`. Why do you want to bypass fee check for `MsgTransfer`?

Already committed :)

alexanderbez:
> * let's update the changelog
> * shall we implement this in Cosmos SDK?

Yes, I forgot the changelog. Thanks!
Re SDK, we cannot because this is IBC stuff, which the SDK does not include (for a reason)

adizere:
> Re SDK, we cannot because this is IBC stuff, which the SDK does not include (for a reason)

It could make sense for the ibc-go team to upstream these changes into the IBC module implementation, so that other chains benefit from the bypass? Or is it more sensible to keep this as a chain-level feature?

alexanderbez:
> > Re SDK, we cannot because this is IBC stuff, which the SDK does not include (for a reason)
> 
> It could make sense for the ibc-go team to upstream these changes into the IBC module implementation, so that other chains benefit from the bypass? Or is it more sensible to keep this as a chain-level feature?

That's a great question. My guess is that most developers and chains would find this non-contentious. However, there might be some that strictly do not want this. If we're to upstream to IBC directly, I propose that we do it a configurable manner, which should already be possible since this is all just middleware.

So I'm all for it.

adizere:
> That's a great question. My guess is that most developers and chains would find this non-contentious. However, there might be some that strictly do not want this. If we're to upstream to IBC directly, I propose that we do it a configurable manner, which should already be possible since this is all just middleware.
> 
> So I'm all for it.

Thanks for highlighting that, it's indeed more nuanced than it seems at first glance & not all chains may want the feature. CC-ing @crodriguezvega @charleenfei to potentially pass the ball on the IBC-go court.

tac0turtle:
> > Re SDK, we cannot because this is IBC stuff, which the SDK does not include (for a reason)
> 
> It could make sense for the ibc-go team to upstream these changes into the IBC module implementation, so that other chains benefit from the bypass? Or is it more sensible to keep this as a chain-level feature?

Instead of upstreaming we should allow applications to add messages to an array to be checked here. This way application developers can add these messages and possibly other ones to avoid fees on. Its a simple change in the sdk for now, maybe we can do it for 0.46

anilcse:
> 

I like this approach

okwme:
one difference i pointed out about 0 fees on osmosis is that if there is something like a spam attack, all that needs to happen is for individual validators is to raise their min fees. if we in-protocol make the skip happen, then we don't have a similar way of protecting the network. are there options that still enable a path to protecting the network? either via gov param (2 weeks is better than never) or some kind of a validator override? so it would be opt-in to setting validator gas fees instead of opt-out?

i love the idea of making these free but don't want to rely on current good conditions to last forever.

alexanderbez:
> if we in-protocol make the skip happen, then we don't have a similar way of protecting the network.

Yes, spamming would still be possible if bad actors just submitted **bogus** **IBC-only** txs. I can't really think of another way to both have these txs skip local checks but also be prevented from being used as spam.

Maybe @ValarDragon can weight in?

robert-zaremba:
One idea is to have a quota of free transactions. For example 1000 free txs / minute.

anilcse:
> > if we in-protocol make the skip happen, then we don't have a similar way of protecting the network.
> 
> Yes, spamming would still be possible if bad actors just submitted **bogus** **IBC-only** txs. I can't really think of another way to both have these txs skip local checks but also be prevented from being used as spam.
> 
> Maybe @ValarDragon can weight in?

Can we make it as a node setting? Similar to mingasfees, another setting for special txs? And a way to add/remove these special txs via governance

alexanderbez:
That is an option @anilCSE, yes.

ante/fee.go
```diff
@@ -0,0 +1,80 @@
...
+			if !feeCoins.IsAnyGTE(requiredFees) {
+				return ctx, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
+			}
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func isIBCRelayerTx(msgs []sdk.Msg) bool {
```
okwme, line 62: just to confirm, this means that a tx with IBC messages plus other messages will not get a free pass. only transactions with only approved IBC messages will get to pass.

app/ante_handler.go
```diff
@@ -6,8 +6,13 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/auth/ante"
 	channelkeeper "github.com/cosmos/ibc-go/v2/modules/core/04-channel/keeper"
 	ibcante "github.com/cosmos/ibc-go/v2/modules/core/ante"
+
+	gaiaante "github.com/cosmos/gaia/v6/ante"
 )
 
+// TODO: Move this to the root ante package after this is backported to 6.x.x
```
okwme, line 13: does this need to be backported? it would be state breaking so would only go into v8-Rho anyway, right?

okwme:
I like @anilCSE 's suggestion a lot, although I don't think the suggestion for governance enabled list of transactions is feasible unless it isn't enforceable and at that point it doesn't need to be gov-gated imo. Discussed with @alexanderbez as well and he said it wouldn't be difficult to do this way. 

To re-iterate the summary of what I think a perfect balance between enabling this feature but not opening ourselves up to un-retractable spam prevention would be:

add a config setting for validators that lists what msg types they allow into the mempool by skipping min fee in checktx. This is only to decide what messages make it into the mempool. There is one caveat: transactions that have low gas price but high gas are limited by a max gas constant to prevent the most aggressive spam attack where a single transaction could fill an entire block's worth of gas while paying nothing because the price is lower than min price (this was already implemented by bez, very nice 😎).

this means that it's still possible for someone to spam by creating lots of normal gas but low/zero gas price transactions. But should this happen and disrupt validator's income enough to make them notice, we'll be able to give them instructions on how to remedy the problem: change their config files. By making the IBC messages the defaults set in their config files, the new feature will be enabled, but not lock anyone in to accepting the free transactions should an attack happen.

Something I didn't realize before was that this is not state breaking so can be added into v7.0.1 after the v7.0.0 release (hopefully happening tomorrow). It means Bez would redo this PR branched from `Theta-main` and target `Theta-main` and add the config settings.

One question with regard to the release, @alexanderbez , you said this is API breaking but not state breaking. Does that mean it should actually be a v7.1.0 release? I've always seen minor releases treated as state breaking. semver.org describes minor releases as "when you add functionality in a backwards compatible manner". Woudl the config file be backwards compatible? Would validators who start running v7.0.1 need to manually add the config setting for it to work? Or can defaults come from the code instead of the generation event of the config file itself?

alexanderbez:
Replaced by: https://github.com/cosmos/gaia/pull/1352

PR #1352, opened 2022-03-23
feat: bypass min fee check for IBC relayer txs (Theta)

alexanderbez:
Replaces https://github.com/cosmos/gaia/pull/1343 with the caveat that we support config-controlled message types.


cmd/gaiad/cmd/root.go
```diff
@@ -77,21 +79,18 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
...
-	GaiaAppTemplate := serverconfig.DefaultConfigTemplate
-
-	return GaiaAppTemplate, GaiaAppCfg
+	return params.CustomConfigTemplate, params.CustomAppConfig{
+		Config: *srvCfg,
+		BypassMinFeeMsgTypes: []string{
+			sdk.MsgTypeURL(&ibcchanneltypes.MsgRecvPacket{}),
+			sdk.MsgTypeURL(&ibcchanneltypes.MsgAcknowledgement{}),
+			sdk.MsgTypeURL(&ibcclienttypes.MsgUpdateClient{}),
+		},
```
alexanderbez, line 92: These are what Gaia will set by **DEFAULT**.

alexanderbez (commented):
@yaruwangway @okwme This is essentially good-to-go. I have NOT tested this, meaning I have not tested the custom app configuration part.

To test this, we just need to spin up a single node and make sure we capture the configuration. E.g. we can just add a temporary print line in `app.go` 👍

tac0turtle (commented):
can we add this to the cosmos-sdk, this feature is amazing!

robert-zaremba (approved):

app/app.go
```diff
@@ -602,16 +603,17 @@ func NewGaiaApp(
...
 			HandlerOptions: ante.HandlerOptions{
 				AccountKeeper:   app.AccountKeeper,
 				BankKeeper:      app.BankKeeper,
 				FeegrantKeeper:  app.FeeGrantKeeper,
 				SignModeHandler: encodingConfig.TxConfig.SignModeHandler(),
 				SigGasConsumer:  ante.DefaultSigVerificationGasConsumer,
 			},
-			IBCkeeper: app.IBCKeeper,
+			IBCkeeper:            app.IBCKeeper,
+			BypassMinFeeMsgTypes: cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)),
```
okwme, line 616: is it possible to tell with `appOpts.Get` whether the option exists or was just left empty?
If I understand correctly, if the config has a value for `BypassMinFeeMsgTypesKey`, we accept the values provided. But I'm afraid we'll be in the same situation we are now that each validator will need to add this value to their config, if they don't run `gaiad init` (which they wouldn't normally run during an upgrade right?)

new nodes spinning up after this update would have the default values generated but what about if a someone just updates their node binary but keeps the previous config? We would need to notify all validators to update their configs to include the default messages otherwise, right?

alexanderbez: No, there are default values set. See `cmd/gaiad/cmd/root.go`. If the `app.toml` doesn't exist, it'll create it with default values.

PR #1355, opened 2022-03-24
Update testnet docs post upgrade

glnro:
This PR serves to update the `joining testnet` gaia docs with up to date information on the theta release candidate post upgrade


nooomski:
Lgtm, merging!
But moving forward, I would suggest we follow Udit's recommendation about keeping this agnostic to Gaia version and just refer people to detailed instructions in the testnet repo, to avoid maintaining duplicate information.

PR #1371, opened 2022-03-29
change "greeks from finance"

hxrts:
not sure why it said "greeks from finance" but they only use a couple letters so we will run out immediately


okwme:
We still have room with the following finance greeks:
* Lambda
* Epsilon
* Gamma
* Vanna
* Charm
* Vomma
* Veta
* Vera
* Speed
* Zomma
* Color
* Ultima

That's about 3 years of upgrades left

PR #1378, opened 2022-04-01
automate release 

yaruwangway:
automate the release process:
a github action will be triggered by tagging version. 
This action will generate a draft release note:
- [x] create auto- generate and upload the binaries of different os/arch, build_report, zip file to the release.
- [x] generate release note from changelog and build_report.
example pic:
Screenshot 2022-04-12 at 14 03 05

please still check carefully before click publish the release.


.github/workflows/release.yml
```diff
@@ -0,0 +1,90 @@
+name: create release
+
+on:
+    push:
+        tags:
+            - 'v*'
```
okwme: Any reason not to use the way Bez had it?
```suggestion
            - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,90 @@
...
+
+jobs:
+    build:
+        name: Create Release
+        runs-on: ubuntu-latest
+        steps:
+            -   name: Checkout code
+                uses: actions/checkout@v2
```
okwme: bez was using `checkout@v3` is there a reason to use v2?

okwme:
This looks good!
how is it being tested though?

okwme (approved):
couple questions but seems like the PR works as intended, just wondering how to confirm?

tokamak-git:
@okwme I guess we could cut a release for pooltoy, or a SIM app in hub-eng to test this rite? @yaruwangway WDYT?

yaruwangway:
> @okwme I guess we could cut a release for pooltoy, or a SIM app in hub-eng to test this rite? @yaruwangway WDYT?

i tested in my fork before. we can also test in hub-eng.

yaruwangway:
> This looks good! how is it being tested though?

i tested in my fork before

tokamak-git (approved):
Approved

PR #1388, opened 2022-04-06
WIP: Upgrade proposal generation

yaruwangway:
upgrade proposal generation will be run by github action triggered by tag
- [ ] put the generated proposal in this repo ?
- [ ] automatically upload the artifacts when doing release note trigged by tag


tokamak-git:
@yaruwangway This is prioritized?

mmulji-ic:
@yaruwangway the linter's complaining, we should fix those issues first - even though its on contrib.

okwme:
having trouble finding where the theta upgrade proposal.json was stored but noam pointed out this directory in gaia: https://github.com/cosmos/gaia/tree/main/docs/governance/proposals

seems like the right place to store this!

yaruwangway:
> having trouble finding where the theta upgrade proposal.json was stored but noam pointed out this directory in gaia: https://github.com/cosmos/gaia/tree/main/docs/governance/proposals
> 
> seems like the right place to store this!

so this generate upgrade proposal is only generating the cli of submitting upgrade proposal, mainly for avoid manually cp the binary links and shasum. in order to generate the proposal cli, i this action need us manually fill in a json to indicate upgrade name etc. 

this pr is waiting for a decision where to put the generated proposal. because this will be run on github server, which means the generated proposal cli is on server , unless we commit again to a branch or directly print it out, otherwise, we cannot get it. 

other ways of doing it is to not use github aciton to trigger this action, but manually `go run` to generate proposal, this way, we can generate the proposal cli locally without considering moving the proposal from github server to some where we can see.

okwme:
let's just print to the github action to keep it simple. 

limit to releases but it's OK if we just go to the CI and copy and paste the json we want

yaruwangway:
hi how about run generating proposal locally and manually after release, will add to makefile, otherwise, we have to commit a json file to release branch as well, which might be a "noise" doc. 
also, by the moment we release, we might not want to generate a proposal, only later when preparation work is done.

the proposal json file looks like
```
{
  "upgrade_name": "v7-Theta",
  "upgrade_version": "v7.0.0",
  "height": "-1",
  "proposer": "dev",
  "description": " "
}
```

yaruwangway:
move to https://github.com/interchainio/hub-eng

PR #1392, opened 2022-04-10
expose GRPC and REST endpoints from a single docker testnet node

cmwaters:
Small UX improvement. Have the GRPC (9090) and REST (1317) ports exposed on a single node so developers doing local integration work can automatically hit these endpoints.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,6 +6,8 @@ services:
     image: "tendermint/gaiadnode"
     ports:
       - "26656-26657:26656-26657"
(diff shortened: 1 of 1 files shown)


glnro (approved):

glnro (commented):
Seems logical, although I believe this is an out of date image, so I don't know how useful this is without additional changes

cmwaters:
Hmm yeah we might want to update some dockerhub image somewhere but this change was already useful for me for `make localnet-start` and `make localnet-stop`

Issue #916, opened 2021-07-19
Docker for gaia 5.x.x


anhcao142:
Hi, the [tendermint/gaia](https://hub.docker.com/r/tendermint/gaia) docker hub seems to stop at version 4.2.1. Do we have any plan to continue adding new versions to the hub?

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

tac0turtle:
looks like it was removed on purpose https://github.com/cosmos/gaia/pull/850. I am wondering why as well. @okwme could we bring it back? this is a major regression IMO

ghost:
Please add Dockerfile for 5.x version?

nooomski:
Closed by bot, opening again, should we also add a dockerfile for 6.x ? @okwme wdyt?

tokamak-git:
@mmulji-ic could you please help me with updating a secret please? have updated the vault, `SLACK_WEBHOOK_URL`
thanks.

tokamak-git:
@mmulji-ic getting following error, looks like my token does not have enough permissions to push to ghcr.

 ```
 #16 exporting to image
#16 pushing layers 0.3s done
#16 ERROR: server message: insufficient_scope: authorization failed

#20 [auth] cosmos/gaia:pull,push token for registry-1.docker.io
#20 DONE 0.0s

#21 [auth] cosmos/gaia:pull,push token for registry-1.docker.io
#21 DONE 0.0s
------
 > exporting to image:
------
error: failed to solve: server message: insufficient_scope: authorization failed
Error: buildx failed with: error: failed to solve: server message: insufficient_scope: authorization failed
```

tokamak-git:
Blocked by need to create secret environment.
 
cc  @okwme @mmulji-ic

PR #1399, opened 2022-04-11
Feature/adding docker GitHub action

tokamak-git:
issue: https://github.com/cosmos/gaia/issues/916

This pr enables the creation and publishing of official gaia images to ghcr (github container registry)

 have added the details regarding the triggers for the action to the readme.


faddat:
this will work, when merged

:/

ghcr is like that sadly

tokamak-git:
closes #916

tokamak-git:
> this will work, when merged
> 
> :/
> 
> ghcr is like that sadly

@faddat turns out it was a different how GitHub actions parses strings vs act the utility to run actions.

.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,73 @@
+name: Docker - ghcr Build and Push
+
+on:
+  push:
+    schedule:
+      - cron: "0 10 * * *"
```
alexanderbez, line 6: Why is there a need for a cron? Should pushes to release branches suffice?

Makefile
```diff
@@ -197,7 +197,7 @@ endif
 .PHONY: run-tests $(TEST_TARGETS)
 
 docker-build-debug:
```
alexanderbez, line 199: nit: undo. This dockerfile is solely for the e2e tests. It shouldn't be the image we push and use for operators on mainnet.

tests/e2e/genesis.go
```diff
@@ -67,7 +67,10 @@ func addGenesisAccount(path, moniker, amountStr string, accAddr sdk.AccAddress)
 	}
 
 	if accs.Contains(accAddr) {
-		return fmt.Errorf("failed to add account to genesis state; account already exists: %s", accAddr)
+		return fmt.Errorf(
```
alexanderbez, line 70: was this causing lint failure?

.github/workflows/docker-push.yml
tokamak-git: True, the pushes to release and tagging would suffice moving forward, but how do we trigger the previous images?
The solution  i came up with was using cron flag to trigger a job run that would pickup old tags and release branches.
At least that's how I hope it works.

tokamak-git: we could remove it of course once legacy images are built

Makefile
tokamak-git: Wouldn't we want to run our tests in an environment as close to production as possible?
I can revert the change and create a dedicated Dockerfile. I have one somewhere. But what pitfalls do you see of us using the same image?

tac0turtle:
Is help needed to get this over the finish line?

tokamak-git:
incorrect merge branch, closing

tokamak-git:
closing in favour of #1552

PR #1400, opened 2022-04-12
Update docs with v7

glnro:
This PR should have been waiting in the wings to be merged after the upgrade...better late than never.


nooomski (approved):
LGTM

mmulji-ic (approved):

PR #1413, opened 2022-04-15
Gaia v7 with pebbledb support

faddat:
This is a **draft** PR for review, of Gaia with PebbleDB. 

I am testing against Gaia, because I am having some issues that I suspect may be caused by fast IAVL, so currently doing the same thing for Gaia, that I am doing for Osmoisis, specifically, attempting a sync from genesis.  My early findings with Pebble are that it outperforms rocks, while avoiding the troubles of CGO.  Both the valnodes team and the notional team have contributed to the PebbleDB implmeentation, in an attempt to get a more rigorous understanding of database performance in cosmos.


faddat:
There are two related pull requests:

[Rocksdb v7 with grocksdb,  deprecating cosmos/gorocksdb](https://github.com/tendermint/tm-db/pull/229)
[PebbleDB (draft implementation)](https://github.com/tendermint/tm-db/pull/231)

okwme:
thanks @faddat , i think the choice to switch db should be considered at the level of the SDK rather on a per-app basis. What do you think about moving this to https://github.com/cosmos/cosmos-sdk?

faddat:
Billy, unfortunately, this PR is left over from an effort that has failed.

Thanks for your input but I'll need to close this one.  I'll continue to work on pebble as best I can.

PR #1418, opened 2022-04-17
Update docs to point to correct file location

k-yang:
The networks directory was moved to contrib/testnets in https://github.com/cosmos/gaia/pull/992.


jgimeno (approved):

okwme:
Thanks @k-yang !
We still need to update this section of the docs to point to https://github.com/cosmos/testnets where more up to date testnet information lives

PR #1447, opened 2022-04-25
feat: bypass min fee check for custom message types

alexanderbez:
replaces: https://github.com/cosmos/gaia/pull/1352

Support custom message types to bypass minimum fee checks for during `CheckTx`. By default, the following message types are set:

- `MsgRecvPacket`
- `MsgAcknowledgement`
- `MsgUpdateClient`


alexanderbez:
`Check Markdown links / markdown-link-check` failure is unrelated to this PR (should be fixed on `main`)

ante/fee.go
```diff
@@ -0,0 +1,75 @@
...
+func (mfd MempoolFeeDecorator) bypassMinFeeMsgs(msgs []sdk.Msg) bool {
+	for _, msg := range msgs {
+		if tmstrings.StringInSlice(sdk.MsgTypeURL(msg), mfd.BypassMinFeeMsgTypes) {
+			continue
+		}
+
+		return false
+	}
+
+	return true
```
yaruwangway, line 74: is it possible this happens: `msgs` is empty, so that `bypassMinFeeMsgs([]sdk.Msg{})` will return true ?

alexanderbez: Having a tx w/o any message is a no-op really, I don't think it's even a valid case.

yaruwangway (approved):

PR #1448, opened 2022-04-25
change linter settings

faddat:
* In order to make maintaining gaia easier, stop checking diffs for CI and just run the linter each time instead
* Remove deprecated linters

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,12 +14,6 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: 1.17
-      - uses: technote-space/get-diff-action@v6.0.1
-        with:
-          PATTERNS: |
-            **/**.go
-            go.mod
-            go.sum
+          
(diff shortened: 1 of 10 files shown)


faddat:
If we merge this, then it can be backported to the v6.x and v5.x releases

As for v4.x, I do have some concerns about how bumping sdk and tendermint broke apphash

faddat:
https://github.com/tendermint/tm-db/runs/6163210118?check_suite_focus=true#step:5:30

cosmos codecov generally broken

okwme:
hey @faddat , what would be the benefit of running all CI jobs on PRs that have no changes relevant to what the CI checks?
I think actually it woudl be good to do the opposite for instance not run simulation tests on PRs that only update docs.
otherwise we're waiting a long time for each PR to run tests that might not be relevant at all to the changes (and eats up CI minutes on top).
can you elaborate on what's improved on testing with your suggestions?

okwme:
gonna close this since no response

Issue #1244, opened 2022-02-09
DOCS: Deploy Testnet needs an update


nooomski:
I was checking the broken links in the Check Markdown Link and found that [Deploy Your Own Gaia Testnet](https://github.com/cosmos/gaia/blob/main/docs/hub-tutorials/deploy-testnet.md) is heavily outdated as its still relying on the supporting code in the now missing "networks directory".

We can use this issue to discuss what we want to achieve with this article and hopefully provide a revamped article (or decide to remove it all-together).

pankaj0812:
Hi @nooomski . I was trying to setup a MultiNode Local Automated testnetwork following the docs, and found an issue with the Makefile which relies on the networks/ directory, which currently has been moved to contrib/ in the main branch. 

In order to get the setup to work, we probably need to change `
	$(MAKE) -C networks/local` to 
`$(MAKE) -C contrib/testnets/local`. I have raised a PR for the same #1349. Please have a look.

okwme:
as part of this we should move the other devnet documentation into this tutorial as a larger devnet tutorial.

similarly the migration tutorials should be moved to the mainnet repo.

nooomski:
Closed by #1466

PR #1456, opened 2022-04-26
refactor: liveness tests

alexanderbez:
- Remove old legacy docker-compose-based liveness setup and tests in favor of a more streamlined use of ignite CLI.
- Fix CI workflows for `Build & Test`


okwme:
ah just realized this is failing lint chcker cause you removed all the `deploy-testnet.md` docs. This is realted to this issue @alexanderbez : https://github.com/cosmos/gaia/issues/1244
Do you have an opinion about the contents of the `contrib/testnets` directory Bez? Should we just nuke all of it or is any of it still used / useful in light of https://github.com/cosmos/testnets?

.github/workflows/test.yml
```diff
@@ -70,11 +63,42 @@ jobs:
...
+        run: |
+          curl https://get.ignite.com/cli@v0.20.0! | bash
+        if: env.GIT_DIFF
+      - name: Start Local Network via Ignite CLI
+        run: |
+          make start-localnet-ci > ignite.out 2>&1 &
+        if: env.GIT_DIFF
+      - name: Test Local Network Liveness
+        run: |
+          sleep 3m
```
yaruwangway, line 102: why sleep ? and sleep so long ?

contrib/scripts/test_localnet_liveness.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ -z "$4" ]; then
+  echo "Invalid argument: missing node address"
+  echo "sh test_localnet_liveness.sh <iterations> <sleep> <num-blocks> <node-address>"
+  exit 1
+fi
+
+docker_containers=($(docker ps -q -f name=umeed --format='{{.Names}}'))
+
+while [ ${CNT} -lt $ITER ]; do
```
yaruwangway, line 35: is `$ITER` used for only counting a max looping ? if this way, can  we just write `while [ ${CNT} -lt $(expr ${NUMBLOCKS} + 50) ]; do` then we do not need to pass the arg `ITER=$1`

.github/workflows/test.yml
alexanderbez: `ignite` start a blocking process where it runs a chain, so we run the process in the background and pipe both stdout and stderr to a file. `ignite` takes a bit to startup (builds app, builds proto, etc...), so the 3m is just a safe "buffer" to wait for `ignite` to actually start the chain.

contrib/scripts/test_localnet_liveness.sh
alexanderbez: We've had this script in the SDK for years at this point. We could probably clean it up. I'm indifferent to it as it's pretty simple.

yaruwangway (approved):

yaruwangway:
> ah just realized this is failing lint chcker cause you removed all the `deploy-testnet.md` docs. This is realted to this issue @alexanderbez : #1244 Do you have an opinion about the contents of the `contrib/testnets` directory Bez? Should we just nuke all of it or is any of it still used / useful in light of https://github.com/cosmos/testnets?

I thought it was accidentally removed, so i add the `deploy-testnet.md` back to `docs/hub-tutorials`

Issue #704, opened 2021-02-27
Generalized bloat


faddat:
I'm of the opinion that economic growth, along with with DEX implementation & IBC is going to lead to a scenario I've named "generalized bloat".  It was originally named something else, but we will no longer refer to the problem this way.

https://forum.cosmos.network/t/proposal-draft-gaia-v8-rho-upgrade/6570/4

The Generalized bloat situation is going to at least affect validators and exchanges, but will also likely touch on other matters, as well.  

## Generalized bloat

* https://github.com/cosmos/gaia/issues/670
* https://github.com/cosmos/gaia/issues/669
* Validators are more expensive to operate
* Sync times are lengthened (though state sync takes care of this)
* Various resource requirements go up

The above issues are from only a moderately bloated Gaia: IIRC, the codebase was clocking in ~300GB at the time.  

After the DEX is added, I'm thinking that Gaia is going to get MUCH more bloated:  Terabytes

I was thinking out loud about this in the Tendermint engineering slack, and realized that I should publish my thoughts, because if we're not testing what 1TB+ Gaia looks like, we should be.  

I think it'd be a shame to strap on a best in class AMM system, and then get stuck because the load it puts on the network isn't manageable, smilar to ETH and Uniswap today.

tac0turtle:
This is a concern of mine as well. I am hoping with the upcoming work around reworking storage this problem will becomes less present. 

There is also ongoing work on the Tendermint side to make storage more efficient.

faddat:
Any thoughts on how to set up a (preferably automated) test of a bloated network?

One could likely run simapp, and eventually you'd get a bloated gaia, but ideally you'd want a real world network scenario, with all the strange things that tend to happen only on a live network.

faddat:
@shahankhatch this is making me think about system requirements for Gaia.  

So, she runs on a pi:

![Screen Shot 2021-03-21 at 5 09 57 PM](https://user-images.githubusercontent.com/7142025/111901037-455eba80-8a68-11eb-8c06-12f87f3d7b51.png)

but I suspect that pi has a bad microsd card.  Unfortunately, I have a tough time getting legit microSD cards here (Hanoi). The ones that I have that are known-good, don't suffer from this write problem and I've got another Gaia coming up on another SD card and I think I am going to smash this one to prevent it getting in my workflow again.

**anyway**

I am wondering how fast and furious we should expect state growth to be when the liquidity module is added to gaia.

faddat:
#817

[...]

PR #1463, opened 2022-05-01
Note database choices

faddat:
#704 can be addressed somewhat with different database choices.

This adds those to the state sync script to make it easier to use.


faddat:
This pull request provides guidance to users on how to use different databases when doing a state sync. It does literally nothing else.

tac0turtle (approved):

PR #1468, opened 2022-05-03
gofumpt -w -l

faddat:
Gorgeous code formatting

go install mvdan.cc/gofumpt@latest
gofumpt -w -l .


okwme:
thanks for the suggestion @faddat but we're going to stay consistent with SDK for now

faddat:
Billy the SDK uses gofumpt

https://github.com/cosmos/cosmos-sdk/pull/11839

Let's stay consistent with the SDK

PR #1470, opened 2022-05-03
WIP: update rho-base-main

yaruwangway:
this pr is for add changes from main to Rho main

PR #1467, opened 2022-05-02
WIP: add skip min fee to Rho

okwme:
Still need help on the end here. I'm not sure I understand how to convert `MempoolFeeDecorator` into a `tx.Middleware` type in order to add it to `NewDefaultTxHandler` in `middleware.go`.

PR #1436, opened 2022-04-19
WIP: Rho 

okwme:
replaces #1410

PR #1410, opened 2022-04-13
WIP: Rho base

okwme:
I took the work that Lido began on their [fork of wasmd](https://github.com/cosmos/wasmd) and started porting it over to gaia. I did this by pasting the entirety of each file from theirs to ours, then went through the git diff and added or tweaked as seemed resonable. Some `TODO`s still left to confirm certain things.

Updated the go.mod similarly and did a replace on `github.com/CosmWasm/wasmd` with `github.com/cosmos/wasmd` to use the latest [commit](74e01f79d8f43c7e3a94073d94468104787e5e10) from their [working branch](feat/wasmd-ibcv3-ls-module-alongside-cosmos-sdk). 

There are still a number of errors, but most of them seem to be around importing `liquidity` module and other sdk v0.45 discrepancies.

I started this with a new branch off of `Theta-main`, that might not have been right now that it's merged into main. Unfortunately when I target `main` there are over 100 files changed. Would rebasing main fix that?

PR #1478, opened 2022-05-11
Rho Base

okwme:
Started a new branch of gaia using Cosmos SDK v0.46.0-beta2 and Tendermint v0.35.2 but not including CosmWasm.

Includes new Gov and Groups modules.

Had to remove following modules which are not yet v0.46.0-beta2 compatible:
* github.com/cosmos/ibc-go/v3
* github.com/gravity-devs/liquidity
* github.com/strangelove-ventures/packet-forward-middleware/v2

Had to remove E2E due to changes in v0.46 that weren't immediately apparent (`tests/e2e/keys.go` was where it started)

Had to remove the new skip fee middleware until it's been converted from antehandler to middleware

Should replace #1436, #1410, #1470 and #1467


okwme:
~@alexanderbez you mentioned about the following line on the previous PR:~
```
	cdc := params.MakeTestEncodingConfig().Codec
```
> There should already be a private global for the codec -- let's use that instead of creating a new variable here.

~Can you explain how to access the global codec without using the new variable?~

nm i see it's just available as `cdc` :+1:

dongsam:
Hi @okwme @yaruwangway, 
We released liquidity  ~~v2.0.0-alpha2~~ [v2.0.0-beta1](https://github.com/Gravity-Devs/liquidity/releases/tag/v2.0.0-beta1) which is based on cosmos-sdk v0.46.0-beta2. 
The liquidity v2 final version will be released when the official cosmos-sdk v0.46.0 is released, and all other work is done
https://github.com/cosmos/gaia/pull/1482

okwme:
all tests passing except codecoverage, going to merge as it seems to be tracking commented lines as well

PR #1479, opened 2022-05-11
trying to fix e2e

okwme:


tests/e2e/keys.go
```diff
@@ -38,7 +37,7 @@ func createMemoryKey() (mnemonic string, info *keyring.Record, err error) {
 }
 
 func createMemoryKeyFromMnemonic(mnemonic string) (*keyring.Record, error) {
-	cdc := params.MakeTestEncodingConfig().Codec
+	// cdc := params.MakeTestEncodingConfig().Codec
```
okwme:
```suggestion
```

okwme:
so the keys issue has been upstreamed here but gaia is still failing to produce blocks.
not sure how to debug the logs from the docker images though, @alexanderbez do you have an idea how to approach this?

okwme:
merging this in to focus on the liveness tests as a separate PR

tests/e2e/validator.go
```diff
@@ -129,9 +131,8 @@ func (v *validator) createConsensusKey() error {
 }
 
 func (v *validator) createKeyFromMnemonic(name, mnemonic string) error {
-	cdc := params.MakeTestEncodingConfig().Codec
```
yaruwangway, line 132: here is `cdc := gaia.MakeTestEncodingConfig().Codec`

yaruwangway: see here https://github.com/cosmos/gaia/pull/1483/files

tests/e2e/e2e_setup_test.go
```diff
@@ -52,9 +51,9 @@ type IntegrationTestSuite struct {
 	valResources   map[string][]*dockertest.Resource
 }
 
-func TestIntegrationTestSuite(t *testing.T) {
-	suite.Run(t, new(IntegrationTestSuite))
-}
+//func TestIntegrationTestSuite(t *testing.T) {
```
yaruwangway: this is the entry point to test e2e, when i comment out, there is no test to run, if i only comment out ibc test, will also show no test to run. @okwme

tests/e2e/validator.go
okwme: @alexanderbez noted that this should be removed in light of a globally available `cdc`. Unless was this the culprit of the keys encoding issue?

tests/e2e/e2e_setup_test.go
okwme: are you suggesting i uncomment it or leave it as is?
i believe this change came from rebasing the `rho` branch with your changes, or?

yaruwangway: yeah, i should not comment out, if we still want to run e2e

tests/e2e/validator.go
yaruwangway: hi @okwme , i think yours is right!

tests/e2e/e2e_setup_test.go
glnro: Is this bc of IBC?

PR #1488, opened 2022-05-16
feat: global fee module

yaruwangway:
Global fee is from [tgrade/globalfee](https://github.com/confio/tgrade/tree/main/x/globalfee)
- [x] 1 new file is added: `x/globalfee/types/genesis.go` 
- [x] e2e tests is added.


Makefile
```diff
@@ -222,7 +222,7 @@ format:
 ###############################################################################
 
 start-localnet-ci:
-	@ignite chain serve --reset-once -v -c ./ignite.ci.yml
+	@ignite chain serve --mode validator --reset-once -v -c ./ignite.ci.yml
```
mmulji-ic: remember to revert this

okwme:
need to confirm the flow:
* should we continue to use app.toml min fee but only allow it to be valid when it is above the global min fee?
* or should we get rid of app.toml min fee entirely?

Feels like the former is what we want so that when we launch and have the default value to 0uatom as the min fee, the network will continue operating as normal with all the previous settings. Then we can make a gov proposal if we want to raise it above 0uatom so all operators are aware of the changes.

if the app.toml min fee is missing the denom of the current tx, the global fee should be the only check that is used.

Makefile
okwme: done

proto/gaia/globalfee/v1beta1/genesis.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+package gaia.globalfee.v1beta1;
+
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+// import "cosmwasm/wasm/v1/genesis.proto";
```
okwme:
```suggestion
```

okwme (approved):

Issue #1449, opened 2022-04-25
CI settings prevent updates to older gaias


faddat:
## Summary of Bug

* https://github.com/cosmos/gaia/runs/6164055056?check_suite_focus=true

## Version

all

## Steps to Reproduce

Update older versions of gaia, or in the case of the codecov issue, do anything at all

https://github.com/cosmos/gaia/runs/6164055056?check_suite_focus=true#step:8:30

## Suggested solutions

* Stop using the diff ci action
* consider replacing codecov with a more portable solution that does not rely on an api key
* backport this to all gaias

https://github.com/cosmos/gaia/pull/1448

## note

due to the fact that there seems an issue with the cosmos codecov api key generally or with the whole service, any PR's that have cone in recently will need their ci jobs re-run

this has hit tm-db as well:

https://github.com/tendermint/tm-db/runs/6163210118?check_suite_focus=true

## 2nd note:
it hurts more than the linters
https://github.com/faddat/gaia/blob/84de16ef50b0dad6ec164b99610a80bf16d9ec01/.github/workflows/sims.yml#L125

ci jobs are free, why are we trying to conserve them?

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

faddat:
Like this:

https://github.com/cosmos/gaia/runs/6615907131?check_suite_focus=true#step:4:75

faddat:
yep.  this one, dated a month ago, detailed the issues with all the pr's

mmulji-ic:
Hey @faddat do have suggestions for alternative codecov tooling?

faddat:
Just had a call with @okwme -- so, we found that the linter was missing this line:

#1623 
#1624 

However, the main branch contains linting issues.  They're not being caught because of the diff.

They are grandfathered-in.  We saw this with:

* https://github.com/cosmos/cosmos-sdk/pull/12702
* https://github.com/cosmos/cosmos-sdk/pull/12141
* https://github.com/cosmos/ibc-go/pull/1418

okwme:
Linting errors should all be fixed at this point in main.
Since we removed checkdiff linters will be run all the time as well.
still needs to be backported to v7.0.x (can go into current work on v7.0.3) and v6.0.x according to #1549

okwme:
[...]

PR #1507, opened 2022-05-26
Fix error handling and fix linting

faddat:
This pr:

Fixes linting, we used obsolete linters.
Fixes linting, we had a diff check in place that made PR's to other versions of gaia fail by default because we would run the linters on the 
diffs. 
Fixes error handling in export.go, we did not check errors.

I'll now take those same changes and apply to v6.0.5 and since it won't run linters on diffs we won't have errors.  I honestly figured that the hub maintainers were aware of this.


faddat:
Oh, and the codecov report?

total red herring

this pr definitely improves gaia.

yaruwangway (approved):

okwme:
Hi @faddat thank you for the PR.
Can you clarify further, my understanding of the use of diff was so that we don't run linters on code that already had linters running on it.
you're saying that the diff setting is not having this effect and is instead resulting in linters being run on incomplete code and therefore incorrectly failing?
In general I would like to change our CI so that linting is not unnecessarily run since every PR takes so long to pass tests, even when the changes are just markdown files.

faddat:
Yes, if we use that diff thingy for linting, we do not end up linting historical code, and if we continue to use the diff thingy it causes matienance issues on older gaias.  As explained in #1449 -- the diff doodad it'll work if we've just got one release we're maintaining and otherwise it ... it is a straight path to not-good-times.

#1449 

golangci-lint does not add any significant time to the ci runs. 

It is run in parallel with the other linters, and since it is slower than the slow ones, is nbd.

faddat:
codecov is a red herring factory, see #1449

faddat:
Odd thing about response times.


https://github.com/cosmos/gaia/pull/1577

faddat:
this is just getting silly.

I've been elected to validate the hub by my delegators. 

Validators are supposed to check through the code.

I found that we were behind where we should be, linting-wise.  

I solved that. 

What is going on here guys?

Please don't lock this PR, it is the appropriate place for this conversation.

This is a security-enhancing PR from May 26.  It is July 10.

If you do not wish my contributions to hit this repository, **I understand** but I need for the cosmos hub team lead, @okwme  to say that so that I can inform my delegators.

tac0turtle (approved):
Lgtm. Let's keep the scope of prs small. Can we spin out the action changes to a different pr please

faddat:
sure thing

faddat:
I've removed github actions changes from this PR.

okwme:
thanks for the pr @faddat and thank you for the review @marbar3778 
merging now

PR #1493, opened 2022-05-19
feat: update x/bank blocked addresses 

alexanderbez:
Update the deny-list provided to `x/bank` to allow `x/gov` and `x/groups` s.t. that they can RECEIVE funds :)

Blocked on:

- [ ] https://github.com/cosmos/cosmos-sdk/pull/11998
- [ ] https://github.com/cosmos/gaia/pull/1484

PR #1531, opened 2022-06-12
E2E Testing New Gov Module

glnro:
This PR includes three new e2e tests to cover the major functionality of the sdk's new gov module in `v0.46` for the upcoming Rho upgrade.

##### `TestSendTokensFromNewGovAccount`
Moves tokens from the community pool to fund the gov account and then sends those tokens to a recipient address, submitting both legacy and new gov proposals.
**Subtests**
1. `submitLegacyProposalFundGovAccount`
2. `depositGovProposal`
3. `voteGovProposal`
4.  `submitNewGovProposal` executing `MsgSend`
5. `depositGovProposal`
6. `voteGovProposal`

##### `TestGovSoftwareUpgrade` 
Submits a proposal executing `MsgSoftwareUpgrade` and includes weighted voting.
**Subtests**
1. `submitNewGovProposal` executing `MsgSoftwareUpgrade`
2.  `depositGovProposal` 
3. `voteGovProposal` with `weighted-vote`

##### `TestGovCancelSoftwareUpgrade` 
Submits a proposal executing `MsgSoftwareUpgrade` and another executing `MsgCancelUpgrade`
**Subtests**
1. `submitNewGovProposal` executing `MsgSoftwareUpgrade`
2.  `depositGovProposal` 
3. `voteGovProposal`
4. `submitNewGovProposal` executing `MsgCancelUpgrade`
5. `depositGovProposal` 
6.  `voteGovProposal`

For `TestGovCancelSoftwareUpgrade` it might be worth combining messages to test multiple arbitrary message execution. 

Additionally, the work from #1493 is included in this PR to enable sending funds from the community pool to the gov account address.


app/app.go
```diff
@@ -834,6 +834,19 @@ func (app *GaiaApp) LoadHeight(height int64) error {
...
 }
 
+func (app *GaiaApp) BlockedModuleAccountAddrs() map[string]bool {
+	modAccAddrs := app.ModuleAccountAddrs()
+
+	// remove module accounts that are ALLOWED to received funds
+	//
+	// TODO: Blocked on https://github.com/cosmos/cosmos-sdk/pull/11998 getting
+	// into 0.46
+	// delete(modAccAddrs, authtypes.NewModuleAddress(grouptypes.ModuleName).String())
```
okwme: this says blocked on 11998 but that's regarding invariant checks for gov module. you've got gov module listed below but the groups module is commented out. is the todo related to gov or groups? if groups then the PR shouldn't matter, if gov, then does this current PR work but just at risk of violating an invariant until the 11998 is merged?

glnro: I'm not sure, this was copied over for the time being from #1493 until @alexanderbez 's PR was merged.

tests/e2e/e2e_setup_test.go
```diff
@@ -30,16 +33,40 @@ import (
...
+	govSendMsgRecipientAddress = "cosmos1pkueemdeps77dwrqma03pwqk93nw39nuhccz02"
 )
 
 var (
-	stakeAmount, _  = sdk.NewIntFromString("100000000000")
-	stakeAmountCoin = sdk.NewCoin("stake", stakeAmount)
+	stakeAmount, _    = sdk.NewIntFromString("100000000000")
+	stakeAmountCoin   = sdk.NewCoin("stake", stakeAmount)
+	tokenAmount       = sdk.NewInt64Coin(photonDenom, 3300000000) // 3,300photon
+	fees              = sdk.NewInt64Coin(photonDenom, 330000)
```
okwme:
```suggestion
	fees              = sdk.NewInt64Coin(photonDenom, 330000) // 0.33photon
```

tests/e2e/e2e_setup_test.go
```diff
@@ -30,16 +33,40 @@ import (
...
+	depositAmount     = sdk.NewInt64Coin(photonDenom, 10000000).String()
```
okwme:
```suggestion
	depositAmount     = sdk.NewInt64Coin(photonDenom, 10000000).String() // 10photon
```

tests/e2e/e2e_setup_test.go
```diff
@@ -423,3 +452,108 @@ func noRestart(config *docker.HostConfig) {
...
+		},
+	}
+
+	msgSendBody, err := json.MarshalIndent(struct {
+		Messages []GovMessageSend `json:"messages"`
+		Metadata string           `json:"metadata"`
+		Deposit  string           `json:"deposit"`
+	}{
+		Messages: msgSendMessages,
+		Metadata: "VGVzdGluZyAxLCAyLCAzIQ==",
```
okwme:
```suggestion
		Metadata:     b64.StdEncoding.EncodeToString([]byte("Testing 1, 2, 3!")),
```
or should this be `b64.URLEncoding.EncodeToString([]byte(data))` for URL Encoding?

Did you confirm that the metadata needs to be Base64 encoded before being submitted?

tests/e2e/e2e_setup_test.go
```diff
@@ -423,3 +452,108 @@ func noRestart(config *docker.HostConfig) {
...
+		Title:       "Community Pool Spend",
+		Description: "Fund Gov !",
+		Recipient:   govModuleAddress,
+		Amount:      "1000photon",
+		Deposit:     "5000photon",
+	}, "", " ")
+
+	s.Require().NoError(err)
+
+	for _, val := range c.validators {
```
okwme, line 502: why is the file written once for each validator?

tests/e2e/e2e_setup_test.go
```diff
@@ -423,3 +452,108 @@ func noRestart(config *docker.HostConfig) {
...
+		},
+	}
+
+	upgradeProposalBody, err := json.MarshalIndent(struct {
+		Messages []SoftwareUpgrade `json:"messages"`
+		Metadata string            `json:"metadata"`
+		Deposit  string            `json:"deposit"`
+	}{
+		Messages: softwareUpgradeMessages,
+		Metadata: "VGVzdGluZyAxLCAyLCAzIQ==",
```
okwme: same question about base64 format from last one

tests/e2e/e2e_test.go
```diff
@@ -54,7 +55,7 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
 
 func (s *IntegrationTestSuite) TestBankTokenTransfer() {
 	s.Run("send_photon_between_accounts", func() {
-
+		s.T().Skip()
```
okwme: why is this being skipped now?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+
+		s.Require().Eventually(
+			func() bool {
+				afterDistPhotonBalance, err := getSpecificBalance(chainAAPIEndpoint, distModuleAddress, tokenAmount.Denom)
+				if err != nil {
+					s.T().Logf("Error getting balance: %s", afterDistPhotonBalance)
+				}
+				s.Require().NoError(err)
+
+				return afterDistPhotonBalance.IsValid()
```
okwme: shouldn't the balance be previous balance + tokenAmount to confirm it was received?
just checking isValid means that pool has any balance right?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+	s.fundCommunityPool(chainAAPIEndpoint, sender)
+
+	s.T().Logf("Submitting Legacy Gov Proposal: Community Spend Funding Gov Module")
+	s.submitLegacyProposalFundGovAccount(chainAAPIEndpoint, sender)
+	s.T().Logf("Depositing Legacy Gov Proposal: Community Spend Funding Gov Module")
+	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
+	s.T().Logf("Voting Legacy Gov Proposal: Community Spend Funding Gov Module")
+	s.voteGovProposal(chainAAPIEndpoint, sender, proposalCounter, "yes", false)
+
+	initialGovBalance, err := getSpecificBalance(chainAAPIEndpoint, govModuleAddress, photonDenom)
```
okwme, line 130: should this check to confirm that the `initialGovBalance` is equal to the `spendAmount`?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+
+	s.T().Logf("Submitting Gov Proposal: Sending Tokens from Gov Module to Recipient")
+	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/root/.gaia/config/proposal_2.json")
+	s.T().Logf("Depositing Gov Proposal: Sending Tokens from Gov Module to Recipient")
+	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
+	s.T().Logf("Voting Gov Proposal: Sending Tokens from Gov Module to Recipient")
+	s.voteGovProposal(chainAAPIEndpoint, sender, proposalCounter, "yes", false)
+	s.Require().Eventually(
+		func() bool {
+			sendGovAmount := sdk.NewInt64Coin(photonDenom, 10)
```
okwme: should this 10 be a variable so if it changes in one place it updates in both?

okwme: isn't it included in the gov module send msg as well?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+
+}
+
+func (s *IntegrationTestSuite) TestGovSoftwareUpgrade() {
+	chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
+	senderAddress, err := s.chainA.validators[0].keyInfo.GetAddress()
+	s.Require().NoError(err)
+	sender := senderAddress.String()
+	height := s.getLatestBlockHeight(s.chainA, 0)
+	proposalHeight := height + 35
```
okwme: maybe a note here why 35 is significant?
did the number come about because it is enough blocks that we can assume the proposal has passed but not enough for the proposal to be executed?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+	proposalCounter++
```
okwme, line 162: shouldn't this be increased only after the proposal has been created?

okwme: i see it has to be incremented beforehand in order to query it.

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+		5*time.Second,
+	)
+
+	counter := 0
+	s.Require().Eventually(
+		func() bool {
+			currentHeight := s.getLatestBlockHeight(c, valIdx)
+
+			if currentHeight > upgradeHeight {
+				panic("Chain did not halt at upgrade height")
```
okwme: should this panic or return false until the timer runs out forcing the test framework to register the error? the message could be returned as the fail mode of the actual test right?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
+	s.T().Logf("Weighted Voting Gov Proposal: Software Upgrade")
+	s.voteGovProposal(chainAAPIEndpoint, sender, proposalCounter, "yes=0.8,no=0.1,abstain=0.05,no_with_veto=0.05", true)
+
+	s.verifyChainHaltedAtUpgradeHeight(s.chainA, 0, proposalHeight)
+	s.T().Logf("Successfully halted chain at height %d", proposalHeight)
+
+	currentChain := s.chainA
+
+	for valIdx := range currentChain.validators {
```
okwme, line 179: not sure if we have to delete all the validators and re-launch them, shouldn't we just be able to tell them to start running agin? Maybe there's a benefit to starting from scratch? In another scenario, if anyone were to join the network and replay from 0, they would be forced to halt at the upgrade height and restart themselves but this won't happen here right? Is there any state from earlier transactions we want to keep? Maybe it's ok to nuke em and keep the tests running, just might be confusing if anyone adds an earlier test that expects that state to persist past the governance tests.

tests/e2e/e2e_test.go
```diff
@@ -111,3 +109,257 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+			h := s.getLatestBlockHeight(s.chainA, 0)
+			s.Require().NoError(err)
+
+			return (h > 0)
+		},
+		30*time.Second,
+		5*time.Second,
+	)
+
+	proposalCounter = 0
```
okwme, line 201: if we revert to nodes just starting up again instead of starting suite from scratch we should preserve the old proposal counter

tests/e2e/e2e_setup_test.go
glnro: It doesn't need to be, I can move it in the setup for each test, I did this in the beginning when I was troubleshooting

tests/e2e/e2e_test.go
glnro: what do you mean?

tests/e2e/e2e_test.go
glnro: Yeah I can abstract this to a constant to explain, 35 ended up being the safest number of blocks to guarantee that everything could be processed in enough time.

tests/e2e/e2e_test.go
okwme: discussed out of band with lauren, just wanted to update here that i think it's better to keep as nuking them afterall. it preserves the thinking that the tests should in fact be able to run independent of each other. and it's a little sloppy stopping and starting the chain again with the `--skip-upgrade` flag.

tests/e2e/e2e_test.go
okwme: in `writeGovProposals ` (https://github.com/cosmos/gaia/pull/1531/files#diff-7d53e76218ae6e6128b32ace0c99cf129d5f294a3aac6a6ddb9f115bc85cece2R457) it has the same 10 photon being sent. Aren't these numbers dependent on each other? If the first one were changed to 5 this test would fail because of an insufficient balance right?

tests/e2e/e2e_cli_util.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return amt, err
+	}
+	for _, c := range balances {
+		if strings.Contains(c.Denom, denom) {
+			amt = c
+			break
+		}
+	}
+	return
```
yaruwangway: `return amt, nil`

tests/e2e/e2e_test.go
glnro: Yes you're right, we should pull this out into a single constant

tests/e2e/e2e_test.go
glnro: Ah I left that in when I was trying to isolate a single test I'll bring it back

tests/e2e/e2e_setup_test.go
glnro: When you run 
```shell
$gaiad tx gov submit-proposal --help

Submit a proposal along with some messages, metadata and deposit.
They should be defined in a JSON file.

Example:
$ gaiad tx gov submit-proposal path/to/proposal.json

Where proposal.json contains:

{
  // array of proto-JSON-encoded sdk.Msgs
  "messages": [
    {
      "@type": "/cosmos.bank.v1beta1.MsgSend",
      "from_address": "cosmos1...",
      "to_address": "cosmos1...",
      "amount":[{"denom": "stake","amount": "10"}]
    }
  ],
  "metadata: "4pIMOgIGx1vZGU=", // base64-encoded metadata
  "deposit": "10stake"
}

```

The cli docs suggest metadata should be base64 encoded, but there's no validation or enforcement. For the sake of compliance we could encode the original string in the test instead of hardcoding the encoded result.

tests/e2e/e2e_test.go
glnro: So the issue is if the test is run in isolation or first, it will have no initial balance. We could check for that case and set it to 0 in the test. We can never assume it will always be `0photon` as the community pool will show a different balance depending on how many tests were run before. See:

Run in isolation
```
=== RUN   TestIntegrationTestSuite/TestSendTokensFromNewGovAccount/fund_community_pool
    e2e_test.go:236: Token Amount: 3300000000photon
    e2e_util_test.go:181: Executing gaiad tx distribution fund-community-pool on chain chain-zeafQW
    e2e_util_test.go:198: Successfully funded community pool
    e2e_test.go:248: After Balance: 3300330000photon
    e2e_test.go:251: Send Amount: 3300000000photon
    e2e_test.go:252: Difference: 330000photon
```

Run after several gov tests
```
=== RUN   TestIntegrationTestSuite/TestSendTokensFromNewGovAccount/fund_community_pool
    e2e_test.go:237: Token Amount: 3300000000photon
    e2e_util_test.go:181: Executing gaiad tx distribution fund-community-pool on chain chain-8jbkQ3
    e2e_util_test.go:198: Successfully funded community pool
    e2e_test.go:249: After Balance: 399300660000photon
    e2e_test.go:252: Send Amount: 3300000000photon
    e2e_test.go:253: Difference: 396000660000photon
```

tests/e2e/e2e_test.go
glnro: I feel like this line covers that https://github.com/cosmos/gaia/pull/1531/files#diff-7d1fc1e232b55677582b51e2623aae6d8c15b5f0960ac6100fc7d733cb795dacR148

tests/e2e/e2e_setup_test.go
okwme: cool then let's keep it base64 encoded but update to include the code for encoding for clarity 👍 
the snippet i provided above should work, just need to add `b64` to the imports

tests/e2e/e2e_test.go
okwme: yup, sorry don't know how i missed that

tests/e2e/e2e_setup_test.go
okwme: oh great you already did : )

okwme:
NICE WORK 🚀

PR #1532, opened 2022-06-13
add: globalfee e2e test

yaruwangway:
This pr is to add global fee e2e tests

This only tested one case:

when global fee has two fees:
1. `photon` which is higher than `min_gas_price` in `app.toml`
2. `stake` which is not in  `min_gas_price` in `app.toml`

we send a bank send with a fee of stake fee that is higher than global fee and photon fee that is lower than glob fee, equal or higher than min_gas_price in app.toml. this tx should pass.


tests/e2e/e2e_test.go
```diff
@@ -111,3 +111,76 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
 	})
 }
+
+/*
+bank send with fees two fees.
+The following test is for the case that:
+photon fee: higher than or equal to min_gas_price in app.toml, but lower than the photon fee required in globalfee.
+stake fee: higher than stake fee required in globalfee.
+*/
+func (s *IntegrationTestSuite) TestGlobFee() {
```
glnro:
```suggestion
func (s *IntegrationTestSuite) TestGlobalFee() {
```

tests/e2e/e2e_test.go
```diff
@@ -111,3 +111,76 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
 		)
 	})
 }
+
+/*
+bank send with fees two fees.
+The following test is for the case that:
+photon fee: higher than or equal to min_gas_price in app.toml, but lower than the photon fee required in globalfee.
+stake fee: higher than stake fee required in globalfee.
```
glnro, line 377: So to clarify, as long as the transaction satisfies one of the global fee denoms, it will execute?

tests/e2e/e2e_test.go
```diff
@@ -111,3 +111,76 @@ func (s *IntegrationTestSuite) TestBankTokenTransfer() {
...
+*/
+func (s *IntegrationTestSuite) TestGlobFee() {
+	s.Run("send_tokens_to_test_globalfee", func() {
```
glnro, line 380: To be on the safe side it might be a good idea to write another test with invalid fees and expect the transaction to fail

yaruwangway: i cannot add the fail tests now, because of this issue. https://github.com/cosmos/gaia/issues/1534 which you also tested with simd last time.

tests/e2e/e2e_test.go
yaruwangway: yes, fee has to be 
1. higher or equal to `min_gas_price` in app.toml && 2. higher or equal to one of the fee in the global fee list.

yaruwangway: i think we will set global fee empty in rho, which means global fee would not check. because it works for all nodes.

mmulji-ic (approved):
@yaruwangway need to add failing tests, otherwise good.

yaruwangway:
merged to https://github.com/cosmos/gaia/pull/1594

PR #1536, opened 2022-06-15
chore: rename msg type

yaruwangway:
rename `transferMsgTransfer` to `ibcMsgTransfer`

### DIFF
--- a/app/const.go
+++ b/app/const.go
@@ -25,7 +25,7 @@ const (
 	stakingMsgBeginRedelegate           = "/cosmos.staking.v1beta1.MsgBeginRedelegate"
 	stakingMsgCreateValidator           = "/cosmos.staking.v1beta1.MsgCreateValidator"
 	vestingMsgCreateVestingAccount      = "/cosmos.vesting.v1beta1.MsgCreateVestingAccount"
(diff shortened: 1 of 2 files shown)


tokamak-git:
Created an issue so as to fix this: https://github.com/cosmos/gaia/issues/1543

app/const.go
```diff
@@ -25,7 +25,7 @@ const (
 	stakingMsgBeginRedelegate           = "/cosmos.staking.v1beta1.MsgBeginRedelegate"
 	stakingMsgCreateValidator           = "/cosmos.staking.v1beta1.MsgCreateValidator"
 	vestingMsgCreateVestingAccount      = "/cosmos.vesting.v1beta1.MsgCreateVestingAccount"
-	transferMsgTransfer                 = "/ibc.applications.transfer.v1.MsgTransfer"
+	ibcMsgTransfer                      = "/ibc.applications.transfer.v1.MsgTransfer"
```
tac0turtle, line 29: out of curiosity what is the reason for this file?

yaruwangway: this file is for whitelisting the allowed msg types for ica host, will add comment to this file.

PR #1539, opened 2022-06-16
add bypass-min-fee

yaruwangway:
This PR is to add bypass-min-fee to rho.

in cosmos-sdk 0.46, the  previous `NewMempoolFeeDecorator` and `NewDeductFeeDecorator` are combined into `NewDeductFeeDecorator`.

In rho, bypassfee will be added as a separate decorator which check the min-gas-price is meet meanwhile bypass configured msg types, and the` NewDeductFeeDecorator` will use a custom TxFeeCheck which escape the fee check, so `NewDeductFeeDecorator` only play the role of deduct fee.

please note:
the fee_test is comment out for now till next sdk 0.46 release with tendermint 0.34


tokamak-git:
At a glance LGTM, could we sort out the failing tests?

ante/fee.go
```diff
@@ -1,75 +1,95 @@
...
+
+// FeeWithBypassDecorator will check if the transaction's fee is at least as large
+// as the local validator's minimum gasFee (defined in validator config).
+//
+// If fee is too low, decorator returns error and tx is rejected from mempool.
+// Note this only applies when ctx.CheckTx = true. If fee is high enough or not
+// CheckTx, then call next AnteHandler.
+//
+// CONTRACT: Tx must implement FeeTx to use FeeWithBypassDecorator
+type FeeWithBypassDecorator struct {
```
glnro:
```suggestion
type BypassMinFeeDecorator struct {
```

ante/fee.go
```diff
@@ -1,75 +1,95 @@
...
+	BypassMinFeeMsgTypes []string
+}
+
+func NewFeeWithBypassDecorator(bypassMsgTypes []string) FeeWithBypassDecorator {
```
glnro:
```suggestion
func NewBypassMinFeeDecorator(bypassMsgTypes []string) BypassMinFeeDecorator {
```

ante/fee.go
```diff
@@ -1,75 +1,95 @@
...
-// }
+import (
+	"math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	tmstrings "github.com/tendermint/tendermint/libs/strings"
+)
+
+const maxBypassMinFeeMsgGasUsage = uint64(200_000)
```
mmulji-ic, line 11: Should this be a config file?

yaruwangway: there is no config file, there is only a `const.go` file but not in this package.

app/app.go
```diff
@@ -690,6 +683,18 @@ func NewGaiaApp(
...
+					}
+
+					feeCoins := feeTx.GetFee()
+					priority := gaiaante.GetTxPriority(feeCoins)
+
+					return feeCoins, priority, nil
+				},
 			},
 			IBCkeeper:            app.IBCKeeper,
 			BypassMinFeeMsgTypes: cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)),
```
okwme: OK i did some digging to figure out what exactly the `Get` is doing here.
it's an interface referenced [here](https://github.dev/cosmos/cosmos-sdk/blob/af8ad3d20d5c78b730063c8ed8d42f8ee8d22232/server/types/app.go#L34-L35).
The AppOptions are created [here](https://github.dev/cosmos/cosmos-sdk/blob/38a11320247c46c4454c345c67c6b9a73a548cca/server/start.go#L274) by passing `ctx.Viper` which means the `Get` method is defined by Viper [here](https://github.dev/spf13/viper/blob/da55858fffe6093e995244bfc0aa2d9fba370e30/viper.go#L884) and then cast with the `ToStringSlice` defined [here](https://github.dev/spf13/cast/blob/2b0eb0f724e320b655240e331aef36d1175986c2/caste.go#L1234).


It makes me believe that if the key was not present in the config file, `Get` would return `nil` whereas if it was defined but left empty (as in `""`) that it would return as an empty string.

That means that in this line of `app.go` we could separate the command so that it was:
```
bypassMinFeeConfig := appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)
if bypassMinFeeConfig != nil {
  bypassMinFeeMsgTypes := cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey))
} else {
  bypassMinFeeMsgTypes := " our list of default passed messages"
}
```

This way when people upgrae they don't have to add a new line to their config file to start using the new feature (and this PR will pass the E2E test that uses default Cosmos SDK config generation.

okwme: would be great if @alexanderbez could confirm this would work as expected

ante/fee.go
tokamak-git: Agree with @mmulji-ic  I guess we could find a way to pass this value as part of the app.toml file rite? add a new section that deals with configuration of the min fee. @yaruwangway  what do you think?

app/app.go
tokamak-git: @yaruwangway would it be possible to test the code snippet billy mentioned via a unit test?

app/helpers/test_helpers.go
```diff
@@ -48,21 +56,22 @@ func (EmptyAppOptions) Get(o string) interface{} { return nil }
...
 	t.Helper()
 
-	app, genesisState := setup(!isCheckTx, invCheckPeriod)
-	if !isCheckTx {
-		// InitChain must be called to stop deliverState from being nil
-		stateBytes, err := json.MarshalIndent(genesisState, "", " ")
-		require.NoError(t, err)
+	privVal := mock.NewPV()
+	//todo check context here
+	pubKey, err := privVal.GetPubKey(nil)
```
tokamak-git: @yaruwangway I agree with the linter, could we make the recommended changes?

Also do you see any issues with the recommended approach?

ante/fee.go
yaruwangway: we can, but the const is not in app.toml, these are two different things. the app.toml only sets the bypass fee types.

app/app.go
alexanderbez: So your understanding on how AppOpts work is correct @okwme, however, regarding defaults, this is what happens:

1. Gaia defines the defaults [here](https://github.com/cosmos/gaia/blob/main/cmd/gaiad/cmd/root.go#L105-L109) when `app.toml` does NOT exist. Note, I don't understand why the default was commented out???
2. The Gaia config extension is defined [here](https://github.com/cosmos/gaia/blob/main/app/params/config.go)
3. When `app.toml` DOES exist, it will parse an empty `[]string` for `bypass-min-fee-msg-types` when not explicitly set.

So regardless if the result is `nil` or `[]string`, the ante-handler should still work fine as iterating over a `nil` slice is a no-op.

alexanderbez: > It makes me believe that if the key was not present in the config file, Get would return nil whereas if it was defined but left empty (as in "") that it would return as an empty string.

No, it will return `nil` or `[]string{...}`.

yaruwangway: @okwme, do we want all nodes have the same bypass tx types ? i thought this is depending on what each node's wish. which means if you want to bypass ibc tx, you need to find a node that setup ibc as bypass msg types.

ante/fee.go
tokamak-git: @yaruwangway Do you think it would make sense for us to simply add a step to update a config file when a new field has been added with a default value?
Essentially a migration to the config files similar to how we do it for a migration to a new schema?

[tokamak](https://app.slack.com/team/U0310FRK1U4)
  [1 hour ago](https://interchainfoundation.slack.com/archives/C02QDK3JSJK/p1656334528805619?thread_ts=1656333756.169929&cid=C02QDK3JSJK)
we could always create a backup of the exiting app.toml

app/app.go
yaruwangway: @be

> So your understanding on how AppOpts work is correct @okwme, however, regarding defaults, this is what happens:
> 
> 1. Gaia defines the defaults [here](https://github.com/cosmos/cosmos-sdk/issues/12244#issuecomment-1155350896) when `app.toml` does NOT exist. Note, I don't understand why the default was commented out???
> 2. The Gaia config extension is defined [here](https://github.com/cosmos/gaia/blob/main/app/params/config.go)
> 3. When `app.toml` DOES exist, it will parse an empty `[]string` for `bypass-min-fee-msg-types` when not explicitly set.
> 
> So regardless if the result is `nil` or `[]string`, the ante-handler should still work fine as iterating over a `nil` slice is a no-op.

Hi @alexanderbez, can you check the the above link about the "Gaia defines the defaults here" , it seems not about the default config? I will have a look.

tokamak-git:
@yaruwangway Could you update the branch name to follow the `{moniker}/{issue#}-branch-name` convention?
refer: `https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md`

yaruwangway:
> @yaruwangway Could you update the branch name to follow the `{moniker}/{issue#}-branch-name` convention? refer: `https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md`

hi maybe next time. i tried to rename it , then this pr will be closed. reopen a new one will lose present pr comments.

ante/fee.go
tokamak-git: Plus my assumption is that this field will become part of the config recommendations moving forward rite?

app/app.go
alexanderbez: Whoops! So sorry!

Here is the intended link: https://github.com/cosmos/gaia/blob/main/cmd/gaiad/cmd/root.go#L105-L109

okwme: thanks @alexanderbez !
you mentioned that if no app.go exists the result will be an empty array of strings but what happens if the app.go exsists however the value for the bypassminfee messages is not set? The wording in the config extension says the values "may" be set but not "must" be set. So the config file is still valid while missing `bypass-min-fee-msg-types`. I'd like to ensure if it's missing that the default list is used.

I made this PR to demonstrate what I thought would be a solution to ensure node operators who don't update their config will use the default set of messages. Can you confirm this thinking is correct and the desired results would be achieved?

https://github.com/cosmos/gaia/pull/1567

yaruwangway:
Hi @alexanderbez , could you please have a look of this pr ? Thank you!

app/helpers/test_helpers.go
```diff
@@ -41,47 +50,83 @@ var DefaultConsensusParams = &tmproto.ConsensusParams{
 	},
 }
 
+// todo: check SetupOptions is not used.
```
glnro: I don't believe `SetupOptions` is actually being used anywhere in the code, this can be deleted.

ante/fee.go
glnro: At minimum we should write up a tech debt ticket to pull out all variables like this into some local const/vars file, for each directory

yaruwangway: some dir in gaia, there is a file called `alias.go` or `const.go`, we should name them all`alias.go` maybe.  @glnro 

for app config. we can custom it, here is [an example of code](https://github.com/cosmos/gaia/blob/main/app/params/config.go)
is this what you meant ? @tokamak-git

app/app.go
okwme: merged this PR!

app/app.go
```diff
@@ -682,6 +676,14 @@ func NewGaiaApp(
 	app.MountTransientStores(tkeys)
 	app.MountMemoryStores(memKeys)
 
+	var bypassMinFeeMsgTypes []string
+	bypassMinFeeConfig := appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)
+	if bypassMinFeeConfig != nil {
+		bypassMinFeeMsgTypes = cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey))
```
okwme:
```suggestion
		bypassMinFeeMsgTypes = cast.ToStringSlice(bypassMinFeeConfig)
```

okwme (approved):

ante/fee.go
okwme: Config migration is an alternative solution but since we have the fallback option working going to resolve this conversation.

This feature will eventually be added to the Cosmos SDK where [config validation](https://github.dev/cosmos/cosmos-sdk/blob/3283617cfb14dc90bf93e409a47bd1e37cd1f309/server/config/config.go#L363) can take place.

app/helpers/test_helpers.go
yaruwangway: the lint issue is solved.

PR #1509, opened 2022-05-26
gofumpt and automatic linting for gaia v2

faddat:
Gofumpted gaia v2 and added linting.

#1449 

- [ ] Linked to github-issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote tests
- [ ] Updated relevant documentation (`docs/`)
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Reviewed `Files changed` in the github PR explorer

______

For Admin Use:
- Added appropriate labels to PR (ex. wip, ready-for-review, docs)
- Reviewers Assigned
- Squashed all commits, uses message "Merge PR #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/master/README.md#merging-a-pr))

PR #1547, opened 2022-06-21
Update lint.yml

okwme:
opening in place of #1509 to make gaia consistent with cosmos sdk


.golangci.yml
```diff
@@ -58,9 +50,11 @@ issues:
     - text: "ST1016:"
       linters:
         - stylecheck
-    - text: "SA1019: codec.LegacyAmino is deprecated"
+    - path: "migrations"
+      text: "SA1019:"
```
tokamak-git, line 55: @okwme not quiet sure  i understand this bit. is the `text` field used as a regex? if so we are changing it?

okwme: would actually need to ask @marbar3778 or @alexanderbez i just copied from SDK's config

.github/workflows/lint.yml
```diff
@@ -1,25 +1,41 @@
...
-    name: Run golangci-lint
+    name: golangci-lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.17
+          go-version: 1.18
       - uses: technote-space/get-diff-action@v6.0.1
```
faddat, line 18: This right here is the problem

.github/workflows/lint.yml
```diff
@@ -1,25 +1,41 @@
...
+        id: git_diff
         with:
           PATTERNS: |
             **/**.go
             go.mod
             go.sum
-      - name: Run golangci-lint
-        run: make lint
+      - name: Get data from Go build cache
```
faddat: this right here is another problem.

.golangci.yml
faddat: The sdk's config is why this isn't firing the linter:

https://github.com/cosmos/cosmos-sdk/pull/11453

tac0turtle: I don't think this is needed. @faddat is right in why it's not firing

okwme: @marbar3778 can you make changes or specific suggestions of what is needed to change before merge?

.github/workflows/lint.yml
```diff
@@ -1,25 +1,41 @@
...
+        # if: env.GIT_DIFF
+        if: ${{ false }}
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/golangci-lint
+            ~/.cache/go-build
+          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
```
tac0turtle:
```suggestion
```

tac0turtle: this isn't needed as golangci=lint caches on its own. we need to clean up in the sdk

.github/workflows/lint.yml
tac0turtle: this is fine, on gaia is not super needed, but doesn't hurt anyone

tac0turtle (approved):

PR #1511, opened 2022-05-26
gofumpt for v4

faddat:
## Description

Applies gofumpt to gaia v4, standardizing the codebase

#1449 
______

For contributor use:

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Added appropriate labels to PR (ex. `WIP`, `R4R`, `docs`, etc)
- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/main/README.md#merging-a-pr))

PR #1510, opened 2022-05-26
gofumpt for v3

faddat:
## Description

Applies gofumpt to gaia v3 to make code easier to read

#1449 same
______

For contributor use:

- [ ] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/main/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing)
- [ ] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Added appropriate labels to PR (ex. `WIP`, `R4R`, `docs`, etc)
- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/main/README.md#merging-a-pr))

PR #1549, opened 2022-06-21
DRAFT: update Stable Release Policy

okwme:
Would be great to get input from @marbar3778 and @mmulji-ic on how to finalize this.
@faddat did a lot of work updating v2, v3 and v4 of gaia with changes to linting.
while the work is appreciated those kinds of changes to make sense to backport into a new release of those versions. EOL should only be bypassed by security updates.


okwme:
should add a definition for "bug-fix" since bugs that impact operating the network don't matter wince the previous version networks are no longer in operation. More so it should be privilege escalation or anything that makes the software actually dangerous to operate on a user's machine. Maybe these should actually be applied to all previous versions since archive use will require them to be run.

faddat:
#1510 
#1511 

So, code formatting and the like are in fact security updates.  If something were to happen, having maintained older gaias-- as well as people who maintain them, would be very important.  That is why it makes sense to backport changes like these-- at the very least as far back as cosmoshub-4 genesis (iirc that is v4)

tac0turtle (approved):
generally looks good to me. Id put a EOL for v5 since this is coming up in the middle of things

yaruwangway:
if we find a bug in v6, and our testnet and and hub are both at v7, we fix the bug of v6, and spin off a new testnet to test them, right? if it is not bug, it's just improvement of code, do we still improve v6 when we are at v7 ?

faddat:
@yaruwangway my opinion is that we should aim for consistency on these items:

* ci
* code formatting
* use (and verification) of latest cosmos-sdk

## Goal

The goal of standardizing ci, code formatting, and version bumps is to make it easy to correct issues, if their root cause is in the past.

RELEASING.md
```diff
@@ -133,3 +133,18 @@ Checksums-Sha256:
  3895518436b74be8b042d7d0b868a60d03e1656e2556b12132be0f25bcb061ef  gaiad-4.0.0.tar.gz
 ```
 
+# Major Release Maintenance
+
+Major Release series continue to receive bug fixes (released as a Patch Release) until they reach End Of Life. Major Release series is maintained in compliance with the Stable Release Policy as described in this document. Note: not every Major Release is denoted as stable releases.
+
+Only the following major release series have a stable release status:
+
+v7 Theta is supported until v9 Epsilon. A fairly strict **bugfix-only** rule applies to pull requests that are requested to be included into a stable point-release.
```
mmulji-ic: Epsilon = 10

faddat:
wdyt about this PR, with respect to this policy:

* https://github.com/cosmos/gaia/pull/1695

Is it suprerfluous?

okwme:
@faddat my understanding is that this PR should support #1695 as in scope.
there are other items to include in the next 7.0.x release being tracked here: https://github.com/cosmos/gaia/issues/1669

PR #1567, opened 2022-06-28
create fallback default bypass min fee messages

okwme:
Based on the discussion [here](https://github.com/cosmos/gaia/pull/1539#pullrequestreview-1019995309) these changes should make it possible to fall back to a default set of bypass min fee messages so that if a node operator hasn't added the new value to their config file, the default is used.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -97,6 +97,7 @@ import (
 	ibcclient "github.com/cosmos/ibc-go/v3/modules/core/02-client"
 	ibcclientclient "github.com/cosmos/ibc-go/v3/modules/core/02-client/client"
 	ibcclienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
+	ibcchanneltypes "github.com/cosmos/ibc-go/v3/modules/core/04-channel/types"
 	porttypes "github.com/cosmos/ibc-go/v3/modules/core/05-port/types"
 	ibchost "github.com/cosmos/ibc-go/v3/modules/core/24-host"
 	ibckeeper "github.com/cosmos/ibc-go/v3/modules/core/keeper"
@@ -675,6 +676,14 @@ func NewGaiaApp(
 	app.MountTransientStores(tkeys)
 	app.MountMemoryStores(memKeys)
 
+	var bypassMinFeeMsgTypes []string
+	bypassMinFeeConfig := appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)
+	if bypassMinFeeConfig != nil {
+		bypassMinFeeMsgTypes = cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey))
+	} else {
+		bypassMinFeeMsgTypes = GetDefaultBypassFeeMessages()
+	}
+
 	anteHandler, err := gaiaante.NewAnteHandler(
 		gaiaante.HandlerOptions{
 			HandlerOptions: ante.HandlerOptions{
@@ -697,7 +706,7 @@ func NewGaiaApp(
 				},
 			},
 			IBCkeeper:            app.IBCKeeper,
-			BypassMinFeeMsgTypes: cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)),
+			BypassMinFeeMsgTypes: bypassMinFeeMsgTypes,
 		},
 	)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -675,6 +676,14 @@ func NewGaiaApp(
...
 	app.MountMemoryStores(memKeys)
 
+	var bypassMinFeeMsgTypes []string
+	bypassMinFeeConfig := appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey)
+	if bypassMinFeeConfig != nil {
+		bypassMinFeeMsgTypes = cast.ToStringSlice(appOpts.Get(gaiaappparams.BypassMinFeeMsgTypesKey))
+	} else {
+		bypassMinFeeMsgTypes = GetDefaultBypassFeeMessages()
+	}
+
```
yaruwangway, line 686: Hi, I think this logic should be moved to `app.go` upgrade handler. 

also, since we already have bypass min fee in v7. would people feel ok that one upgrade changes their configs even though not all users ?  actually, if v7 node's config is from gaiad v7 init, they will have ibc as default bypass, if upgrade, the bypass fee  in app.toml is empty list.

app/app.go
```diff
@@ -784,6 +793,14 @@ func NewGaiaApp(
 	return app
 }
 
+func GetDefaultBypassFeeMessages() []string {
+	return []string{
+		sdk.MsgTypeURL(&ibcchanneltypes.MsgRecvPacket{}),
```
yaruwangway, line 798: this makes think maybe in our `const.go` we should use `sdk.MsgTypeURL()` rather than directly a string.
for example:  use `sdk.MsgTypeURL(&ibcchanneltypes.MsgRecvPacket{})` rather than direclty `"/ibc.core.channel.v1.MsgRecvPacket"`

cmd/gaiad/cmd/root.go
```diff
@@ -101,12 +99,8 @@ func initAppConfig() (string, interface{}) {
 	srvCfg.StateSync.SnapshotKeepRecent = 10
 
 	return params.CustomConfigTemplate, params.CustomAppConfig{
-		Config: *srvCfg,
-		// BypassMinFeeMsgTypes: []string{
```
yaruwangway, line 105: here is when we do `gaia init`, and get a default bypass min fee  in `app.toml`, this should be the same msg types getting from `GetDefaultBypassFeeMessages()` if we move the setting of default bypass min fee to upgrade handler in `app.toml` ?

app/app.go
okwme: this doesn't update the config file though, just allows the default values to be used in the case it has not been added to the config. does that change your suggestion or maybe i misunderstand?

app/app.go
okwme: This should be made into a separate issue where we could discuss the potential benefits

cmd/gaiad/cmd/root.go
okwme: Still don't understand exactly why to move to upgrade handler?

app/app.go
yaruwangway: you are right, my misunderstanding! there is no change of the config.

cmd/gaiad/cmd/root.go
yaruwangway: sry, i misunderstood that this will change app.toml, so no need to move to upgrade handler.

but when doing `gaia init`, if we want to write the default bypass msg types to app.toml ? if yes, we can uncomment here. line 106-109

okwme: Yes that's what we want, i already uncommented them in the pr : )

app/app.go
okwme: acutally IBC is going to add an "addAll" configuration for interchain accounts so we shouldn't optimize the list we're currently using

yaruwangway:
another question is what if the node just does not want to bypass any msgs?

okwme:
@yaruwangway the node can make an empty string and all messages will be bypassed. This is just if they don't include anything at all (no empty string or anything) which would signal they aren't aware of the setting at all and therefore the default values should be used.

yaruwangway:
> @yaruwangway the node can make an empty string and all messages will be bypassed. This is just if they don't include anything at all (no empty string or anything) which would signal they aren't aware of the setting at all and therefore the default values should be used.

ha, right !

yaruwangway:
will add e2e later to test this case later!

yaruwangway (approved):

PR #1570, opened 2022-06-30
custom global fee: make global accept fee bypass, define fee denoms

yaruwangway:
Global fee is a fee that all txs have to meet except bypass msg. Global fee can be set by gov params change proposals.

- bypass msg types will skip both global fee and min gas price fee checks.(but bypass msg fee if not empty, still need to be in the same denom of global fee)
- fee  must be higher than both global fee and min gas price
- if fee denom only exists in global fee, and does not exit in min gas price, then skip the min gas price checks

fee paying principle: 
global fee will check if the fee is in the denom of global fee, even for bypass msg or global fee =0. we do not want random denom to be sent to distr module acct. even the fee=0 random_denom.

if globalfee has a denom that is not in min_gas_pirce, you can take min_gas_price has 0 that denom. for example, `globalfee=1stake, min_gas_price=1photon`. it is the same as `globalfee=1stake, min_gas_price=1photon,0stake `

 examples for paying fees:
  globalfee = [1stake, 2photon], min_gas_price = 2stake. 
- tx fee = 1stake, fail
- tx fee = 2stake, pass
- tx fee = 2photon, pass
- tx fee = 3quark, fail
- tx fee= 2stake,3quark. fail
- x fee= 2stake,0quark. fail

bypass msgs fee:
- bypass msg fee =[], pass
- bypass msg fee = 0stake, 0photon, pass
- bypass msg fee = 0.5stake, pass
- bypass msg fee = 0quark, fail

**please note**: 
for gaia, if the global fee =[], we set the default globalfee = 0uatom.

@okwme  if there is more changes about fee desgin, please add comments here.


yaruwangway:
~~1. min_gas_price = 1stake, globalfee = 2 stake, then fee must be >=2 stake * gas~~

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
-	_, err = antehandler(s.ctx, oracleTx, false)
-	s.Require().NoError(err, "expected min fee bypass for IBC messages")
-
-	s.ctx = s.ctx.WithIsCheckTx(false)
-
-	// antehandler should not error since we do not check min gas prices in DeliverTx
-	_, err = antehandler(s.ctx, tx, false)
-	s.Require().NoError(err, "unexpected error during DeliverTx")
+// test global fees and min_gas_price with bypass msg types.
+// please note even globalfee=0, min_gas_price=0, we do not let fee=0random_denom pass, to prevent fee is flooded with a long list of 0fees
```
yaruwangway: i choose globalfee=0stake, min_gas_price=0stake,   fee = 0uatom fail.
please let me know if globalfee=0stake, min_gas_price=0stake,   fee = 0uatom pass is desired.

x/globalfee/types/params.go
```diff
@@ -32,11 +35,65 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 	}
 }
 
+// this requires the fee non-negative
```
yaruwangway, line 37: presently, when setup globalfee, in genesis or gov propose to write to store, the global fee must be sorted by denom. otherwise, panic.
if we do not require user to pay attention to denom order when setup global fee, 
we can sort it when call 
`func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {}`
```
func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
	p.MinimumGasPrices = p.GetMinimumGasPrices().Sort()
	return paramtypes.ParamSetPairs{
		paramtypes.NewParamSetPair(
			ParamStoreKeyMinGasPrices, &p.MinimumGasPrices, validateMinimumGasPrices,
		),
	}
}
```
however, the  *Param 's   MinimumGasPrices  might pass or fail `func (p Params) ValidateBasic() error ` depends on the validation is before or after the ` ParamSetPairs() `.
we can either sort it in `ValidateBasic()` again or leave it to users to pay attention of denom order when setup global fee.

ante/fee_test.go
okwme: `globalfee=0stake, min_gas_price=0stake, fee = 0uatom fail` is correct

x/globalfee/types/params.go
okwme: let's sort it for them in ValidateBasic 👍

ante/fee.go
```diff
@@ -5,26 +5,41 @@ import (
...
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
 	tmstrings "github.com/tendermint/tendermint/libs/strings"
+
+	"github.com/cosmos/gaia/v8/x/globalfee"
+	"github.com/cosmos/gaia/v8/x/globalfee/types"
 )
 
 const maxBypassMinFeeMsgGasUsage = uint64(200_000)
 
+var defaultZeroGlobalFee = []sdk.DecCoin{sdk.NewDecCoinFromDec("uatom", sdk.NewDec(0))}
```
okwme: This should change into a function that accepts `ctx` as a parameter so it can access the `BondDenom` of the staking module.

Something like this:

```go
func DefaultZeroGlobalFee(ctx sdk.Context) string {
	return []sdk.DecCoin{sdk.NewDecCoinFromDec(stakingKeeper.BondDenom(ctx), sdk.NewDec(0))} 
}
```

ante/fee.go
```diff
@@ -33,17 +48,34 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
 	if !ok {
 		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
 	}
-
-	feeCoins := feeTx.GetFee()
+	// getFee only get non zero fees?
```
okwme: there's no reason for someone to include a 0fee, it would be the same as including an empty fee array. If they want to include a 0fee it would increase the gas a bit but the danger in them increasing the gas by a lot and still not paying a fee is mediated by a max gas limit that we already have.

so there's no real risk in allowing 0fee here, although there's also no reason to do so.
if it makes later steps easier it would be fine to only get non zero fees.

yaruwangway: `feeCoins := feeTx.GetFee()` will get both zero and non zero fees now. i will remove the comment.

ante/fee.go
```diff
@@ -33,17 +48,34 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
 
-	// Only check for minimum fees if the execution mode is CheckTx and the tx does
+	//todo if need to check gas > uint64(len(msgs))*maxBypassMinFeeMsgGasUsage) here ?
+
+	// Only check for minimum fees and global fee if the execution mode is CheckTx and the tx does
 	// not contain operator configured bypass messages. If the tx does contain
 	// operator configured bypass messages only, it's total gas must be less than
-	// or equal to a constant, otherwise minimum fees are checked to prevent spam.
+	// or equal to a constant, otherwise minimum fees and global fees are checked to prevent spam.
 	if ctx.IsCheckTx() && !simulate && !(mfd.bypassMinFeeMsgs(msgs) && gas <= uint64(len(msgs))*maxBypassMinFeeMsgGasUsage) {
```
okwme: Could make this a bit more legible with variable names

```suggestion

	containsOnlyBypassMinFeeMsgs := mfd.bypassMinFeeMsgs(msgs)
	doesNotExceedMaxGasUsage := gas <= uint64(len(msgs))*maxBypassMinFeeMsgGasUsage
	allowedToBypassMinFee := containsOnlyBypassMinFeeMsgs && doesNotExceedMaxGasUsage

	if ctx.IsCheckTx() && !simulate && !allowedToBypassMinFee {
```

ante/fee.go
```diff
@@ -54,16 +86,52 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
+			// requiredFees here is ensured not all zero, when check min_gas_price, fee might be zero. if min_gas_price=0stake,1photon, and feecoins is 0stake, it should not return err. so use IsAnyGTE() rather than IsAnyGTE() from sdk.
+			if !IsAnyGTE(feeCoins, requiredFees) {
 				return ctx, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
 			}
 		}
 	}
 
+	// when the tx is bypass msg type, still need to check the denom is not random denom
+	// this is to prevent the situation that a bypass msg carries random fee denoms
+	if ctx.IsCheckTx() && !simulate && (mfd.bypassMinFeeMsgs(msgs) && gas <= uint64(len(msgs))*maxBypassMinFeeMsgGasUsage) && mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) {
```
okwme:
```suggestion
	if ctx.IsCheckTx() && !simulate && allowedToBypassMinFee {
```
why does it check for `mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) ` here?
I read the logic as:
* it's a checkTx
* it's not a simulation
* it's allowed to bypass min fee
* the global bypass parameter is not empty

if the global bypass parameter is empty it should be considered to have the `DefaultZeroGlobalFee` and then the denom should still be checked, right?

ante/fee.go
```diff
@@ -54,16 +86,52 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
 }
 
+//coins must be sorted
+func (mfd BypassMinFeeDecorator) getGlobalFee(ctx sdk.Context, feeTx sdk.FeeTx) sdk.Coins {
+	var globalMinGasPrices sdk.DecCoins
+	mfd.GlobalMinFee.Get(ctx, types.ParamStoreKeyMinGasPrices, &globalMinGasPrices)
+
+	// global fee is empty set, set global fee to 0uatom
+	if len(globalMinGasPrices) == 0 {
+		globalMinGasPrices = defaultZeroGlobalFee
```
okwme: don't forget to replace with `DefaultZeroGlobalFee()` if that function is used instead.

ante/fee.go
```diff
@@ -93,3 +162,54 @@ func GetTxPriority(fee sdk.Coins) int64 {
...
+
+// overwrite DenomsSubsetOf from sdk, to allow zero amt coins. e.g. 1stake is DenomsSubsetOf 0stake.
+// DenomsSubsetOf returns true if coins's denoms is subset of coinsB's denoms.
+// if coins is empty set, empty set is any sets' subset
+func DenomsSubsetOf(coins, coinsB sdk.Coins) bool {
+	for _, coin := range coins {
+		// validate denom
+		err := sdk.ValidateDenom(coin.Denom)
+		if err != nil {
+			panic(err)
```
okwme: have these denoms been validated already?
is there a reason to validate them again if so?
furthermore if they fail the validation shouldn't we return an error and make the checkTx fail instead of a panic?
wouldn't a panic stop the node?

ante/fee.go
yaruwangway: for bypass msg, if because if we do not have global fee module, then we do not check global fee denom.

ante/fee.go
```diff
@@ -33,17 +48,34 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
+	// or equal to a constant, otherwise minimum fees and global fees are checked to prevent spam.
 	if ctx.IsCheckTx() && !simulate && !(mfd.bypassMinFeeMsgs(msgs) && gas <= uint64(len(msgs))*maxBypassMinFeeMsgGasUsage) {
+		// check global fees
+		if mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) {
+			//requiredGlobalFees is sorted
+			requiredGlobalFees := mfd.getGlobalFee(ctx, feeTx)
+			if !DenomsSubsetOf(feeCoins, requiredGlobalFees) {
+				return ctx, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "fees %s is not a subset of required global fees %s", feeCoins, requiredGlobalFees)
+			}
+
```
okwme:
```suggestion
			// At least one feeCoin amount must be greater than or equal to the requiredGlobalFees
```

ante/fee.go
```diff
@@ -33,17 +48,34 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
 	// not contain operator configured bypass messages. If the tx does contain
 	// operator configured bypass messages only, it's total gas must be less than
-	// or equal to a constant, otherwise minimum fees are checked to prevent spam.
+	// or equal to a constant, otherwise minimum fees and global fees are checked to prevent spam.
 	if ctx.IsCheckTx() && !simulate && !(mfd.bypassMinFeeMsgs(msgs) && gas <= uint64(len(msgs))*maxBypassMinFeeMsgGasUsage) {
+		// check global fees
+		if mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) {
+			//requiredGlobalFees is sorted
+			requiredGlobalFees := mfd.getGlobalFee(ctx, feeTx)
+			if !DenomsSubsetOf(feeCoins, requiredGlobalFees) {
```
okwme:
```suggestion
                         // All denoms inside the transaction fee must be included in the list of required global fee denoms
			if !DenomsSubsetOf(feeCoins, requiredGlobalFees) {
```

ante/fee.go
yaruwangway: this is a good point, so the DefaultZeroGlobalFee is not in the store or genesis, it only exists in logics when checking the fee or `getGlobalFee` the fee.

` mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) == true` means we have the global fee module, but the value might be empty, zero, or nonzero.  

so basically the `mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) ` is just to make sure that the global fee module exsits.

ante/fee.go
yaruwangway: i think the coinB, which is globalfee is validated when adding to store. but the coins  which is paid fee are not ? i will check this again.

x/globalfee/types/params.go
yaruwangway: if we sort them, we sort them in two places, one is in `ParamSetPairs()` one is in `ValidateBasic()`, so maybe we call `ValidateBasic()` -> sortAndvalidateBasics()`

ante/fee.go
```diff
@@ -93,3 +162,54 @@ func GetTxPriority(fee sdk.Coins) int64 {
...
+		if ok, _ := coinsB.Find(coin.Denom); !ok {
+			return false
+		}
+	}
+
+	return true
+}
+
+// overwrite the IsAnyGTE from sdk to allow zero coins.
+// IsAnyGTE returns true iff coins contains at least one denom that is present
```
okwme:
```suggestion
// IsAnyGTE returns true if coins contain at least one denom that is present
```

ante/fee.go
```diff
@@ -93,3 +162,54 @@ func GetTxPriority(fee sdk.Coins) int64 {
...
+// at a greater or equal amount in coinsB; it returns false otherwise.
+// NOTE: IsAnyGTE operates under the invariant that both coin sets are sorted
+// by denominations.
+func IsAnyGTE(coins, coinsB sdk.Coins) bool {
```
okwme: May be good to distinguish the behavior in the function name
```suggestion
func IsAnyGTEIncludingZero(coins, coinsB sdk.Coins) bool {
```

ante/fee.go
```diff
@@ -93,3 +162,54 @@ func GetTxPriority(fee sdk.Coins) int64 {
 
 	return priority
 }
+
+// overwrite DenomsSubsetOf from sdk, to allow zero amt coins. e.g. 1stake is DenomsSubsetOf 0stake.
+// DenomsSubsetOf returns true if coins's denoms is subset of coinsB's denoms.
+// if coins is empty set, empty set is any sets' subset
+func DenomsSubsetOf(coins, coinsB sdk.Coins) bool {
```
okwme: May be good to distinguish the behavior in the function name
```suggestion
func DenomsSubsetOfIncludingZero(coins, coinsB sdk.Coins) bool {
```

ante/fee.go
```diff
@@ -93,3 +162,54 @@ func GetTxPriority(fee sdk.Coins) int64 {
...
+// at a greater or equal amount in coinsB; it returns false otherwise.
+// NOTE: IsAnyGTE operates under the invariant that both coin sets are sorted
+// by denominations.
+func IsAnyGTE(coins, coinsB sdk.Coins) bool {
+	// this is different from sdk, sdk return false
+	if len(coinsB) == 0 {
+		// global fee should not be not empty
+		panic("empty coin set")
+	}
+	// if feecoins empty, and globalfee has one denom of amt zero. feecoins equals to that 0denom.
```
okwme: if `len(coins) == 0` shouldn't it just pass?
Replacing `coins` with `coinsB` means coins will always be GTE to `coinsB`, right?

ante/fee_test.go
okwme:
```suggestion
// please note even globalfee=0, min_gas_price=0, we do not let fee=0random_denom pass
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+    globalfeeParams0 :=  &globfeetypes.Params{MinimumGasPrices: []sdk.DecCoin{
+        sdk.NewDecCoinFromDec("photon", sdk.NewDec(0)),
+        sdk.NewDecCoinFromDec("uatom", sdk.NewDec(0)),
+    }}
+    minGasPrice0 := []sdk.DecCoin{
+       sdk.NewDecCoinFromDec("stake", sdk.NewDec(0)),
+        sdk.NewDecCoinFromDec("uatom", sdk.NewDec(0))}
+    globalfeeParamsHigh := &globfeetypes.Params{
+        MinimumGasPrices: []sdk.DecCoin{
+            sdk.NewDecCoinFromDec("uatom", sdk.NewDec(400).Quo(sdk.NewDec(100000))),
```
okwme:
```suggestion
            sdk.NewDecCoinFromDec("uatom", high), // 0.004uatom
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+        },
+    }
+    minGasPrice := []sdk.DecCoin{
+        sdk.NewDecCoinFromDec("uatom", sdk.NewDec(200).Quo(sdk.NewDec(100000))),
```
okwme:
```suggestion
        sdk.NewDecCoinFromDec("uatom", med), // 0.002uatom
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+        sdk.NewDecCoinFromDec("stake", sdk.NewDec(200).Quo(sdk.NewDec(100000))),
```
okwme:
```suggestion
        sdk.NewDecCoinFromDec("stake", med), // 0.002stake
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+    }
+    globalfeeParamsLow := &globfeetypes.Params{
+        MinimumGasPrices: []sdk.DecCoin{
+            sdk.NewDecCoinFromDec("uatom", sdk.NewDec(100).Quo(sdk.NewDec(100000))),
```
okwme:
```suggestion
            sdk.NewDecCoinFromDec("uatom", low), // 0.001uatom
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+        },
+    }
+    // global fee must be sorted in denom
+    globalfeeParamsNewDenom := &globfeetypes.Params{
+        MinimumGasPrices: []sdk.DecCoin{
+            sdk.NewDecCoinFromDec("photon", sdk.NewDec(400).Quo(sdk.NewDec(100000))),
```
okwme:
```suggestion
            sdk.NewDecCoinFromDec("photon", high), // 0.004photon
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+            sdk.NewDecCoinFromDec("quark", sdk.NewDec(400).Quo(sdk.NewDec(100000))),
```
okwme:
```suggestion
            sdk.NewDecCoinFromDec("quark", high), // 0.004quark
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
-	s.Require().NoError(err, "unexpected error during DeliverTx")
+// test global fees and min_gas_price with bypass msg types.
+// please note even globalfee=0, min_gas_price=0, we do not let fee=0random_denom pass, to prevent fee is flooded with a long list of 0fees
+func (s *IntegrationTestSuite) TestGlobalFeeMinimumGasFeeAnteHandler() {
+    // setup test
+    s.SetupTest()
+    s.txBuilder = s.clientCtx.TxConfig.NewTxBuilder()
+    priv1, _, addr1 := testdata.KeyTestPubAddr()
+    privs, accNums, accSeqs := []cryptotypes.PrivKey{priv1}, []uint64{0}, []uint64{0}
+
```
okwme:
```suggestion
    denominator := 100000
    high := sdk.NewDec(400).Quo(sdk.NewDec(denominator))  // 0.004
    med := sdk.NewDec(200).Quo(sdk.NewDec(denominator))  // 0.002
    low := sdk.NewDec(100).Quo(sdk.NewDec(denominator))  // 0.001
```

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+    globalfeeParamsNewDenom := &globfeetypes.Params{
+        MinimumGasPrices: []sdk.DecCoin{
+            sdk.NewDecCoinFromDec("photon", sdk.NewDec(400).Quo(sdk.NewDec(100000))),
+            sdk.NewDecCoinFromDec("quark", sdk.NewDec(400).Quo(sdk.NewDec(100000))),
+        },
+    }
+    testCases := map[string]struct {
+        minGasPrice     []sdk.DecCoin
+        globalFeeParams *globfeetypes.Params
+        feeAmount       sdk.Coins
```
okwme: I think we should replace `feeAmount` with `gasPrice` so that it's easier to see whether it is above, below or the same as the minGasPrice or globalFeeParam in the test. We should be able to calculate the fee by multiplying the `gasPrice` by the expected gas amount later on. We can either hard-code the expected gas amount per message later on or get it by simulating the transaction with the transaction builder.

that's why i suggested changing all the `minGasPrice` and `globalFeeParams` to be functions of variables instead of hard coded numbers. Then you can do like:

```go
testCases := map[string]struct {
        minGasPrice     []sdk.DecCoin
        globalFeeParams *globfeetypes.Params
        gasPrice       []sdk.DecCoin
        gasLimit        sdk.Gas
        txMsg           sdk.Msg
        txCheck         bool
        expErr          bool
    }{
        // test fees
        // empty min_gas_price or empty global fee
        "empty min_gas_price, nonempty global fee, fee higher/equal than global_fee": {
            minGasPrice:     minGasPriceEmpty,
            globalFeeParams: globalfeeParamsHigh,
            gasPrice:       sdk.NewDecCoinFromDec("uatom", high.MulInt(2)),
            gasLimit:        testdata.NewTestGasLimit(),
            txMsg:           testdata.NewTestMsg(addr1),
            txCheck:         true,
            expErr:          false,
        },
}
```
to clearly demonstrate that the gasPrice is greater than the high global fee because it is 2x as much.

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+            expErr:          false,
+        },
+    }
+    
+    for name, testCase := range testCases {
+        s.Run(name, func() {
+            // set globalfees and min gas price
+            subspace := s.setupTestGlobalFeeStoreAndMinGasPrice(testCase.minGasPrice, testCase.globalFeeParams)
+            // setup antehandler
+            mfd := ante.NewBypassMinFeeDecorator([]string{
```
okwme: we should use `gaia.GetDefaultBypassFeeMessages()` here.

okwme: actually better to be explicit since they're attached to the test

ante/fee_test.go
```diff
@@ -1,59 +1,455 @@
...
+                sdk.MsgTypeURL(&ibcchanneltypes.MsgRecvPacket{}),
+                sdk.MsgTypeURL(&ibcchanneltypes.MsgAcknowledgement{}),
+                sdk.MsgTypeURL(&ibcclienttypes.MsgUpdateClient{}),
+            }, subspace)
+            antehandler := sdk.ChainAnteDecorators(mfd)
+            
+            s.Require().NoError(s.txBuilder.SetMsgs(testCase.txMsg))
+            s.txBuilder.SetFeeAmount(testCase.feeAmount)
```
okwme: this is where we'd want to set the fee as a calculation of `estimatedGas` * `gasPrice`.
Still trying to figure out how to get the `estimatedGas` though.
There's a `CalculateGas` function from the `tx` package but it requires `Factory` as a parameter which looks like a cumbersome thing to make.

okwme: This is what's going on inside of `CalculateGas` maybe we can just run it directly...
```go
import (
        "context"
	"github.com/cosmos/cosmos-sdk/types/tx"
	gogogrpc "github.com/gogo/protobuf/grpc"

)

CalculateGas(clientCtx  gogogrpc.ClientConn, msgs ...sdk.Msg) (uint64, error) {
	txSvcClient := tx.NewServiceClient(clientCtx)
	simRes, err := txSvcClient.Simulate(context.Background(), &tx.SimulateRequest{
		TxBytes: txBytes,
	})
	if err != nil {
		return nil, 0, err
	}
        return float64(simRes.GasInfo.GasUsed), nil
)
```

okwme: this is how the SDK tests the `Simulate` function: https://github.dev/cosmos/cosmos-sdk/blob/b75c2ebcfab1a6b535723f1ac2889a2fc2509520/baseapp/baseapp_test.go#L1099

Could be used but we need an instance of `app` to run `Simulate` on.

x/globalfee/types/params.go
```diff
@@ -9,7 +11,8 @@ import (
 // ParamStoreKeyMinGasPrices store key
 var ParamStoreKeyMinGasPrices = []byte("MinimumGasPricesParam")
 
-// DefaultParams returns default wasm parameters
+// todo set DefaultGlobalFee
```
okwme: can we access the staking keeper from here?
i don't think so, so maybe shouldn't set the default here but rather make the default empty and interpret empty as Default later on.

x/globalfee/types/params.go
```diff
@@ -32,11 +35,65 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
...
+	case 0:
+		return nil
+
+	case 1:
+		// match the denom reg expr
+		if err := sdk.ValidateDenom(coins[0].Denom); err != nil {
+			return err
+		}
+		if coins[0].IsNegative() {
+			return fmt.Errorf("coin %s amount is negtive", coins[0])
```
okwme, line 63: I think we want to accept 0 as an acceptable global min fee.
For instance at genesis when it is empty it will be equivalent to `0uatom`.
If we want to add a new consumer chain, we will want to allow the denom of that consumer chain to be accepted on the hub.
We may also still be OK with min fee being 0 of that new denom.
The param in that instance should be `0uatom,0ibc/asdfasdfasdf` so that `uatom` and the new `ibc/asdfasdfaf` denom can be used to pay fees.
Then it is up to each validator to increase the min fee in their app.toml, or a later proposal may decide to increase the min fee globally for one or both to something above 0.
we want to ensure it's still possible to have multiple denoms where one or more are 0.

okwme: this logic seems to only be applied when there is one fee listed, which means the scenario I outlined above pass inside the `default` case below. That's great but we also want to ensure that a single 0 fee denom is still valid. For instance it might just be good housekeeping to make the `0uatom` global min fee explicit instead of relying on the default fallback value (maybe we don't want to use the `BondDenom` as the global min fee for instance).

okwme (changes requested):
HEROIC PR @yaruwangway great work !!!!
I left a few suggestions to make the test cases more legible.
We're dealing with some pretty hairy logic statements and I want to make sure they're crystal clear when reading the code 💎

ante/fee.go
yaruwangway: this is checking if globalfee=[], since we set default zero fee, so our global fee will not be empty, so we can either `if len(coinsB) == 0 {return false} ` or `if len(coinsB) == 0 {panic("empty globalfee")} `

yaruwangway: or maybe a better way:
no set is empty set's subset except empty set.
so 
if len(coinsB) == 0= && len(coins) ==0. return true
if len(coinsB) == 0 && len(coins) !=0. return false

x/globalfee/types/params.go
yaruwangway: I just do not know if we automatically sort them, if directly change genesis with unsorted globalfee,
 our validation will sort them and say its valid genesis. write to the db is not a problem, because we will sort them and write to store.
just do not know if in genesis, globalfee not sorted, if there is potential issues?

ante/fee.go
yaruwangway: i changed it to 
```
const defaultZeroGlobalFeeDenom = "uatom"


func DefaultZeroGlobalFee() []sdk.DecCoin {
	return []sdk.DecCoin{sdk.NewDecCoinFromDec(defaultZeroGlobalFeeDenom, sdk.NewDec(0))}
}
```
i think we cannot access staking.bondDenom now? since it is at app, but we do not get app yet.

another way to do it is to store default denom and even amt at globalfee params in proto. sth like 
```
type Params struct {
	MinimumGasPrices github_com_cosmos_cosmos_sdk_types.DecCoins 
       DefaultMinGasPrice github_com_cosmos_cosmos_sdk_types.DecCoins 
}
```

ante/fee.go
yaruwangway: i think we need to validiate here.

ante/fee_test.go
yaruwangway: i think this is not needed, since we already say in min gas price and global fee if its high or low. there is no need to set their amt to be var named high, low?

x/globalfee/types/params.go
yaruwangway: i dont understand. the if we. do not set anything for global fee , it is empty. we want default to be 0uatom right ? so that we can constrain paying fees denom.

x/globalfee/types/params.go
yaruwangway: 1. "For instance at genesis when it is empty it will be equivalent to 0uatom." yes, we are doing this now. just " 0uatom" is not written to genesis or store, it only exists in logic. unless gov propose 0uatom to global fee, then it will be written to store.
2. "We may also still be OK with min fee being 0 of that new denom."  do you mean we also want to accept ibc/xxx as fee denom ? like paying 1 ibc/xxx will pass.
if we accept ibc/xxxx as fee, as long as consumer chain accept any denom transfered from other chains to consumer chain, these denom can be transfered to hub as fee ?  node's min_gas_price cannot filter too much because the attacker can always setup a node with 0ibc/xxx if global fee does not control this
3. "this logic seems to only be applied when there is one fee listed", we can always gov propose 0fees as well, 0uatom, 0ibc/xxxx.

ante/fee_test.go
yaruwangway: i see why, changed accordingly.

ante/fee_test.go
yaruwangway: `s.txBuilder.SetGasLimit(testCase.gasLimit)` set gas.
in sdk, it is [here](https://github.com/cosmos/cosmos-sdk/blob/5ef43f4919d59715df096145439a1022bfe54327/x/auth/migrations/legacytx/stdtx_builder.go#L56).

yaruwangway:
shall we refactor `gaia/ante` to `gaia/x/globalfee/ante` ?  @okwme

ante/ante.go
```diff
@@ -48,11 +47,10 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 		ante.NewTxTimeoutHeightDecorator(),
 		ante.NewValidateMemoDecorator(opts.AccountKeeper),
 		ante.NewConsumeGasForTxSizeDecorator(opts.AccountKeeper),
-		NewBypassMinFeeDecorator(opts.BypassMinFeeMsgTypes),
+		NewBypassMinFeeDecorator(opts.BypassMinFeeMsgTypes, opts.GlobalFeeSubspace),
```
alexanderbez, line 50: It's cleaner to instead pass the keeper directly here instead of it's x/param subspace (which is being deprecated anyway).

ante/fee.go
```diff
@@ -33,17 +52,38 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
 		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
 	}
-
-	feeCoins := feeTx.GetFee()
+	feeCoins := feeTx.GetFee().Sort()
 	gas := feeTx.GetGas()
 	msgs := feeTx.GetMsgs()
 
-	// Only check for minimum fees if the execution mode is CheckTx and the tx does
+	//todo if need to check gas > uint64(len(msgs))*maxBypassMinFeeMsgGasUsage) here ?
```
alexanderbez: Should address any TODOs in the PR 👍

ante/fee.go
```diff
@@ -54,16 +94,53 @@ func (mfd BypassMinFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
+				return next(ctx, tx, simulate)
+			}
+			// requiredFees here is ensured not all zero, when check min_gas_price, fee might be zero. if min_gas_price=0stake,1photon, and feecoins is 0stake, it should not return err. so use IsAnyGTEIncludingZero() rather than IsAnyGTEIncludingZero() from sdk.
+			if !IsAnyGTEIncludingZero(feeCoins, requiredFees) {
 				return ctx, sdkerrors.Wrapf(sdkerrors.ErrInsufficientFee, "insufficient fees; got: %s required: %s", feeCoins, requiredFees)
 			}
 		}
 	}
 
+	// when the tx is bypass msg type, still need to check the denom is not random denom
```
alexanderbez: I wouldn't use the nomenclature "random" here, but rather denoms that are not supported for x/globalfee.

ante/fee.go
```diff
@@ -76,7 +153,8 @@ func (mfd BypassMinFeeDecorator) bypassMinFeeMsgs(msgs []sdk.Msg) bool {
 	return true
 }
 
-//utils function: GetTxPriority
+//utils function: GetTxPriority, DenomsSubsetOf, IsAnyGTEIncludingZero
```
alexanderbez: ???

yaruwangway:
presently design  eccef034c3947eeaf36e713a2a0d689c805e662e
custom global fee ante handler does not allow zero random denom fees passing (for example, `gas-prices=1uatom,0photon`) will fail. if the global fee ='0.5uatom'

This design does not work!!! because when [parsing fee from flag, the zero fee is already removed]( https://github.com/cosmos/cosmos-sdk/blob/e716e4103e934344aa7be6dc9b5c453bdec5f225/client/tx/factory.go#L144). till reaching global fee ante handler. there is no zero fees. we do not need to consider/ and cannot control that  "zero randnom denom fee should not pass."

change global fee ante handler:
if globalfee=["0.5uatom"],  `--gas-price=1uatom,0photon`pass, `--gas-price=1uatom,1photon` fail.

yaruwangway:
discussion: 
1. validate globalfee =[0uatom], valid.
2. validate globalfee =[1uatom,1stake], valid.
3. validate globalfee =[0uatom,1stake], invalid ?  you can choose not pay or pay >=1stake or pay >=0uatom ?

globalfee=[1stake, 0.5uatom], fee=[0.5stake] , fail ? because users can pay any amt uatom, but we only accept stake higher than 1 ?
globalfee=[1stake, 0.5uatom], fee=[] , pass ? because fee=[] is equal to fee=[0uatom] ?

yaruwangway:
For the new design:
because when [parsing fee from flag, the zero fee is already removed](https://github.com/cosmos/cosmos-sdk/blob/e716e4103e934344aa7be6dc9b5c453bdec5f225/client/tx/factory.go#L144).
so we might have the strange cases:
globalfee=[0uatom, 1stake], if fees=[0.5stake] fail
globalfee=[0uatom, 1stake], if fees=[0uatom, 0.5stake] fail
globalfee=[0uatom, 1stake], if fees=[] pass
globalfee=[0uatom, 1stake], if fees=[1uatom, 0.5stake] pass
globalfee=[0uatom, 1stake], if fees=[0uatom, 0stake] pass, due to the the parsing of fees, it makes fees=[]
globalfee=[0uatom, 1stake], if fees=[1stake] pass

we cannnot control this in fee antehandler, because when feetx reach antehandler, the zero coins are removed already.

ante/fee.go
okwme:
```suggestion
	// when the tx is bypass msg type, still need to check the denom is not some unknown denom
```

ante/ante.go
okwme: going to leave this for when params is removed since this PR has been open so long and this seems like a "nice to have"

PR #1574, opened 2022-07-04
swap out ignite for liveness test

okwme:
currently fails as it tries to run `go mod tidy` but gets the following error:
```
go mod tidy: go.mod file indicates go 1.18, but maximum supported version is 1.17
```


mmulji-ic:
make start-localnet-ci failed with ...

make: *** [start-localnet-ci] Error 1
Error: set min gas price in app.toml or flag or env variable: error in app.toml

We should add some toml mangling.

mmulji-ic (approved):

PR #1577, opened 2022-07-07
Delete CODE_OF_CONDUCT.md

faddat:
consumer chains need rigorous protection from all forms of abuse.


mmulji-ic:
Hello @faddat , there is an in-progress activity looking at the current code of conduct. Having any code of conduct in place is there to protect all contributors including yourself.

faddat:
@mmulji-ic it doesn't protect me, it was used in an attempt to silence me from disclosing information about prop 69 and the interchain foundation, here's some additional information and I believe that closing this PR, just goes to show:

Interchain foundation isn't with the fing validators, just with itself.

--and that is why the hub isn't the right place for consumer chains (yet)

we need to free her from bondage.

The behavior of the hub team towards myself, documented here, shows very clearly that the hub team uses the code of conduct to implement its political objectvies.


https://github.com/notional-labs/notional/blob/master/incidents/Letter%20to%20the%20cosmos%20ecosystem%20concerning%20the%20care%20and%20maintenance%20of%20the%20cosmos%20hub.pdf

Unlike my other pull requests, this one was responded to with blinding speed.  Why is that?

PS:

You're lying

image

PR #1583, opened 2022-07-11
Fix typo

rootulp:


okwme (approved):
"Whether the weather be cold Or whether the weather be hot, We'll weather the weather, Whatever the weather, Whether we like it or not!"

thank you @rootulp !

PR #1588, opened 2022-07-15
Fix formatting in join-testnet.md

k-yang:
The code block endquote was missing, causing the page to have weird formatting.

Screen Shot 2022-07-15 at 8 38 31 AM


mmulji-ic (approved):
Thxs @k-yang for catching this.

PR #1590, opened 2022-07-18
Update markdown links/config

mmulji-ic:
Http config header check to (hopefully) remove the 403 error for some url's --> didn't work, reverted
Disabled check on 2 links that were causing issues

### DIFF
--- a/docs/hub-tutorials/join-mainnet.md
+++ b/docs/hub-tutorials/join-mainnet.md
@@ -51,6 +51,7 @@ There are many explorers for the Cosmos Hub. For reference while setting up a no
 - [Mintscan](https://www.mintscan.io/cosmos)
 - [Big Dipper](https://cosmos.bigdipper.live/)
 - [Hubble](https://hubble.figment.io/cosmos/chains/cosmoshub-4)
+<!-- markdown-link-check-disable-next-line -->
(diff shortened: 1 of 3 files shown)


docs/hub-tutorials/join-mainnet.md
```diff
@@ -51,6 +51,7 @@ There are many explorers for the Cosmos Hub. For reference while setting up a no
 - [Mintscan](https://www.mintscan.io/cosmos)
 - [Big Dipper](https://cosmos.bigdipper.live/)
 - [Hubble](https://hubble.figment.io/cosmos/chains/cosmoshub-4)
+<!-- markdown-link-check-disable-next-line -->
```
yaruwangway, line 54: why not put at the beginning of the file ?

mmulji-ic: I'm disabling just one link, not everything.

yaruwangway: why disable it ?

mmulji-ic: Because it gives a 403 error, site is still available though

yaruwangway (approved):

Issue #1564, opened 2022-06-27
Rho: add bypass min fee test in e2e test


yaruwangway:
## Summary

add bypass min fee test to e2e test.

## Problem Definition
custom app.toml file to add `bypass-min-fee-msg-types = []` 

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

yaruwangway:
i find a way to add bypass fee to app.toml  to form a  custom config! @okwme 

node setup:
min_gas_price in app.toml non-zero
set bypass msg types in app.toml

submit bypass msg txs to this node.
@tokamak-git

PR #1594, opened 2022-07-19
test: global fee e2e tests, bypass fee e2e

yaruwangway:
Close: #1564
e2e tests for global fee
- [x] add test: gov proposing global fee
- [x] add test: query global fee
- [x] add tests: bypass fee by withdraw reward.
- [x] refactor gov propose, deposit, vote  by passing fee as argument
- [x] refactor banksend to banksend with consider expected Err or not

ways of testing global fee:
initial global fee in genesis:  set the same value as min_gas_price ->
gov propose new global fee -> test different cases by bank send ->
gov propose another new global fee -> test different cases by bank send ->
...
check if the receiver receive all the success bank send tokens ->
gov propose back the global fee so that global fee will not influence other e2e tests.

please note: 
further refactor of e2e will be done later when all e2e prs are merged.
https://github.com/cosmos/gaia/issues/1608


tests/e2e/e2e_test.go
```diff
@@ -368,3 +384,340 @@ func (s *IntegrationTestSuite) verifyChainPassesUpgradeHeight(c *chain, valIdx i
...
+   - tx with fee 0.000001photon, fail
+   - tx with fee 0.000001photon, pass
+   - tx with empty fee, pass
+   - tx with fee uatom pass
+   - tx with fee 0uatom, 0.000005photon fail
+   - tx with fee 0uatom, 0.00001photon pass
+5. check balance correct: all the sucessful tx sent token amt is received
+6. gov propose change back to initial globalfee = 0.00001photon, This is for not influence other e2e tests.
+*/
+func (s *IntegrationTestSuite) TestGlobalFees() {
```
glnro, line 427: We should probably start moving these into their own files

yaruwangway: what does this mean ?

okwme: let's add it to the refactor issue.
@yaruwangway i think she means that we could have a separate file for each module to keep this file from being too large.

tests/e2e/e2e_test.go
```diff
@@ -358,3 +374,370 @@ func (s *IntegrationTestSuite) verifyChainPassesUpgradeHeight(c *chain, valIdx,
...
+			return globalFees.IsEqual(emptyGlobalFee)
+		},
+		15*time.Second,
+		5*time.Second,
+	)
+
+	paidFeeAmt := sdk.MustNewDecFromStr(minGasPrice).Mul(sdk.NewDec(gas)).String()
+
+	s.T().Logf("test case: empty global fee, globalfee=%s, min_gas_price=%s", globalFees.String(), minGasPrice+uatomDenom)
+	s.T().Logf("Tx fee is zero coin with correct denom: uatom, fail")
```
okwme, line 494: if the global fee is empty shouldn't the default be 0uatom, which means that a zero coin fee with uatom should pass?

okwme: ah i see min_gas_price is not zero

PR #1600, opened 2022-07-20
Correcting the email address in Code of Conduct

syedc5:
contact@interchain.berlin is no longer active/monitored. Changing to hello@interchain.io.

Not considering this to be a material change to the CoC


mmulji-ic:
Thanks @syed-Interchain for the update!

PR #1604, opened 2022-07-21
feat: globalfee/min_gas_price antehandler

yaruwangway:
Closes: https://github.com/cosmos/gaia/pull/1570

1.  there is no zero coin in fees when reaching the fee antehandler. we do not need to consider/ and cannot control that "zero randnom denom fee should/ should not pass." This is becuase when [parsing fee from flag, the zero fee is already removed](https://github.com/cosmos/cosmos-sdk/blob/e716e4103e934344aa7be6dc9b5c453bdec5f225/client/tx/factory.go#L144).
2. fee does not contain zero coins, min_gas_price does not contain zero coins, but global fee can be set to contain zero coins. Zero coins in global fee play the role of defining denoms.
3. the global fee and min_gas_price will be combined into one fee array by calling `CombinedFeeRequirement`

- [x] unit test for functions `CombinedFeeRequirement`, `IsAnyGTEIncludingZero`, `DenomsSubsetOfIncludingZero`
- [x] unit test for fee ante handler contains various edge cases, empty, zero coins.
- [ ] docs

More discussion please see [here](https://github.com/cosmos/gaia/pull/1570)


yaruwangway:
move to https://github.com/cosmos/gaia/pull/1609

PR #1610, opened 2022-07-26
ci: Update lint.yml

faddat:
The git diff thing causes linter errors to be grandfathered in. 

It hasn't been firing, and that's why we didn't know until now that the new configuration doesn't work at all. 

Time isn't a concern here, as codeql and e2e both take significantly longer.  We should lint on each commit, and we should not use diffs to determine weather or not we lint otherwise, it is likely we aren't actually linting.

* #1449 

linter runtime: 2m 19s


faddat:
please view the ci logs for the linter for the past two commits.

PR #1616, opened 2022-07-28
Cosmos Hub with ibc-go v5 and cosmos-sdk v0.46.0

faddat:
Ok, so --  this PR will cause the diffs in the CI system to trigger.

Then, we will see two kinds of issues:

1) grandfathered-in stuff
2) migration stuff -- all in e2e

the solution is outlined in #1449

PR #1623, opened 2022-08-01
add checkout for lint

okwme:
also present in #1616 good eye @faddat


okwme:
closing in favor of #1616

PR #1624, opened 2022-08-01
Update lint.yml

faddat:
Turns out that lint.yml wasn't checking out the code at all, so I've added the checkout.


okwme:
closing in favor of https://github.com/cosmos/gaia/pull/1623 (and incoming)

PR #1625, opened 2022-08-01
add back ibc e2e tests

okwme:


faddat:
like how you handled this!

okwme:
hmm, getting an error on E2E but it passed locally.
Will rebuild the docker image and see if i can reproduce the error locally

faddat:
my feeling: maybe it is building the docker image locally, but not here.  

I'll check out the container registry sometime next 24h, it mentions that it is failing.

okwme:
found the issue.
it passed locally because i was skipping all the other tests.
there's a gov test that checks whether the chain stops after a successful upgrade proposal and the chains are torn down and setup suite is called again but looks like with the addition of the second chain it's got a problem.

@glnro I added a commit where I used the `s.TearDownSuite()` instead of focusing on just the specific chain container like before. I think we need to tear it all down if we're going that route so that the second chain can create a new IBC channel. Can you confirm why the fund_community_pool test fails now? I have a feeling it's grabbing an account balance or an account address from the previous chain but I thought it was supposed to be independent of other tests?

Another option would be to just move the gov upgrade test to the very end of the suite and tear it all down after the chain halts.

glnro:
> fund_community_pool

I'm looking into this now, will make sure all tests can be run in any order.

PR #1626, opened 2022-08-01
remove now-unnecessary replace statement in go.mod

faddat:
We wrote the upgrade with a replace that referenced Notional's ibc-go v5 branch, but it's been merged, so this PR removes the replace.

### DIFF
--- a/ante/ante.go
+++ b/ante/ante.go
@@ -52,7 +52,7 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 		ante.NewSigGasConsumeDecorator(opts.AccountKeeper, sigGasConsumer),
 		ante.NewSigVerificationDecorator(opts.AccountKeeper, opts.SignModeHandler),
 		ante.NewIncrementSequenceDecorator(opts.AccountKeeper),
-		ibcante.NewAnteDecorator(opts.IBCkeeper),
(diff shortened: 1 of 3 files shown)


faddat:
* https://github.com/Gravity-Devs/liquidity/pull/20

and 

* https://github.com/cosmos/ibc-go/pull/1844


Will enable this PR to work right.  I'll refresh it when they've been merged.

okwme:
looks like that branch is different somehow, lots of tests failing when referencing it directly @faddat

faddat:
I'll make it this a draft until the upstream PR's are merged.  

This one has to do with arcane versioning:

at some point, all our code began to contain an import that exists nowhere, and it makes go very angry.  

99% likely, the branch in ibc-go now differs only in that we aren't using replace, which was changing around imports some.  

(eg: the difference is likely here)

faddat:
Here's an example of it working:

* https://github.com/Pylons-tech/pylons/pull/710

so basically ours is failing here, because we're trying to use workspaces for e2e

faddat:
@okwme - I wrote a bug in this.

Squashing it....

> looks like that branch is different somehow, lots of tests failing when referencing it directly @faddat



correct.

Thanks!

okwme (approved):

okwme:
thank you @faddat !

PR #1628, opened 2022-08-01
closes #1621

### LINKED ISSUE #1621: build: use distroless/static instead of gcr.io/distroless/cc
## Summary of Bug

The Dockerfile uses the base image https://github.com/cosmos/gaia/blob/91097a3cc85f78a96ecff3532eb25c3d728b120d/Dockerfile#L14

gcr.io/distroless/cc is a big image, not well maintained, dates of updated not even maintained but even more it isn't signed so at the mercy of whoever uploads it. For a quick comparison, please see this comparison
```shell
$ docker images | grep distro
distroless.dev/static                    latest      47fb349c19cc   23 hours ago    1.44MB
gcr.io/distroless/cc                     latest      af03863dbb6a   52 years ago    22.6MB
```

The current image is 22.6MB, claims it was last uploaded 52 years ago
Screen Shot 2022-07-29 at 7 41 44 PM

whereas distroless.dev/static is 1.44MB, uploaded nightly, signed, auditable and well maintained.

## Recommendation
Use distroless.dev/static and incorporate #1620 too to build static binaries linked against musl

Kindly ccing my colleagues at Orijtech @kirbyquerby @elias-orijtech @willpoint

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Is a spike necessary to map out how the issue should be approached?

okwme:
uses a smaller distroless container
closes #1621

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,5 +15,6 @@ FROM distroless.dev/static:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
 EXPOSE 26656 26657 1317 9090
+USER 0
 
 ENTRYPOINT ["gaiad", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN apk add --no-cache $PACKAGES
 RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
-FROM gcr.io/distroless/cc:$IMG_TAG
+FROM distroless.dev/static:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
 EXPOSE 26656 26657 1317 9090
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,7 @@ endif
 .PHONY: run-tests $(TEST_TARGETS)
 
 docker-build-debug:
-	@docker build -t cosmos/gaiad-e2e --build-arg IMG_TAG=debug -f e2e.Dockerfile .
+	@docker build -t cosmos/gaiad-e2e -f e2e.Dockerfile .
 
 # TODO: Push this to the Cosmos Dockerhub so we don't have to keep building it
 # in CI.
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -11,7 +11,7 @@ RUN apk add --no-cache $PACKAGES
 RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
-FROM gcr.io/distroless/cc:$IMG_TAG
+FROM distroless.dev/static:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
 EXPOSE 26656 26657 1317 9090
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -14,7 +14,6 @@ RUN CGO_ENABLED=0 make install
 FROM distroless.dev/static:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
-USER 0
 EXPOSE 26656 26657 1317 9090
 
 ENTRYPOINT ["gaiad", "start"]
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -14,6 +14,7 @@ RUN CGO_ENABLED=0 make install
 FROM distroless.dev/static:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
+USER 0
 EXPOSE 26656 26657 1317 9090
 
 ENTRYPOINT ["gaiad", "start"]
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strconv"
@@ -308,10 +309,12 @@ func (s *IntegrationTestSuite) runValidators(c *chain, portOffset int) {
 			Name:      val.instanceName(),
 			NetworkID: s.dkrNet.Network.ID,
 			Mounts: []string{
-				fmt.Sprintf("%s/:/root/.gaia", val.configDir()),
+				fmt.Sprintf("%s/:/home/nonroot/.gaia", val.configDir()),
 			},
 			Repository: "cosmos/gaiad-e2e",
 		}
+
+		s.Require().NoError(exec.Command("chmod", "-R", "0777", val.configDir()).Run())
 
 		// expose the first validator for debugging and communication
 		if val.index == 0 {
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -245,7 +245,7 @@ func (s *IntegrationTestSuite) initGenesis(c *chain) {
 	bz, err = tmjson.MarshalIndent(genDoc, "", "  ")
 	s.Require().NoError(err)
 
-	// write the updated genesis file to each validator
+	// write the updated genesis file to each validator.
 	for _, val := range c.validators {
 		err = writeFile(filepath.Join(val.configDir(), "config", "genesis.json"), bz)
 		s.Require().NoError(err)
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -130,7 +130,7 @@ func (s *IntegrationTestSuite) TestSendTokensFromNewGovAccount() {
 	proposalCounter++
 
 	s.T().Logf("Submitting Gov Proposal: Sending Tokens from Gov Module to Recipient")
-	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/root/.gaia/config/proposal_2.json")
+	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/home/nonroot/.gaia/config/proposal_2.json")
 	s.T().Logf("Depositing Gov Proposal: Sending Tokens from Gov Module to Recipient")
 	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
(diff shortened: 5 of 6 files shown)


okwme:
getting an issue with E2E where it doesn't seem to find the docker mounted directory where app.toml is stored and min-fees is a required setting.

this worked with `gcr.io/distroless/cc` and i can't figure out what is different in `distroless.dev/static` to make it fail

odeke-em:
@kirbyquerby from Orijtech Inc. is taking a look and is going to help out here @okwme; thank you for addressing this fast!

kirbyquerby:
After `docker inspect`ing the containers generated when building from main vs this PR, I noticed this diff:
```diff
152c152
<             "User": "0",
---
>             "User": "65532
```

It looks like [the new base image defaults to running as user `65532` rather than `0` (root)](https://github.com/distroless/static/blob/7e7572e578de7c51a2f1a1791f025cf315503aa2/apko.yaml#L16), so doesn't have read access to the volume-mounted directory. Adding this line to `e2e.Dockerfile` fixed the issue for me:
```dockerfile
USER 0
```

I also tried making edits to chmod 777 the directory, but couldn't seem to get it to work, but it could be a better change so that the container doesn't have to run as root :)

kirbyquerby:
We don't need root! Thanks to a comment from @kaniini , I realized that if we want to run as the non-root user, we also need to change the mount directory to `/home/nonroot/.gaia` ( as well as corresponding references to `root` / `root`'s home directory).

I'm able to get tests passing locally with this diff:


<details>
<summary>diff</summary>

```diff
diff --git a/tests/e2e/e2e_setup_test.go b/tests/e2e/e2e_setup_test.go
index f2b834968..13f6b8d67 100644
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strconv"
@@ -308,11 +309,13 @@ func (s *IntegrationTestSuite) runValidators(c *chain, portOffset int) {
 			Name:      val.instanceName(),
 			NetworkID: s.dkrNet.Network.ID,
 			Mounts: []string{
-				fmt.Sprintf("%s/:/root/.gaia", val.configDir()),
+				fmt.Sprintf("%s/:/home/nonroot/.gaia", val.configDir()),
 			},
 			Repository: "cosmos/gaiad-e2e",
 		}
 
+		s.Require().NoError(exec.Command("chmod", "-R", "0777", val.configDir()).Run())
+
 		// expose the first validator for debugging and communication
 		if val.index == 0 {
 			runOpts.PortBindings = map[docker.Port][]docker.PortBinding{
diff --git a/tests/e2e/e2e_test.go b/tests/e2e/e2e_test.go
index 24dd4d02f..b0c9537c2 100644
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -130,7 +130,7 @@ func (s *IntegrationTestSuite) TestSendTokensFromNewGovAccount() {
 	proposalCounter++
 
 	s.T().Logf("Submitting Gov Proposal: Sending Tokens from Gov Module to Recipient")
-	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/root/.gaia/config/proposal_2.json")
+	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/home/nonroot/.gaia/config/proposal_2.json")
 	s.T().Logf("Depositing Gov Proposal: Sending Tokens from Gov Module to Recipient")
 	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
 	s.T().Logf("Voting Gov Proposal: Sending Tokens from Gov Module to Recipient")
@@ -162,7 +162,7 @@ func (s *IntegrationTestSuite) TestGovSoftwareUpgrade() {
 	s.writeGovUpgradeSoftwareProposal(s.chainA, proposalHeight)
 
 	s.T().Logf("Submitting Gov Proposal: Software Upgrade")
-	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/root/.gaia/config/proposal_3.json")
+	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/home/nonroot/.gaia/config/proposal_3.json")
 	s.T().Logf("Depositing Gov Proposal: Software Upgrade")
 	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
 	s.T().Logf("Weighted Voting Gov Proposal: Software Upgrade")
@@ -211,14 +211,14 @@ func (s *IntegrationTestSuite) TestGovCancelSoftwareUpgrade() {
 	s.writeGovUpgradeSoftwareProposal(s.chainA, proposalHeight)
 
 	s.T().Logf("Submitting Gov Proposal: Software Upgrade")
-	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/root/.gaia/config/proposal_3.json")
+	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/home/nonroot/.gaia/config/proposal_3.json")
 	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
 	s.voteGovProposal(chainAAPIEndpoint, sender, proposalCounter, "yes", false)
 
 	proposalCounter++
 
 	s.T().Logf("Submitting Gov Proposal: Cancel Software Upgrade")
-	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/root/.gaia/config/proposal_4.json")
+	s.submitNewGovProposal(chainAAPIEndpoint, sender, proposalCounter, "/home/nonroot/.gaia/config/proposal_4.json")
 	s.depositGovProposal(chainAAPIEndpoint, sender, proposalCounter)
 	s.voteGovProposal(chainAAPIEndpoint, sender, proposalCounter, "yes", false)
 
@@ -254,7 +254,7 @@ func (s *IntegrationTestSuite) fundCommunityPool(chainAAPIEndpoint string, sende
 
 func (s *IntegrationTestSuite) submitLegacyProposalFundGovAccount(chainAAPIEndpoint string, sender string, proposalId int) {
 	s.Run("submit_legacy_community_spend_proposal_to_fund_gov_acct", func() {
-		s.execGovSubmitLegacyGovProposal(s.chainA, 0, chainAAPIEndpoint, sender, "/root/.gaia/config/proposal.json", fees.String(), "community-pool-spend")
+		s.execGovSubmitLegacyGovProposal(s.chainA, 0, chainAAPIEndpoint, sender, "/home/nonroot/.gaia/config/proposal.json", fees.String(), "community-pool-spend")
 
 		s.Require().Eventually(
 			func() bool {
diff --git a/tests/e2e/e2e_util_test.go b/tests/e2e/e2e_util_test.go
index e119fe788..853f82e55 100644
--- a/tests/e2e/e2e_util_test.go
+++ b/tests/e2e/e2e_util_test.go
@@ -28,7 +28,7 @@ func (s *IntegrationTestSuite) connectIBCChains() {
 		AttachStdout: true,
 		AttachStderr: true,
 		Container:    s.hermesResource.Container.ID,
-		User:         "root",
+		User:         "nonroot",
 		Cmd: []string{
 			"hermes",
 			"create",
@@ -77,7 +77,7 @@ func (s *IntegrationTestSuite) sendMsgSend(c *chain, valIdx int, from, to, amt,
 		AttachStdout: true,
 		AttachStderr: true,
 		Container:    s.valResources[c.id][valIdx].Container.ID,
-		User:         "root",
+		User:         "nonroot",
 		Cmd: []string{
 			"gaiad",
 			"tx",
@@ -137,7 +137,7 @@ func (s *IntegrationTestSuite) sendIBC(srcChainID, dstChainID, recipient string,
 		AttachStdout: true,
 		AttachStderr: true,
 		Container:    s.hermesResource.Container.ID,
-		User:         "root",
+		User:         "nonroot",
 		Cmd: []string{
 			"hermes",
 			"tx",
@@ -336,7 +336,7 @@ func (s *IntegrationTestSuite) executeGaiaTxCommand(ctx context.Context, c *chai
 				AttachStdout: true,
 				AttachStderr: true,
 				Container:    s.valResources[c.id][valIdx].Container.ID,
-				User:         "root",
+				User:         "nonroot",
 				Cmd:          gaiaCommand,
 			})
 			s.Require().NoError(err)
@@ -424,7 +424,7 @@ func (s *IntegrationTestSuite) getLatestBlockHeight(c *chain, valIdx int) int {
 				AttachStdout: true,
 				AttachStderr: true,
 				Container:    s.valResources[c.id][valIdx].Container.ID,
-				User:         "root",
+				User:         "nonroot",
 				Cmd:          []string{"gaiad", "status"},
 			})
 			s.Require().NoError(err)
```

</details>

With this patch, there are still 10 lines in the repository containing `/root`, many of which reference `/root/.hermes`. I didn't touch them, but they might need to be changed at some point as well.

kirbyquerby:
@okwme I think it would make more sense from a security perspective to go with the non-root approach. You can find the diff in my above comment or in https://gist.github.com/kirbyquerby/0b6be3491f27ff99cb4190577331af5f

To apply a diff, you can use `git apply`. To apply the non-root patch, you would run:
```bash
curl https://gist.githubusercontent.com/kirbyquerby/0b6be3491f27ff99cb4190577331af5f/raw/60a45047916fe9c577e3727bcdb3850c3cc0127a/patch.diff | git apply
```

okwme:
slick trick with the diff 😎
i've updated to be nonroot
thanks @kirbyquerby !

Dockerfile
```diff
@@ -11,9 +11,10 @@ RUN apk add --no-cache $PACKAGES
 RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
-FROM gcr.io/distroless/cc:$IMG_TAG
+FROM distroless.dev/static:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
 EXPOSE 26656 26657 1317 9090
+USER 0
```
odeke-em, line 18: Given that you've applied the patch @okwme, please remove this line as we want the code to run as non-root. After that this PR will be good to go and not have a high security surface :-) Thank you @okwme for jumping onto this ASAP.

okwme: actually the change was made in `e2e.Dockerfile`. I added the `USER 0` to `Dockerfile` because I couldn't find any use of `Dockerfile` in our code and am not sure why it was added originally or who is using it. Until I'm able to confirm that I'd like to preserve the behavior as much as possible and previously that docker image was run from root so i'd like it to stay that way for now. 

`e2e.Dockerfile` is the docker file we're using in our e2e and it now has the relevant changes.

odeke-em: Gotcha, cool!

PR #1644, opened 2022-08-02
DRAFT: 1643/E2E tests failing intermittently

okwme:
Still a WIP.
querying accounts to confirm the sequence but getting `acc %!v(PANIC=String method: interface conversion: interface {} is nil, not string)` instead of the expected account from the query.


okwme:
to test this PR build the docker image used for testing first then run the actual tests as follows:
```
make docker-build-debug
make test-e2e
```

okwme:
closing in favor of #1652

PR #1646, opened 2022-08-04
Revert "closes #1621"

### LINKED ISSUE #1621: build: use distroless/static instead of gcr.io/distroless/cc

okwme:
Reverts cosmos/gaia#1628

I'm unable to get e2e tests to run on my machine anymore and #1625 seems to be failing as a result of this change.
i'm reverting until it can be discovered why this is happening.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,9 @@ RUN apk add --no-cache $PACKAGES
 RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
-FROM distroless.dev/static:$IMG_TAG
+FROM gcr.io/distroless/cc:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
 EXPOSE 26656 26657 1317 9090
-USER 0
 
 ENTRYPOINT ["gaiad", "start"]
--- a/Makefile
+++ b/Makefile
@@ -197,7 +197,7 @@ endif
 .PHONY: run-tests $(TEST_TARGETS)
 
 docker-build-debug:
-	@docker build -t cosmos/gaiad-e2e -f e2e.Dockerfile .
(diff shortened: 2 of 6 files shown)


okwme:
@kirbyquerby or @odeke-em have any ideas here?
you can see the e2e tests failing on this PR: https://github.com/cosmos/gaia/runs/7670076628?check_suite_focus=true
while there are no changes that should have impacted the docker build.
they were passing here before i merged main into the branch: https://github.com/cosmos/gaia/runs/7656634962?check_suite_focus=true

kirbyquerby:
@okwme Looks like the issue was that my proposed diff in #1628 accidentally updated some of the commands sent to the hermes container as well, which still runs as root. The issue only showed up in #1625 because before that, the hermes portion of the tests was disabled. I've mailed https://github.com/cosmos/gaia/pull/1648 with a fix.

PR #1648, opened 2022-08-04
use distroless.dev/static base image

kirbyquerby:
This change is a redo of https://github.com/cosmos/gaia/pull/1628 with fixes for the failed tests in https://github.com/cosmos/gaia/pull/1625 . Specifically, the hermes container still runs as `root`, so commands need to be sent to it as `root` rather than `nonroot`.

Updates https://github.com/cosmos/gaia/issues/1621

/cc @okwme @odeke-em

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,9 +11,10 @@ RUN apk add --no-cache $PACKAGES
 RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
-FROM gcr.io/distroless/cc:$IMG_TAG
+FROM distroless.dev/static:$IMG_TAG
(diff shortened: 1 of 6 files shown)


okwme:
thanks @kirbyquerby !
could you check the linting errors?

kirbyquerby:
The lint errors were in files I didn't touch :'(
I've rebased on the latest main, which should fix things as the lint errors appear to have been fixed in #1651

mmulji-ic (approved):

mmulji-ic:
Thanks @kirbyquerby @odeke-em

PR #1615, opened 2022-07-27
docs: all fees

yaruwangway:
a doc for fees and fee antehandler.

PR #1657, opened 2022-08-08
docs starter PR

okwme:
starting point for home for gaia module docs


okwme:
we can close in light of #1615

PR #1659, opened 2022-08-09
chore: Update codeowners for Primary & Governance sections

mmulji-ic:


okwme:
looks like there might be an issue with `@LexaMichaelides` either not the right username or doesn't have write permission on the repo i think. Did you get reply from @ebuchman and @mariapao ?

mmulji-ic:
@okwme , asked @uditvira to add Lexa to his team. Am checking with Maria and Ethan atm. Once confirmed, will merge or append as required.

mmulji-ic:
Maria and Ethan approved removal.

Issue #1630, opened 2022-08-01
Add Staking E2E Tests


okwme:
#### Testing
- [ ] Unit Tests
- [ ] E2E Tests
	- [ ] [User Flows](#User-Flows)
- [ ] Contract Tests
	- [ ] [REST](#REST-Endpoints)
	- [ ] [GRPC](#GRPC-Endpoints)
	- [ ] [CLI](#CLI)
- [ ] [Simulations](https://github.com/cosmos/hub-eng/tree/main/tools/testing/scripts/simulations/gov)

#### Docs
- [ ] [Link]() to SDK tutorial
- [x] [Link](https://docs.cosmos.network/v0.46/modules/staking/) to SDK API Docs
- [ ] Update on Gaia Module Page

##### User Flows
- [ ] [Scenario 1](): A user wants to delegate tokens to a known validator
- [ ] [Scenario 2](): A user wants to redelegate tokens that are staked to a new validator
- [ ] [Scenario 3](): A user wants to unbond tokens that are staked
- [ ] [Scenario 4](): A user wants to cancel the unbonding and delegate tokens to original validator

##### REST Endpoints
- [ ] [delegations by delegator address](https://docs.cosmos.network/v0.46/modules/staking/09_client.html#delegatordelegations-2) `/cosmos/staking/v1beta1/delegations/{delegatorAddr}`
- [ ] [redelegations by delegator address](https://docs.cosmos.network/v0.46/modules/staking/09_client.html#redelegations-3) `/cosmos/staking/v1beta1/delegators/{delegatorAddr}/redelegations`
- [ ] [unbonding delegations by delegator address](https://docs.cosmos.network/v0.46/modules/staking/09_client.html#delegatorunbondingdelegations-2) `/cosmos/staking/v1beta1/delegators/{delegatorAddr}/unbonding_delegations`
- [ ] [validators by delegator address](https://docs.cosmos.network/v0.46/modules/staking/09_client.html#delegatorvalidators-2) `/cosmos/staking/v1beta1/delegators/{delegatorAddr}/validators`
- [ ] [validator by validator and delegator address](https://docs.cosmos.network/v0.46/modules/staking/09_client.html#delegatorvalidator-2) `/cosmos/staking/v1beta1/delegators/{delegatorAddr}/validators/{validatorAddr}`
[...]

PR #1673, opened 2022-08-16
test: Staking module e2e tests

glnro:
This PR fulfills basic test commitment for #1630 and introduces a new suite of tests for the staking module.

CLI commands tested
- `keys`
  - `add` Add new key
- `tx staking`
  - `delegate` Delegate staked tokens to a validator
  - `redelegate` Redelegate previously staked tokens to a new validator

REST endpoints tested
- [Delegation](https://docs.cosmos.network/v0.46/modules/staking/09_client.html#delegation-3) `/cosmos/staking/v1beta1/validators/{validatorAddresess}/delegations/{delegatorAddress}` 

User Flows
- A user wants to delegate tokens to a known validator
- A user wants to redelegate tokens that are staked to a new validator

This PR also includes a few refactors related to minor formatting in the group tests.

### DIFF
--- a/tests/e2e/e2e_cli_util.go
+++ b/tests/e2e/e2e_cli_util.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
+	staketypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	"io"
 	"net/http"
 	"strings"
@@ -104,24 +105,23 @@ func queryGaiaDenomBalance(endpoint, addr, denom string) (sdk.Coin, error) {
 }
 
 func queryGovProposal(endpoint string, proposalID int) (govv1beta1.QueryProposalResponse, error) {
-	var emptyProp govv1beta1.QueryProposalResponse
+	var govProposalResp govv1beta1.QueryProposalResponse
 
 	path := fmt.Sprintf("%s/cosmos/gov/v1beta1/proposals/%d", endpoint, proposalID)
 
 	resp, err := http.Get(path) //nolint:gosec // this is only used during tests
 	if err != nil {
-		return emptyProp, fmt.Errorf("failed to execute HTTP request: %w", err)
+		return govProposalResp, fmt.Errorf("failed to execute HTTP request: %w", err)
 	}
 	defer resp.Body.Close()
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return emptyProp, err
+		return govProposalResp, err
 	}
 
-	var govProposalResp govv1beta1.QueryProposalResponse
 	if err := cdc.UnmarshalJSON(body, &govProposalResp); err != nil {
-		return emptyProp, err
+		return govProposalResp, err
 	}
 
 	return govProposalResp, nil
@@ -145,4 +145,25 @@ func queryGlobalFees(endpoint string) (amt sdk.DecCoins, err error) {
 	}
 
 	return fees.MinimumGasPrices, nil
+}
+
+func queryDelegation(endpoint string, validatorAddr string, delegatorAddr string) (staketypes.QueryDelegationResponse, error) {
+	var delegationRes staketypes.QueryDelegationResponse
+
+	resp, err := http.Get(fmt.Sprintf("%s/cosmos/staking/v1beta1/validators/%s/delegations/%s", endpoint, validatorAddr, delegatorAddr)) //nolint:gosec // this is only used during tests
+	if err != nil {
+		return delegationRes, fmt.Errorf("failed to execute HTTP request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return delegationRes, err
+	}
+
+	if err := cdc.UnmarshalJSON(body, &delegationRes); err != nil {
+		return delegationRes, err
+	}
+
+	return delegationRes, nil
 }
--- a/tests/e2e/e2e_cli_util.go
+++ b/tests/e2e/e2e_cli_util.go
@@ -3,7 +3,6 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
-	staketypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	"io"
 	"net/http"
 	"strings"
@@ -11,6 +10,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
+	staketypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 
 	globalfee "github.com/cosmos/gaia/v8/x/globalfee/types"
 )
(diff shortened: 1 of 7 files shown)


tests/e2e/e2e_cli_util.go
```diff
@@ -104,24 +105,23 @@ func queryGaiaDenomBalance(endpoint, addr, denom string) (sdk.Coin, error) {
...
 func queryGovProposal(endpoint string, proposalID int) (govv1beta1.QueryProposalResponse, error) {
-	var emptyProp govv1beta1.QueryProposalResponse
+	var govProposalResp govv1beta1.QueryProposalResponse
 
 	path := fmt.Sprintf("%s/cosmos/gov/v1beta1/proposals/%d", endpoint, proposalID)
 
 	resp, err := http.Get(path) //nolint:gosec // this is only used during tests
 	if err != nil {
-		return emptyProp, fmt.Errorf("failed to execute HTTP request: %w", err)
+		return govProposalResp, fmt.Errorf("failed to execute HTTP request: %w", err)
```
mmulji-ic, line 114: Would've preferred to keep the group stuff in a separate PR

tests/e2e/e2e_setup_test.go
```diff
@@ -55,7 +55,7 @@ const (
 
 var (
 	stakeAmount       = math.NewInt(100000000000)
-	stakeAmountCoin   = sdk.NewCoin("stake", stakeAmount)
+	stakeAmountCoin   = sdk.NewCoin(uatomDenom, stakeAmount)
```
mmulji-ic: Mismatch in denoms? uatom/stake?

tests/e2e/e2e_groups_test.go
```diff
@@ -54,33 +52,30 @@ var (
 )
 
 func (s *IntegrationTestSuite) GroupsSendMsgTest() {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
-	defer cancel()
-
 	chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
-	s.setup(ctx)
+	s.setup()
```
mmulji-ic, line 56: Does .setup() default to without timeout?

tests/e2e/e2e_cli_util.go
glnro: Yeah started to get ahead of myself, I'll follow up with a more thorough refactor pr

tests/e2e/e2e_setup_test.go
glnro: I changed the bonding denom to uatom to keep everything in the test consistent...tweaked the variable names to reflect it's the amount for staking not a stake denom amount.

tests/e2e/e2e_cli_util.go
```diff
@@ -3,6 +3,7 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
+	staketypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	"io"
```
yaruwangway: formatting:
staketypes "github.com/cosmos/cosmos-sdk/x/staking/types" is at last with line space.

tests/e2e/genesis.go
```diff
@@ -118,13 +119,21 @@ func modifyGenesis(path, moniker, amountStr string, accAddr sdk.AccAddress, glob
...
+	}
+	appState[staketypes.ModuleName] = stakingGenStateBz
+
 	// Refactor to separate method
 	amnt := math.NewInt(10000)
 	quorum, _ := sdk.NewDecFromStr("0.000000000000000001")
 	threshold, _ := sdk.NewDecFromStr("0.000000000000000001")
 
 	govState := govv1beta1.NewGenesisState(1,
-		govv1beta1.NewDepositParams(sdk.NewCoins(sdk.NewCoin("uatom", amnt)), 10*time.Minute),
```
yaruwangway, line 127: why pass denom as argument. since amt is hard-coded here, `amnt := math.NewInt(10000)` can just hard code denom

tests/e2e/e2e_staking_test.go
```diff
@@ -0,0 +1,65 @@
...
+	validatorB := s.chainA.validators[1]
+	sender, err := validatorA.keyInfo.GetAddress()
+	s.NoError(err)
+	validatorBAddr, err := validatorB.keyInfo.GetAddress()
+	s.NoError(err)
+
+	valOperA := sdk.ValAddress(sender)
+	valOperB := sdk.ValAddress(validatorBAddr)
+
+	alice := s.executeGKeysAddCommand(s.chainA, 0, "alice", home)
```
yaruwangway, line 27: we can refactor here later, i added an addr to genesis as ica owner, who  already has some tokens. so that we can simplify here that no need to add keys, and fund alice, can use ica onwer as delegator. but this solution is also good!

tests/e2e/genesis.go
```diff
@@ -3,6 +3,7 @@ package e2e
 import (
 	"encoding/json"
 	"fmt"
+	staketypes "github.com/cosmos/cosmos-sdk/x/staking/types"
```
yaruwangway: formatting

tests/e2e/e2e_groups_test.go
```diff
@@ -246,8 +228,8 @@ func (s *IntegrationTestSuite) setup(ctx context.Context) {
 	s.Require().NoError(err)
 	aliceAddr = alice.String()
 
-	bobAddr = s.executeGaiaKeysAddCommand(ctx, s.chainA, 0, "bob")
-	charlieAddr = s.executeGaiaKeysAddCommand(ctx, s.chainA, 0, "charlie")
+	bobAddr = s.executeGKeysAddCommand(s.chainA, 0, "bob", "/home/nonroot/.gaia")
```
mmulji-ic, line 231: Hard coding directory, when above in dataDirectoryHome you avoided.

tests/e2e/e2e_util_test.go
```diff
@@ -863,3 +878,103 @@ func (s *IntegrationTestSuite) executeGaiaKeysAddCommand(ctx context.Context, c
...
+func (s *IntegrationTestSuite) executeKeysList(c *chain, valIdx int, home string) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	var (
+		outBuf bytes.Buffer
+		errBuf bytes.Buffer
+	)
+
+	gaiaCommand := []string{
```
mmulji-ic, line 891: This kind of operation I've isolated in a helper fn, seems unnecessary to repeat, why not adapt executeGaiaTxCommand

glnro: Yeah so there's the main helper function `executeGaiaTxCommand` and the difference in this case is it queries the transaction hash after executing to confirm the transaction didn't fail...we could refactor this further to a general purpose function that know whether it's executing a tx or keys command but I fear refactoring too much at this point, lmk what you think if we should go ahead and just do this

tests/e2e/genesis.go
glnro: either or I just thought it would be cleaner to refactor to using the denom const everywhere as we go

tests/e2e/e2e_staking_test.go
glnro: Sure, I'll refactor to this way in the `x/dist` test

PR #1687, opened 2022-08-23
feat: add accounts to genesis in e2e tests

yaruwangway:
- [x] add accounts to genesis with balances 
- [x] use normal accounts as relayer wallet rather than validators
- [x] bump hermes version to v1.0.0-rc2

### DIFF
--- a/tests/e2e/e2e_setup_test.go
+++ b/tests/e2e/e2e_setup_test.go
@@ -172,6 +172,7 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	s.initValidatorConfigs(s.chainB)
 	s.runValidators(s.chainB, 10)
 
+	time.Sleep(10 * time.Second)
 	s.runIBCRelayer()
 }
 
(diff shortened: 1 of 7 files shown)


tests/e2e/validator.go
```diff
@@ -41,6 +41,13 @@ type validator struct {
 	nodeKey          p2p.NodeKey
 }
 
+type account struct {
```
glnro, line 44: I might differentiate this as a local account, but then again I suppose it's only accessible from the chain

tests/e2e/e2e_setup_test.go
```diff
@@ -207,17 +208,27 @@ func (s *IntegrationTestSuite) TearDownSuite() {
 
 func (s *IntegrationTestSuite) initNodes(c *chain) {
 	s.Require().NoError(c.createAndInitValidators(2))
-
+	// add two more addr to val0 dir, one addr is used as relayer wallet, one is used as ica owner
+	s.Require().NoError(c.addAccountFromMnemonic(2))
 	// initialize a genesis file for the first validator
 	val0ConfigDir := c.validators[0].configDir()
+	addrAll := []sdk.AccAddress{}
```
glnro, line 215:
```suggestion
	localAddresses := []sdk.AccAddress{}
```

tests/e2e/validator.go
```diff
@@ -167,6 +174,49 @@ func (v *validator) createKeyFromMnemonic(name, mnemonic string) error {
...
+	}
+
+	keyringAlgos, _ := kb.SupportedAlgorithms()
+	algo, err := keyring.NewSigningAlgoFromString(string(hd.Secp256k1Type), keyringAlgos)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < counts; i++ {
+		name := fmt.Sprintf("acct-%d", i)
```
glnro, line 191: To avoid confusion I might not just name each one by number...for example the wallet relayer account should be separate than alternative accounts we might want to use just for testing purposes...lmk what you think

glnro (approved):

tests/e2e/validator.go
yaruwangway: all accounts are local right ? what else account type are there ?

tests/e2e/validator.go
yaruwangway: sounds good, can add this to refactor. i have the same code in my  e2e branch now.

tests/e2e/e2e_setup_test.go
```diff
@@ -172,6 +172,7 @@ func (s *IntegrationTestSuite) SetupSuite() {
 	s.initValidatorConfigs(s.chainB)
 	s.runValidators(s.chainB, 10)
 
+	time.Sleep(10 * time.Second)
```
mmulji-ic, line 175: Do we want some kind of facility where we retry the request with increasing backoff? Would be nicer & more efficient than playing with the time outs & sleeping. If this is interesting, then it's a new testing feature.

mmulji-ic (approved):
Remember to add a new test issue if we think the backoff mechanism is useful.

PR #1698, opened 2022-08-26
feature: implement ante handle for min commission

### LINKED ISSUE #1697: Minimum Commission for proposal 76
## Summary

If proposal 76 passes, validator min commission should be enforced onchain.

## Problem Definition

Right now MsgEditValidator and MsgCreateValidator allow any commission value, we should check that c.Rate is above the amount decided by governance.

## Proposal

- Write AnteHandle for minimum commission check
- Write Upgrade Handler for validator commission update

____

#### For Admin Use

- [ ] Not duplicate issue

MiB-Intern:
If proposal 76 passes, validator min commission should be enforced onchain.

Right now MsgEditValidator and MsgCreateValidator allow any commission value, we should check that c.Rate is above the amount decided by governance.

-- 

Need to add something like this to the next upgrade handler
```golang
// force an update of validator min commission
validators := app.StakingKeeper.GetAllValidators(ctx)
		
minCommissionRate := sdk.NewDecWithPrec(5, 2)
for _, v := range validators {
    if v.Commission.Rate.LT(minCommissionRate) {
        if v.Commission.MaxRate.LT(minCommissionRate) {
            v.Commission.MaxRate = minCommissionRate
        }

    v.Commission.Rate = minCommissionRate
    v.Commission.UpdateTime = ctx.BlockHeader().Time

    // call the before-modification hook since we're about to update the commission
    app.StakingKeeper.BeforeValidatorModified(ctx, v.GetOperator())

    app.StakingKeeper.SetValidator(ctx, v)
    }
}
```

Closes #1697 

Original code by giansalex, Osmosis and Juno teams

### DIFF
--- a/ante/ante.go
+++ b/ante/ante.go
@@ -1,6 +1,7 @@
 package ante
 
 import (
+	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/x/auth/ante"
@@ -16,6 +17,7 @@ type HandlerOptions struct {
 	IBCkeeper            *ibckeeper.Keeper
 	BypassMinFeeMsgTypes []string
 	GlobalFeeSubspace    paramtypes.Subspace
+	Cdc                  codec.BinaryCodec
 }
 
 func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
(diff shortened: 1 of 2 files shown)


faddat:
Thanks very much!

I think that we should merge this, even if 76 doesn't pass, but with the parameters set to zero.

```go
minCommissionRate := sdk.NewDecWithPrec(0, 2)
```

This would give gaia more flexibility in the future.

mmulji-ic:
> Thanks very much!
> 
> I think that we should merge this, even if 76 doesn't pass, but with the parameters set to zero.
> 
> ```go
> minCommissionRate := sdk.NewDecWithPrec(0, 2)
> ```
> 
> This would give gaia more flexibility in the future.

@okwme any opinion?

okwme:
Thanks for the contribution @MiB-Intern !
I'm going to close it for now in light of current priorities but we can always re-open if the issue goes to vote again.

PR #1703, opened 2022-08-30
Update number of validators from 150 to 175

simonguenther:


mmulji-ic (approved):

mmulji-ic:
Thanks for the update @simonguenther

PR #1706, opened 2022-08-30
Cross-site Scripting in Prism

imhunterand:
## Describe the bugs: 🐛
A Cross-site scripting attack was found in Prism. The command-line plugin did not properly escape its output. This issue leads to the input text being inserted into the Document Object Model (DOM) as HTML code, which can be exploited by an attacker.

Prism is a syntax highlighting library. Starting with version 1.14.0 and prior to version 1.27.0, Prism's command line plugin can be used by malicious users to achieve a cross-site scripting attack. The command line plugin did not properly escape its output, leading to the input text being inserted into the DOM as HTML code. Server-side usage of Prism is not impacted. Websites that do not use the Command Line plugin are also not impacted. This bug has been fixed in v1.27.0. As a workaround, do not use the command line plugin on untrusted inputs, or sanitize all code blocks (remove all HTML code text) from all code blocks that use the command line plugin.

**CVE-2022-23647**
`CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:C/C:H/I:L/A:L`
GHSA-3949-f494-cm99

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -9312,9 +9312,9 @@
       }
     },
     "node_modules/prismjs": {
-      "version": "1.25.0",
-      "resolved": "https://registry.npmjs.org/prismjs/-/prismjs-1.25.0.tgz",
-      "integrity": "sha512-WCjJHl1KEWbnkQom1+SzftbtXMKQoezOCYs5rECqMN+jP+apI7ftoflyqigqzopSO3hMhTEb0mFClA8lkolgEg=="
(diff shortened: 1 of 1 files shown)


mmulji-ic (approved):

imhunterand:
Hi! Thank you for your respond and approved could please mark my reports as Resolved for appreciated the solved.
The refferences as [#1691667](https://hackerone.com/reports/1691667)

Best regars,
@imhunterand

PR #1711, opened 2022-08-31
Code Injection via Object Prototype 

imhunterand:
## Describe the bugs: 🐛
The json-schema Node.JS library was vulnerable to prototype pollution during the validation of a JSON object. An attacker, able to provide a specially crafted JSON file for validation, could use this flaw to modify the behavior of the node program, to, for example, execute arbitrary code. json-schema before version 0.4.0 is vulnerable to Improperly Controlled Modification of Object Prototype Attributes ('Prototype Pollution').

**Details:**
Prototype Pollution is a vulnerability affecting JavaScript. Prototype Pollution refers to the ability to inject properties into existing JavaScript language construct prototypes, such as objects. JavaScript allows all Object attributes to be altered, including their magical attributes such as `_proto_`, `constructor` and `prototype`. An attacker manipulates these attributes to overwrite, or pollute, a JavaScript application object prototype of the base object by injecting other values. Properties on the `Object.prototype` are then inherited by all the JavaScript objects through the prototype chain. When that happens, this leads to either denial of service by triggering JavaScript exceptions, or it tampers with the application source code to force the code path that the attacker injects, thereby leading to remote code execution.

There are two main ways in which the pollution of prototypes occurs:
  * Unsafe `Object` recursive merge
  * Property definition by path

Unsafe Object recursive merge
The logic of a vulnerable recursive merge function follows the following high-level model:
```js
merge (target, source)
foreach property of source

if property exists and is an object on both the target and the source merge(target[property], source[property]) else target[property] = source[property]
```
When the source object contains a property named `_proto_` defined with `Object.defineProperty()` , the condition that checks if the property exists and is an object on both the target and the source passes and the merge recurses with the target, being the prototype of Object and the source of `Object` as defined by the attacker. Properties are then copied on the `Object` prototype.

Clone operations are a special sub-class of unsafe recursive merges, which occur when a recursive merge is conducted on an empty object: `merge({},source).` `lodash` and `Hoek` are examples of libraries susceptible to recursive merge attacks.

## Property definition by path
There are a few JavaScript libraries that use an API to define property values on an object based on a given path. The function that is generally affected contains this signature: `theFunction(object, path, value)` If the attacker can control the value of “path”, they can set this value to `_proto_.myValue`. `myValue` is then assigned to the prototype of the class of the object.

**How to prevent**
  * Freeze the prototype— use `Object.freeze (Object.prototype)`.
  * Require schema validation of JSON input.
  * Avoid using unsafe recursive merge functions.
  * Consider using objects without prototypes (for example, `Object.create(null)`), breaking the prototype chain and preventing pollution.
  * As a best practice use `Map` instead of `Object`.

**CVE-2021-3918**
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`
GHSA-896r-f27r-55mw

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -7152,9 +7152,9 @@
       "integrity": "sha512-mrqyZKfX5EhL7hvqcV6WG1yYjnjeuYDzDhhcAAUrq8Po85NBQBJP+ZDUT75qZQ98IkUoBqdkExkukOU7Ts2wrw=="
     },
     "node_modules/json-schema": {
-      "version": "0.2.3",
-      "resolved": "https://registry.npmjs.org/json-schema/-/json-schema-0.2.3.tgz",
-      "integrity": "sha1-tIDIkuWaLwWVTOcnvT8qTogvnhM="
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/json-schema/-/json-schema-0.4.0.tgz",
+      "integrity": "sha512-es94M3nTIfsEPisRafak+HDLfHXnKBhV3vU5eqPcS3flIWqcxJWgXHXiey3YrpaNsanY5ei1VoYEbOzijuq9BA=="
     },
     "node_modules/json-schema-traverse": {
       "version": "0.4.1",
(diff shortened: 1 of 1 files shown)


okwme:
hello this has conflicts can you take another look?

mmulji-ic:
@imhunterand we're looking at other options for our documentation and this issue is non-critical for our core application.
@okwme think for the time being we leave all the vue stuff on hold.
Reopen and ping if necessary.

Issue #1424, opened 2022-04-19
Integrate Gov module updates


mmulji-ic:
#### Implementation
- [x] Integration into app.go

#### Testing
- [ ] Unit Tests
- [ ] E2E Tests
	- [ ] [User Flows](#User-Flows)
- [ ] Contract Tests
	- [ ] [REST](#REST-Endpoints)
	- [ ] [GRPC](#GRPC-Endpoints)
	- [ ] [CLI](#CLI)
- [ ] [Simulations](https://github.com/cosmos/hub-eng/tree/main/tools/testing/scripts/simulations/gov)

#### Docs
- [ ] [Link]() to SDK tutorial
- [x] [Link](https://docs.cosmos.network/v0.46/modules/gov/07_client.html) to SDK API Docs
- [ ] Update on Gaia Module Page

##### User Flows
- [x] [Scenario 1](https://github.com/cosmos/gaia/blob/main/tests/e2e/e2e_test.go#L106): The community wants to reward an individual for work, so a Gov proposal is submitted and voted on to both fund the Gov account via the Community Pool and to send tokens from the Gov account to the recipient.
  - Submit & pass _legacy_ proposal to fund gov acocunt via community pool
    1. Submit proposal
    2. Deposit proposal
    3. Vote proposal
  - Submit & pass proposal to fund recipient account
    1. Submit proposal
    2. Deposit proposal
    3. Vote proposal

- [x] [Senario 2](https://github.com/cosmos/hub-eng/blob/main/tools/testing/scripts/simulations/gov/gov.sh#L18): The community wants to increase the number of participating validators, so it proposes a legacy param change proposal
  - Submit & pass _legacy_ param change proposal
    1. Submit proposal
    2. Deposit proposal
    3. Vote proposal

- [x] [Scenario 3](https://github.com/cosmos/gaia/blob/main/tests/e2e/e2e_test.go#L148): A new release of Gaia is ready, so the community submits a proposal to upgrade the chain
  - Submit & pass new gov proposal executing `MsgSoftwareUpgrade` message.
    1. Submit proposal
    2. Deposit proposal
    3. Vote proposal

- [x] [Scenario 4](https://github.com/cosmos/gaia/blob/main/tests/e2e/e2e_test.go#L189): The community decides to cancel a pending upgrade and submits a proposal to cancel the upgrade
  - Submit & pass new gov proposal executing `MsgSoftwareUpgrade` message.
    1. Submit proposal
    2. Deposit proposal
[...]

PR #1718, opened 2022-08-31
docs: Update gov module docs

glnro:
This PR fulfills the documentation component of #1424 .


hxrts:
This is really great. My only question is to what extent do we want to be maintaining information about gov module behavior in the gaia repo. If we've modified defaults it certainly makes sense, but if there's every a change to gov we'll need to track that and update gaia docs.

mmulji-ic (approved):
Great Job @glnro

LexaMichaelides (approved):
Looks great!

okwme:
looks like broken links
`https://docs.tendermint.com/v0.35/tendermint-core/block-sync/` could be changed to
`https://github.com/tendermint/tendermint/blob/v0.34.x/spec/p2p/messages/block-sync.md`

`https://docs.tendermint.com/master/tendermint-core/state-sync/#` could be changed to `https://github.com/tendermint/tendermint/blob/v0.34.x/spec/p2p/messages/state-sync.md`

okwme:
is this supposed to be targeting `okwme/docs-home` ?

mmulji-ic:
> is this supposed to be targeting `okwme/docs-home` ?

Yes, should be doc updates only for Rho, which should get merged into main / release branch.

PR #1723, opened 2022-09-05
refactor: Removed test target, added convenience docker-all target & removed ...

### LINKED ISSUE #1489: make targets missing or fail
## Summary of Bug

`make all` fails with ... make: *** No rule to make target `test', needed by `all'.  Stop.
`make start-localnet-ci` fails with ... unknown flag: --mode make: *** [start-localnet-ci] Error 1

#### For Admin Use

- [ ] Not duplicate issue

mmulji-ic:
… unused target for contracts


Makefile
```diff
@@ -221,7 +223,7 @@ format:
 ###                                Localnet                                 ###
 ###############################################################################
 
-start-localnet-ci:
+start-localnet-ci: build
 	./build/gaiad init liveness --chain-id liveness --home ~/.gaiad-liveness
```
mmulji-ic, line 224: Would like to add a `rm -rf here ~/.gaiad-liveness`, opinions?

yaruwangway: maybe not ? in case people still want to keep the data.  let them manually `rm`

okwme: liveness test shouldn't ever be persisted though. it's just to make sure that blocks are produced. i'm ok with removing as long as we're sure there is no chance liveness tests are being configured with custom home directories or anything like that, right?

mmulji-ic: We do use a custom home, but this target is really just for the CI & occasional local testing

okwme: ok let's add the `rm`

Makefile
```diff
@@ -249,4 +248,4 @@ test-docker-push: test-docker
 	@docker push ${TEST_DOCKER_REPO}:latest
```
mmulji-ic, line 248: @okwme do we still want this two docker commands that rely on contrib/Dockerfile.test?
The docker push, pushes 3 things, latest, branch name & commit hash. Think branch name/latest are ok.

okwme: i'm ok with keeping as is if there's no large cost, in time or money, in doing so

okwme (approved):

Makefile
```diff
@@ -221,15 +219,16 @@ format:
...
+	rm -rf ~/.gaiad-liveness
 	./build/gaiad init liveness --chain-id liveness --home ~/.gaiad-liveness
 	./build/gaiad config chain-id liveness --home ~/.gaiad-liveness
 	./build/gaiad config keyring-backend test --home ~/.gaiad-liveness
 	./build/gaiad keys add val --home ~/.gaiad-liveness
 	./build/gaiad add-genesis-account val 10000000000000000000000000stake --home ~/.gaiad-liveness --keyring-backend test
 	./build/gaiad gentx val 1000000000stake --home ~/.gaiad-liveness --chain-id liveness
 	./build/gaiad collect-gentxs --home ~/.gaiad-liveness
-	sed -i'' 's/minimum-gas-prices = ""/minimum-gas-prices = "0uatom"/' ~/.gaiad-liveness/config/app.toml
+	sed -i.bak'' 's/minimum-gas-prices = ""/minimum-gas-prices = "0uatom"/' ~/.gaiad-liveness/config/app.toml
```
mmulji-ic, line 231: Works on Mac, if theres a bak file created.

PR #1738, opened 2022-09-13
feat(e2e): add e2e tests for the vesting module

### LINKED ISSUE #1631: Add Vesting E2E Tests
## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue

Pantani:
close #1631

## Description

Add an end-to-end test for the vesting module. 

- `Delayed` and `Continuous` vesting accounts were created inside the genesis file;
- `Permanent Locked` and `Periodic` vesting accounts are created on demand using the `tx vesting` CLI command;
- Vesting period file for the periodic vesting account is located inside the `<chain_home>/test_period.json`;
- All vesting account tests are running in parallel;
- The same mnemonic is used for `Delayed` and `Continuous` vesting accounts, but different HD paths and wallets.
- I tried to make less refactoring possible, but some parts of the code need refactoring;

#### Delayed vesting account

 - Create a delayed vesting account
 	- Balance should be equal to `original vesting coins` + non-vesting balance
 	- Delegate coins should succeed
  	- Transfer coins should fail

 - Wait until the `EndTime` reach
 	- Transfer coins should succeed

#### Continous vesting account

 - Create a continuous vesting account
 	- Balance should be equal to `original vesting coins` + non-vesting balance
 	- Delegate coins should succeed
 	- Transfer coins should fail
  
 - Wait until the `StartTime` reach
 	- Transfer coins should fail
 
 - Wait until the `EndTime` reach
 	- Transfer all coins should succeed

 	
#### Permanently locked account

 - Create a permanently locked account
 	- Balance should be equal to `original vesting coins` + non-vesting balance
  	- Delegate coins should succeed
 	- Transfer coins should fail

### Periodic vesting account

 - Create a periodic vesting account
 	- Balance should be equal to `original vesting coins` + non-vesting balance
 	- Transfer coins should fail
  
 - Wait until the first-period reach
 	- Delegate coins should succeed
 	- Transfer all coins should succeed
 
 - Wait until the next period reach
 	- Transfer all coins should succeed


tests/e2e/e2e_util_test.go
```diff
@@ -949,6 +959,20 @@ func (s *IntegrationTestSuite) executeDelegate(c *chain, valIdx int, endpoint st
 
 	s.executeGaiaTxCommand(ctx, c, gaiaCommand, valIdx, endpoint)
 	s.T().Logf("%s successfully delegated %s to %s", delegatorAddr, amount, valOperAddress)
+
+	// Validate delegation successful
+	s.Require().Eventually(
```
glnro: My concern with moving this into the execute function is it adds the responsibility of verifying the execution of the cli command...this is fine if we are looking to verify a specific positive outcome but makes the utility function less extensible for other testing scenarios. Personal opinion is to move it back out and keep it single responsibility like the other exec functions.

okwme:
uhoh, looks like the updates to e2e have created some conflicts @Pantani

tests/e2e/e2e_exec_test.go
```diff
@@ -14,6 +15,98 @@ import (
...
+		o[key] = value
+	}
+}
+
+func applyOptions(chainID, home string, options []flagOption) map[string]interface{} {
+	opts := map[string]interface{}{
+		flagKeyringBackend: "test",
+		flagOutput:         "json",
+		flagGas:            "auto",
+		flagFrom:           "alice",
```
glnro, line 50: Perhaps we pass `from` so we can reuse this elsewhere

glnro (approved):
lgtm, there's something up with the spacing formatting on this pr that's inconsistent with the rest of the codebase in certain places like variable declarations in files like `e2e_vesting_test.go`

tests/e2e/e2e_exec_test.go
Pantani: if we pass `from` we will overwrite the default from flag. `alice` is only the default one. eg: https://github.com/cosmos/gaia/pull/1738/files#diff-8f39b3a78fd9e2c7c36381bbd6f05c53940c1fb1c7f4babf538ed571cb756b5fR212-R214

Pantani:
> lgtm, there's something up with the spacing formatting on this pr that's inconsistent with the rest of the codebase in certain places like variable declarations in files like `e2e_vesting_test.go`

great, thanks; nice catch. Can you show me an example and the best format we can use in this case? I can easily change on this PR

PR #1739, opened 2022-09-13
fix non-determinism test in sim_test.go

okwme:
added label `sim` so that `make test-sim-nondeterminism` would be run to confirm it works now.

### DIFF
--- a/app/sim_test.go
+++ b/app/sim_test.go
@@ -107,7 +107,7 @@ func TestAppStateDeterminism(t *testing.T) {
 			}
 
 			db := dbm.NewMemDB()
-			app := gaia.NewGaiaApp(logger, db, nil, true, map[int64]bool{}, gaia.DefaultNodeHome, simapp.FlagPeriodValue, params.MakeTestEncodingConfig(), simapp.EmptyAppOptions{}, interBlockCacheOpt())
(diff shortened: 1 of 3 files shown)


yaruwangway (approved):

app/sim_test.go
```diff
@@ -107,7 +107,7 @@ func TestAppStateDeterminism(t *testing.T) {
 			}
 
 			db := dbm.NewMemDB()
-			app := gaia.NewGaiaApp(logger, db, nil, true, map[int64]bool{}, gaia.DefaultNodeHome, simapp.FlagPeriodValue, params.MakeTestEncodingConfig(), simapp.EmptyAppOptions{}, interBlockCacheOpt())
+			app := gaia.NewGaiaApp(logger, db, nil, true, map[int64]bool{}, gaia.DefaultNodeHome, simapp.FlagPeriodValue, gaia.MakeTestEncodingConfig(), EmptyAppOptions{}, interBlockCacheOpt())
```
mmulji-ic, line 110: @okwme basically this was the problem with the missing messages?

mmulji-ic (approved):

app/sim_test.go
okwme: yes we initiate the encoding config differently in the new version of cosmos sdk

PR #1749, opened 2022-09-19
refactor(e2e): small improvements to the e2e tests

Pantani:
## Description

Small improvements to the e2e tests:

- remove deprecated `ioutil` package usage;
- move all test exec functions to `e2e_exec_test.go`;
- move all test query functions to `e2e_query_test.go`;
- move all generic test functions to `query.go`;
- move all test `IBC` functions to `e2e_ibc_test.go`;
- Use `math.NewInt` instead `sdk.NewInt`;
- Avoid type the home folder everytime;
- Remove unused endpoint variables from most of the methods;
- Improve the import sort;
- Remove unused methods and global variables;
- Use `executeGaiaTxCommand` helper for all exec `gaiad` methods;


tests/e2e/e2e_query_test.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/ory/dockertest/v3/docker"
+)
+
+func (s *IntegrationTestSuite) queryGovProposal(endpoint string, proposalId uint64) (govv1beta1.QueryProposalResponse, error) {
+	var emptyProp govv1beta1.QueryProposalResponse
+
+	path := fmt.Sprintf("%s/cosmos/gov/v1beta1/proposals/%d", endpoint, proposalId)
+	resp, err := http.Get(path)
+	if err != nil {
+		s.T().Logf("This is the err: %s", err.Error())
```
Pantani: Should the test fail here?

tests/e2e/e2e_exec_test.go
```diff
@@ -5,69 +5,15 @@ import (
...
-		errBuf.String(),
-		"successfully opened init channel",
-		"failed to connect chains via IBC: %s", errBuf.String(),
-	)
-
-	s.T().Logf("connected %s and %s chains via IBC", s.chainA.id, s.chainB.id)
-}
-
-func (s *IntegrationTestSuite) sendMsgSend(c *chain, valIdx int, from, to, amt, fees string, expectErr bool) {
+func (s *IntegrationTestSuite) execSendMsgSend(c *chain, valIdx int, from, to, amt, fees string, expectErr bool) {
```
okwme: could we use `s.executeGaiaTxCommand` in here too for consistency?

tests/e2e/e2e_exec_test.go
```diff
@@ -405,110 +303,30 @@ func (s *IntegrationTestSuite) executeGaiaTxCommand(ctx context.Context, c *chai
...
-			return currentHeight > 0
-		},
-		5*time.Second,
-		time.Second,
-		"Get node status returned a non-zero code; stdout: %s, stderr: %s", outBuf.String(), errBuf.String(),
 	)
-	return currentHeight
 }
 
-func (s *IntegrationTestSuite) execCreateGroup(c *chain, valIdx int, endpoint string, adminAddr string, metadata string, groupMembersPath string, fees string) {
```
okwme: i think that gofumpt was adding explicit types to parameters instead of letting them be grouped like this.
we only started using gofumpt recently though and i'm unsure whether it's run as part of our CI or not at this point.
I guess we'd see an error on this PR if it had any problems with this.

all to say that I don't have a strong preference on either (a weak preference for the way you've done it here), but would defer to our CI should we need to.

tests/e2e/e2e_query_test.go
okwme: looks like the next line contains the empty proposal response and an error.
these could be combined or the first check could be removed.

tests/e2e/e2e_test.go
```diff
@@ -753,17 +751,19 @@ func (s *IntegrationTestSuite) TestStaking() {
 	validatorA := s.chainA.validators[0]
 	validatorB := s.chainA.validators[1]
 	validatorAAddr, err := validatorA.keyInfo.GetAddress()
-	s.NoError(err)
+	s.Require().NoError(err)
 	validatorBAddr, err := validatorB.keyInfo.GetAddress()
-	s.NoError(err)
+	s.Require().NoError(err)
```
okwme, line 658: whoops!  : )

okwme (approved):
nice work @Pantani !

okwme (changes requested):
Ah actually looks like the E2E is still failing on starting Hermes.
I wonder if the latest changes @yaruwangway made to get Hermes working are what's missing?

yaruwangway:
> Ah actually looks like the E2E is still failing on starting Hermes. I wonder if the latest changes @yaruwangway made to get Hermes working are what's missing?

this is because we point to the[ latest ](https://github.com/cosmos/gaia/blob/42ac0b2bd95b9fa890eef9eaaad017a83060eb4f/tests/e2e/e2e_setup_test.go#L482)hermes image in `ghcr.io/cosmos/hermes-e2e`, not a specific version, so the cli failed. I will push the fix here.

tests/e2e/e2e_exec_test.go
Pantani: I think so for this one and all of the other exec methods. I believe instead of using this effort now, so we use the `executeGaiaTxCommand` only in this method. We should create a library to exec methods into the container to be more generic and start to use them for all methods. But this needs a bigger refactor

Pantani: I made a small change so we can reuse `executeGaiaTxCommand` in all exec commands, WDYT?

tests/e2e/e2e_ibc_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+	exec, err := s.dkrPool.Client.CreateExec(docker.CreateExecOptions{
+		Context:      ctx,
+		AttachStdout: true,
+		AttachStderr: true,
+		Container:    s.hermesResource.Container.ID,
+		User:         "root",
+		Cmd: []string{
+			"hermes",
+			"create",
```
yaruwangway: this "hermes create channel" might induce "mismatch sequence" err due to the competing of this hermes instance with the instance of "hermes start", so it is better to use ""hermes tx chan-open-init" which is already in main branch.

yaruwangway: i see you also keep "hermes tx chan-open-init" , so the  "hermes create channel"  is not used ?

tests/e2e/e2e_ibc_test.go
```diff
@@ -0,0 +1,382 @@
...
+		fmt.Sprintf("--%s=%s", flags.FlagFees, fees),
+		fmt.Sprintf("--%s=%s", flags.FlagChainID, c.id),
+		"--keyring-backend=test",
+		"--broadcast-mode=sync",
+		"--output=json",
+		"-y",
+	}
+	s.T().Logf("sending %s from %s to %s (%s)", token, s.chainA.id, s.chainB.id, recipient)
+	s.executeGaiaTxCommand(ctx, c, ibcCmd, valIdx, s.defaultExecValidation(c, valIdx))
+	s.T().Log("successfully sent IBC tokens")
```
yaruwangway, line 142: no matter the ibc send successful or not , here will log "successfully sent IBC tokens" ?

okwme: no the `executeGaiaTxCommand` contains require statements that will skip the rest of the test if they fail

okwme (approved):

PR #1752, opened 2022-09-19
test: add daily e2e cron job

mmulji-ic:


mmulji-ic:
@faddat would you add anything else?

.github/workflows/nightly-e2e.yml
```diff
@@ -0,0 +1,14 @@
...
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+
+  run-tests:
+    uses: cosmos/gaia/.github/workflows/test.yml@main
+
+  run-simulations:
+    uses: cosmos/gaia/.github/workflows/sims.yml@main
```
glnro: Just needs a newline

glnro (commented):
Technically this is a nightly run of all test suites not just e2e...if the goal is just to isolate e2e then maybe we emulate something more specific a la [tendermint's nightly e2e run](https://github.com/tendermint/tendermint/blob/main/.github/workflows/e2e-nightly-main.yml#L13)

mmulji-ic:
> Technically this is a nightly run of all test suites not just e2e...if the goal is just to isolate e2e then maybe we emulate something more specific a la [tendermint's nightly e2e run](https://github.com/tendermint/tendermint/blob/main/.github/workflows/e2e-nightly-main.yml#L13)

I'm ok running the additional tests as we had non-determinism in the main branch that crept in. Will change the yml file name to reflect thats its not just e2e.

PR #1758, opened 2022-09-22
security: fix resource leaks from staticmajor + add cosmos/gosec & staticmajor

odeke-em:
This change fixes resource leaks identified by a static analyzer from Orijtech Inc called "staticmajor".

While here, also added cosmos/gosec scanning which will help with running security rules and take advantage of Github's vulnerability scanners on runs.

### DIFF
--- a/.github/workflows/gosec.yml
+++ b/.github/workflows/gosec.yml
@@ -1,0 +1,38 @@
+name: Run Gosec
+on:
+  pull_request:
+    paths:
+      - '**/*.go'
+      - 'go.mod'
+      - 'go.sum'
+  push:
+    branches:
+      - main
+    paths:
+      - '**/*.go'
+      - 'go.mod'
+      - 'go.sum'
+
+jobs:
+  Gosec:
+    permissions:
+      security-events: write
+
+    runs-on: ubuntu-latest
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Checkout Source
+        uses: actions/checkout@v3
+
+      - name: Run Gosec Security Scanner
+        uses: cosmos/gosec@master
+        with:
+          # Let the report trigger a failure with the Github Security scanner features.
+          args: "-no-fail -fmt sarif -out results.sarif ./..."
+
+      - name: Upload SARIF file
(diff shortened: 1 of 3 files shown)


mmulji-ic:
Thanks @odeke-em looks like we have issues we need to fix!

odeke-em:
Cheers @mmulji-ic! Don't worry though, the gosec issues can be progressively fixed and will just be filed in a section on GitHub security and eventually can look at them. A bunch of them might even be false positives. The merge of this PR shouldn't hold y'all back :-)

mmulji-ic (approved):

odeke-em:
Howdy my good folks, please help with a merge.. This code has lingered for more than a month :-)

.github/workflows/gosec.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Checkout Source
+        uses: actions/checkout@v3
+
+      - name: Run Gosec Security Scanner
+        uses: cosmos/gosec@master
+        with:
+          # Let the report trigger a failure with the Github Security scanner features.
+          args: "-no-fail -fmt sarif -out results.sarif ./..."
+
+      - name: Upload SARIF file
```
yaruwangway, line 34: Hi, this will add `results.sarif` to gaia repo ? is this a necessity ?

Pantani: I think we need the Sarif file to show the results on GitHub. But I'm not 100% sure

https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning

okwme:
priority is to confirm whether this new check has exposed anything that needs to be fixed before Rho is released.
If nothing needs to be changed, let's wait until after Rho as changes in Lambda may make many of these moot to begin with.
If there are changes needed, let's make those changes first and then merge this after.
@Pantani will run through the list before Rho release.

Pantani:
I think we can exclude some rules like `G107 (CWE-88)` and `G701 (CWE-)`

Pantani:
@odeke-em, can you please update your branch?

PR #1683, opened 2022-08-19
test: ica e2e test

yaruwangway:
# e2e test to test interchain-account: 

The followings are included in the pr:
- [x] add two accounts to genesis, one account is used as relayer wallet, one is used as ica owner. (please do not the validator as relayer wallet, this might introduce sequence mismatch error)
- [x] add allowed messages to ica host in genesis.
- [x] bump hermes to 1.0.0-rc2 
- [x] e2e test: 
       - register ica and query it 
       `gaiad tx icamauth register --connection-id connection-0 --from icaOnwer`
       - ica bank send tokens on chain-2
       `gaiad tx icamauth submit send.json --connection-id connection-0  --from icaOnwer`
       - ica delegates (optional)
        `gaiad tx icamauth submit delegate.json --connection-id connection-0  --from icaOnwer`
       - ica ibc transfer
       `gaiad tx icamauth submit ibc_transfer.json --connection-id connection-0  --from icaOnwer`

PR #1769, opened 2022-09-27
fix: Rectify e2e test deletion and e2e refactor

glnro:
Reintroduces `group` e2e deleted in https://github.com/cosmos/gaia/pull/1683
Fixes validation error in`gaiad keys add` introduced in https://github.com/cosmos/gaia/pull/1749


mmulji-ic (approved):
Good job spotting that this test was omitted. 
Think we can patch the SDK docs in another commit.

PR #1770, opened 2022-09-27
refactor: refactor group e2e test setup

glnro:
Leverages native sdk types for writing policies & proposals to file


tests/e2e/e2e_setup_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/cosmos/cosmos-sdk/x/group"
```
mmulji-ic: Should be moved down with the other sdk incl.

tests/e2e/e2e_groups_test.go
```diff
@@ -1,14 +1,14 @@
 package e2e
```
mmulji-ic, line 1: Missing some docs to explain the flows that are being tested

tests/e2e/e2e_groups_test.go
```diff
@@ -162,16 +159,18 @@ func (s *IntegrationTestSuite) prepareGroupFiles(c *chain, adminAddr string, mem
...
 		},
 	}
+
 	s.writeGroupMembers(c, members, originalMembersFilename)
 
-	newMembers := append(members, GroupMember{
+	newMembers := append(members, group.MemberRequest{
 		Address:  member3Address,
 		Weight:   "1",
 		Metadata: "Charlie",
```
mmulji-ic, line 188: Maybe use different meta data from charlieAddr below

tests/e2e/e2e_setup_test.go
yaruwangway: please sort the imports

tests/e2e/e2e_groups_test.go
```diff
@@ -162,16 +159,18 @@ func (s *IntegrationTestSuite) prepareGroupFiles(c *chain, adminAddr string, mem
...
 	})
+
 	s.writeGroupMembers(c, newMembers, addMemberFilename)
 
-	removeMembers := append(members, GroupMember{
+	removeMembers := append(members, group.MemberRequest{
```
yaruwangway: can you just directly change the `weight` , this `append` way remove members require the member to be removed the last member in the slice.

tests/e2e/e2e_groups_test.go
glnro: I'm not sure why we would have docs for a test, the acceptance criteria/tested flows were part of the original ticket for this. There's sufficient logging in this suite to understand what is being tested.

glnro: There's also a qa document in hub-eng that explains group flows in better detail

tests/e2e/e2e_groups_test.go
glnro: Metadata field in this case is irrelevant as this is test data, changing it has no effect on the test, and in this particular case, we're targeting the same member in both so it makes sense to keep it consistent. However within the scope of this function we should actually be using `member3Address` passed in and not the globally available `charlieAddr`

tests/e2e/e2e_groups_test.go
glnro: sure thing

tests/e2e/e2e_groups_test.go
mmulji-ic: An overview of whats being tested in comment form, without having a deep dive into the code/read through the logs helps. Remember this is an open source project, other contributors don't have access to hub-eng.

tests/e2e/e2e_groups_test.go
mmulji-ic: I know metadata is not relevant to the test outcome, but I still think its good to keep both test cases clean and not imply connections between, its visually distracting.

glnro: Right but in this case the metadata is denoting the specific user and is meant to be the same entity...the only thing that's supposed to change is the weight

tests/e2e/e2e_groups_test.go
glnro: Totally fair, let's open an issue then to make sure it's consistent across test suites

Pantani (approved):
LGMT

Pantani (approved):
LGMT

PR #1772, opened 2022-09-30
refactor: gov e2e test setup

glnro:
Clean up test setup for writing proposals to file and abstract tests into separate suite


tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
+package e2e
+
+import (
+	"fmt"
+	sdk "github.com/cosmos/cosmos-sdk/types"
```
yaruwangway, line 5: please order imports

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+/*
+GovSoftwareUpgrade tests passing a gov proposal to upgrade the chain at a given height.
+Test Benchmarks:
+1. Submission, deposit and vote of message based proposal to upgrade the chain at a height (current height + buffer)
+2. Validation that chain halted at upgrade height
+3. Teardown & restart chains
+4. Reset proposalCounter so subsequent tests have the correct last effective proposal id for chainA
+TODO: Perform upgrade in place of chain restart
+*/
+func (s *IntegrationTestSuite) GovSoftwareUpgrade() {
```
yaruwangway, line 72: can we ensure `GovSoftwareUpgrade` runs at last since it teardown the network and restart?

yaruwangway (approved):

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
+package e2e
+
+import (
+	"fmt"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"time"
+
+	govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
```
Pantani, line 8:
```suggestion
	"fmt"
	"time"

	sdk "github.com/cosmos/cosmos-sdk/types"
	govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
```

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+5. Validation that funds have been deposited to recipient account
+*/
+func (s *IntegrationTestSuite) SendTokensFromNewGovAccount() {
+	s.writeGovProposals(s.chainA)
+	chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
+	senderAddress, err := s.chainA.validators[0].keyInfo.GetAddress()
+	s.Require().NoError(err)
+	sender := senderAddress.String()
+	// Gov tests may be run in arbitrary order, each test must increment proposalCounter to have the correct proposal id to submit and query
+	proposalCounter++
```
Pantani, line 27: Can we save this inside the `IntegrationTestSuite` object so we can avoid global variables?

Pantani: also, if we move the test to run in parallel will create a race condition

tests/e2e/e2e_gov_test.go
Pantani: Can we only re-run the validators (`s.runValidators(s.chainA, 0)`) and check if they are running querying the status or the next block after the migration? Because the `s.SetupSuite()` will recreate the entire chain before rerunning it.

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+/*
+GovCancelSoftwareUpgrade tests passing a gov proposal that cancels a pending upgrade.
+Test Benchmarks:
+1. Submission, deposit and vote of message based proposal to upgrade the chain at a height (current height + buffer)
+2. Submission, deposit and vote of message based proposal to cancel the pending upgrade
+3. Validation that the chain produced blocks past the intended upgrade height
+*/
+func (s *IntegrationTestSuite) GovCancelSoftwareUpgrade() {
+	s.T().Skip()
```
Pantani, line 122: Why skip this test?

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+SendTokensFromNewGovAccount tests passing a gov proposal that sends tokens on behalf of the gov module to a recipient.
+Test Benchmarks:
+1. Subtest to fund the community pool via the distribution module
+2. Submission, deposit and vote of legacy proposal to fund the gov account from the community pool
+3. Validation that funds have been deposited to gov account
+4. Submission, deposit and vote of message based proposal to send funds from the gov account to a recipient account
+5. Validation that funds have been deposited to recipient account
+*/
+func (s *IntegrationTestSuite) SendTokensFromNewGovAccount() {
+	s.writeGovProposals(s.chainA)
```
Pantani, line 21: can we use a subtest?

```suggestion
	s.Run("description.....", func() {
		s.writeGovProposals(s.chainA)
		...
```

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+GovSoftwareUpgrade tests passing a gov proposal to upgrade the chain at a given height.
+Test Benchmarks:
+1. Submission, deposit and vote of message based proposal to upgrade the chain at a height (current height + buffer)
+2. Validation that chain halted at upgrade height
+3. Teardown & restart chains
+4. Reset proposalCounter so subsequent tests have the correct last effective proposal id for chainA
+TODO: Perform upgrade in place of chain restart
+*/
+func (s *IntegrationTestSuite) GovSoftwareUpgrade() {
+	chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
```
Pantani, line 73: can we use a subtest?

```suggestion
	s.Run("description.....", func() {
		chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
		...
```

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+				return false
+			}
+			if currentHeight == upgradeHeight {
+				counter++
+			}
+			return counter >= 2
+		},
+		8*time.Second,
+		2*time.Second,
+	)
```
Pantani, line 211: If the chain started and the upgrade was successful before calling this method, the `currentHeight == upgradeHeight` will always be false. I think we can assume that if the chain no longer reaches the upgrade height in 30 seconds, the chain is halted. WDYT?

```suggestion
	s.Require().Eventually(
		func() bool {
			currentHeight := s.getLatestBlockHeight(c, valIdx)

			return upgradeHeight > currentHeight
		},
		30*time.Second,
		5*time.Second,
	)
```

tests/e2e/e2e_gov_test.go
```diff
@@ -0,0 +1,296 @@
...
+				return proposal.GetProposal().Status == govv1beta1.StatusDepositPeriod
+			},
+			15*time.Second,
+			5*time.Second,
+		)
+	})
+}
+
+func (s *IntegrationTestSuite) depositGovProposal(chainAAPIEndpoint, sender string, fees string, proposalId int) {
+	s.Run("deposit_gov_proposal", func() {
```
Pantani, line 261: IMHO we can use the subtests for an entire test context, like a test in the gov was succeeded. And only print logs in these helper exec methods. WDYT?

PR #1795, opened 2022-10-10
feat: add globalfee initModule method for upgrade handler

yaruwangway:
# Add globalfee.InitModule

Add globalfee initModule for v8 upgrade Handler to use to init globalfee Params.

### DIFF
--- a/x/globalfee/module.go
+++ b/x/globalfee/module.go
@@ -87,7 +87,7 @@ func NewAppModule(paramSpace paramstypes.Subspace) *AppModule {
 }
 
 // InitModule will initialize the globalfee module.
-//It should only be called once and as an alternative to InitGenesis.
+// It should only be called once and as an alternative to InitGenesis.
 func (a AppModule) InitModule(ctx sdk.Context, globalfeeParams types.Params) {
 	a.paramSpace.SetParamSet(ctx, &globalfeeParams)
 }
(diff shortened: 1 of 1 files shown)


x/globalfee/module.go
```diff
@@ -86,6 +86,12 @@ func NewAppModule(paramSpace paramstypes.Subspace) *AppModule {
 	return &AppModule{paramSpace: paramSpace}
 }
 
+// InitModule will initialize the globalfee module.
+//It should only be called once and as an alternative to InitGenesis.
```
glnro:
```suggestion
// It should only be called once and as an alternative to InitGenesis.
```

glnro (approved):

x/globalfee/module.go
```diff
@@ -86,6 +86,12 @@ func NewAppModule(paramSpace paramstypes.Subspace) *AppModule {
 	return &AppModule{paramSpace: paramSpace}
 }
 
+// InitModule will initialize the globalfee module.
+// It should only be called once and as an alternative to InitGenesis.
+func (a AppModule) InitModule(ctx sdk.Context, globalfeeParams types.Params) {
+	a.paramSpace.SetParamSet(ctx, &globalfeeParams)
```
glnro, line 92: Do we want to do a nil check for `globalFeeParams` ?

yaruwangway: good idea !

yaruwangway: i think this cannot be nil... `globalfeeParams` is a struct.

PR #1800, opened 2022-10-11
fix(export-genesis): fix export genesis command for missing values

### LINKED ISSUE #1384: Exported genesis can't be validated using gaiad validate-genesis
## Summary of Bug

If you use `gaiad validate-genesis` to validate an exported genesis from the Cosmos Hub mainnet, you get the following error:

```
Error: error validating genesis file mainnet-genesis.json: name field cannot be blank
```

This was reproduced by @ASoTNetworks

## Version

I've reproduced this with gaiad 6.00 and 6.0.4

## Steps to Reproduce

1. Stop a synced node
2. Export a genesis file  `gaiad --for-zero-height --height=9961097 &> mainnet-genesis.json` (Notice I need to use `&>` and not `>` to dump contents into a file, because gaiad is writing output to stderr and not stdout. This is an unrelated issue)
3. `gaiad validate-genesis mainnet-genesis.json`
____

#### For Admin Use

- [ ] Not duplicate issue

Pantani:
part of #1384

## Description

Some values are missing in the genesis JSON when we are exporting it. So we can't validate the genesis after export.

***Most of the values should be upgraded based on default genesis parameters, but maybe we should also add the `liquidityPoolRecords` parameter from the liquidity module***


glnro (approved):

mmulji-ic:
@Pantani @glnro this is not tested via the current tests that we have right? If not, is it practical to add something?

app/upgrades/v8/upgrades.go
```diff
@@ -16,6 +18,15 @@ func CreateUpgradeHandler(
 	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		ctx.Logger().Info("start to run module migrations...")
 
+		// Add atom name and symbol into the bank keeper
+		atomMetaData, found := keepers.BankKeeper.GetDenomMetaData(ctx, "uatom")
+		if !found {
+			return nil, errors.New("atom denom not found")
+		}
+		atomMetaData.Name = "Cosmos Hub Atom"
+		atomMetaData.Symbol = "ATOM"
```
yaruwangway, line 27: Hi, when i query the node in public net, denom-metadata is already there.
```
{"metadatas":[{"description":"The native staking token of the Cosmos Hub.","denom_units":[{"denom":"uatom","exponent":0,"aliases":["microatom"]},{"denom":"matom","exponent":3,"aliases":["milliatom"]},{"denom":"atom","exponent":6,"aliases":[]}],"base":"uatom","display":"atom","name":"","symbol":""}],"pagination":{"next_key":null,"total":"1"}}
```

also, the denom is uatom.

glnro: I think this is different than the actual denom

glnro: https://github.com/cosmos/cosmos-sdk/blob/55054282d2df794d9a5fe2599ea25473379ebc3d/x/bank/types/metadata_test.go

Pantani: We can confirm the parameter through mainnet API (`http://<rpc...>:1317/cosmos/bank/v1beta1/denoms_metadata`) . The parameter the SDK use as a key index is `base`, and we are using `uatom`:


```shell
$ curl http://<RPC_ADDRESS>:1317/cosmos/bank/v1beta1/denoms_metadata
{
  "metadatas": [
    {
      "description": "The native staking token of the Cosmos Hub.",
      "denom_units": [
        {
          "denom": "uatom",
          "exponent": 0,
          "aliases": [
            "microatom"
          ]
        },
        {
          "denom": "matom",
          "exponent": 3,
          "aliases": [
            "milliatom"
          ]
        },
        {
          "denom": "atom",
          "exponent": 6,
          "aliases": [
          ]
        }
      ],
      "base": "uatom",
      "display": "atom",
      "name": "",
      "symbol": ""
    }
  ],
  "pagination": {
    "next_key": null,
    "total": "1"
  }
}
```

Pantani: https://github.com/cosmos/cosmos-sdk/blob/a975f67b245c214fa11408c4eb361147b4679d62/x/bank/keeper/keeper.go#L309-L309

okwme:
It's strange that Name and Symbol were left empty when the rest was populated.  
We should ensure whatever it is it matches https://github.com/cosmos/chain-registry/blob/master/cosmoshub/assetlist.json

They have us listed as 
```
        "name": "Cosmos",
        "symbol": "ATOM",
```
Either we update the chain registry or update here so that name = `Cosmos` instead of `Cosmos Hub Atom`.  
I prefer the former since there may be wallets, exchanges or explorers using the chain-registry already.

app/upgrades/v8/upgrades.go
```diff
@@ -16,6 +18,15 @@ func CreateUpgradeHandler(
 	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		ctx.Logger().Info("start to run module migrations...")
 
+		// Add atom name and symbol into the bank keeper
+		atomMetaData, found := keepers.BankKeeper.GetDenomMetaData(ctx, "uatom")
+		if !found {
+			return nil, errors.New("atom denom not found")
+		}
+		atomMetaData.Name = "Cosmos Hub Atom"
```
okwme, line 26:
```suggestion
		atomMetaData.Name = "Cosmos"
```

okwme (changes requested):
change "Name" to "Cosmos" or change registry

app/upgrades/v8/upgrades.go
glnro: I feel like we should change the registry

glnro: @okwme https://github.com/cosmos/chain-registry/pull/848 is pending, if we merge that then we could merge this as is

app/upgrades/v8/upgrades.go
```diff
@@ -16,6 +18,15 @@ func CreateUpgradeHandler(
 	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		ctx.Logger().Info("start to run module migrations...")
 
+		// Add atom name and symbol into the bank keeper
+		atomMetaData, found := keepers.BankKeeper.GetDenomMetaData(ctx, "uatom")
+		if !found {
+			return nil, errors.New("atom denom not found")
+		}
+		atomMetaData.Name = "Cosmos"
```
Pantani:
```suggestion
		atomMetaData.Name = "Cosmos Hub Atom"
```

app/upgrades/v8/upgrades.go
okwme: :+1:

okwme (approved):
tests are timing out but I checked them locally and they pass e2e : )

PR #1634, opened 2022-08-01
Update cosmos/ledger-go to release version

gjermundgaraba:
closes #1576

Issue #1576, opened 2022-07-06
update cosmos/ledger-go after #4 merged


okwme:
## Summary of Bug

after temp fix in #1575 we should make permanent fix once https://github.com/cosmos/ledger-go/pull/4 is merged

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

gjermundgaraba:
[#4](https://github.com/cosmos/ledger-go/pull/4) was just merged. Do we need to wait for it to cut a release before we could make a PR for this (I would be happy to update this, test it and make a PR)?

okwme:
great to hear!
i've just asked in the PR whether there will be a release or not, their last one wasn't since 2019.
would be amazing if you were able to update, test and make a new PR based on their reply @bjaanes thank you so much for the contribution!

gjermundgaraba:
I just saw 0.9.3 was cut, so I created a PR to update that. Let me know if want any changes.

PR #1801, opened 2022-10-11
use master branch of cosmos/ledger-cosmos-go

faddat:
While working on Akash today I noticed that there were some **very 
strange** things going on with our ledger imports.  The tl;dr is:

* we should stop maintaining cosmos/ledger-go
* we should ensure that we routinely cut releases of 
cosmos/ledger-go-cosmos

Here's a PR that gets Gaia on the right track, reflecting those changes.


okwme:
Thanks @faddat .
Zondax maintains both of those repositories.
It was relatively recently that we updated to their latest release (https://github.com/cosmos/gaia/issues/1576, https://github.com/cosmos/gaia/pull/1634).

We need to confirm whether there are API breaking changes that would impact users of gaia to ensure they're included in communications. Have you investigated what are the breaking changes between the updates you've included?
If not, maybe someone from Zondax can help us review this PR—maybe @philipglazman or @jleni?

jleni:
Hi @faddat @okwme 
As Billy explained, we have recently taken over the maintenance in this org and we are moving to a more predictable release schedule.

Could you provide some more info about these "strange things" you describe? Or it is related only to go.mod issues and a release would clear up the problem?

jleni:
I have just created this issue https://github.com/cosmos/ledger-cosmos-go/issues/30 to track the repo reorganization that will take place soon

go.mod
```diff
@@ -256,6 +255,7 @@ require (
 	github.com/yagipy/maintidx v1.0.0 // indirect
 	github.com/yeya24/promlinter v0.2.0 // indirect
 	github.com/zondax/hid v0.9.1-0.20220302062450-5552068d2266 // indirect
+	github.com/zondax/ledger-go v0.12.2 // indirect
```
Pantani, line 258: The `zondax/ledger-go` seems not well maintained. I think we should avoid importing this package into Gaia

Pantani (changes requested):

okwme:
thanks for the link to track progress @jleni !
We'll close this PR until the zondax plan for maintained libraries is complete and reference them directly then

PR #1802, opened 2022-10-11
fix: remove globalfee `initModule`

yaruwangway:
# remove globalfee `initModule`
- [x] move `initModule` of global fee, for upgrade, one can use either add globalfee to store in `v8/upgrades/constans.go` or delete `globalfee` from `vm` in  `CreateUpgradeHandler` and let `mm.RunMigrations` to call `initgenesis` for globalfee.


app/upgrades/v7/upgrades.go
```diff
@@ -57,8 +57,8 @@ func CreateUpgradeHandler(
 		ctx.Logger().Info("start to init interchainaccount module...")
 
 		// initialize ICS27 module
-		icaModule, correctTypecast := mm.Modules[icatypes.ModuleName].(ica.AppModule)
-		if !correctTypecast {
+		icaModule, err := mm.Modules[icatypes.ModuleName].(ica.AppModule)
```
Pantani: This is not an error, it is a [Golang type assertion](https://go.dev/tour/methods/15). The `icaModule` var will be the casted value, and the `err` var will be true if the cast works well and false if not. If we assert this way, the code can panic later because the struct will be empty

yaruwangway: yeah, right! thank you!

glnro (approved):

Pantani (approved):

PR #1809, opened 2022-10-12
chore: rho upgrade handler

glnro:


app/upgrades/v8/upgrades.go
```diff
@@ -16,9 +18,7 @@ func CreateUpgradeHandler(
 	keepers *keepers.AppKeepers,
 ) upgradetypes.UpgradeHandler {
 	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
-		ctx.Logger().Info("start to run module migrations...")
-
-		// Add atom name and symbol into the bank keeper
+			// Add atom name and symbol into the bank keeper
```
Pantani: the indentation seems wrong here

Pantani (approved):
LGMT

app/upgrades/v8/upgrades.go
okwme:
```suggestion
		// Add atom name and symbol into the bank keeper
```

okwme:
had to run e2e locally but they pass

app/upgrades/v8/upgrades.go
```diff
@@ -16,8 +18,6 @@ func CreateUpgradeHandler(
 	keepers *keepers.AppKeepers,
 ) upgradetypes.UpgradeHandler {
 	return func(ctx sdk.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
-		ctx.Logger().Info("start to run module migrations...")
```
okwme, line 19: why do we want to remove this? during a previous upgrade validators complained about the lack of feedback in the terminal when performing an upgrade. it took a really long time and there was no info about what was happening. @yaruwangway added this so that users would know that the upgrade was starting this step in order to track progress.

okwme: ah i see it was just moved lower : )

okwme (approved):

PR #1814, opened 2022-10-13
v5

okwme:


app/keepers/keepers.go
```diff
@@ -361,13 +361,14 @@ func NewAppKeeper(
...
 		appKeepers.TransferKeeper,
-		// appKeepers.IBCKeeper.ChannelKeeper,
+		appKeepers.IBCKeeper.ChannelKeeper,
 		appKeepers.DistrKeeper,
 	)
 
-	appKeepers.RouterModule = router.NewAppModule(appKeepers.RouterKeeper, transferIBCModule) // 0,
-	// routerkeeper.DefaultForwardTransferPacketTimeoutTimestamp,
-	// routerkeeper.DefaultRefundTransferPacketTimeoutTimestamp,
+	appKeepers.RouterModule = router.NewAppModule(appKeepers.RouterKeeper, transferIBCModule, 0,
```
yaruwangway: 0 is for `retriesOnTimeout`, it should not be 0 ?

okwme: i guess they put in the work to make it possible to retry we should at least use the feature

okwme:
```suggestion
	appKeepers.RouterModule = router.NewAppModule(appKeepers.RouterKeeper, transferIBCModule, 1,
```

tests/e2e/e2e_encode_test.go
```diff
@@ -12,6 +12,7 @@ const (
 )
 
 func (s *IntegrationTestSuite) TestEncode() {
+	s.T().Skip()
```
glnro: Is there a reason why we need to skip:?

tests/e2e/e2e_encode_test.go
```diff
@@ -22,6 +23,7 @@ func (s *IntegrationTestSuite) TestEncode() {
 }
 
 func (s *IntegrationTestSuite) TestDecode() {
+	s.T().Skip()
```
glnro: Same here

okwme: good call, just removed them

tests/e2e/e2e_encode_test.go
okwme: removed : )

tests/e2e/e2e_ibc_test.go
```diff
@@ -276,6 +293,192 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
...
+2. Check balance of Account 2 on Chain 1
+3. Account 1 on Chain 1 sends x tokens to Account 2 on Chain 1 via Account 1 on Chain 2
+4. Check Balance of Account 1 on Chain 1, confirm it is original minus x tokens
+5. Check Balance of Account 2 on Chain 1, confirm it is original plus x tokens
+
+*/
+func (s *IntegrationTestSuite) TestMultihopIBCTokenTransfer() {
+	time.Sleep(30 * time.Second)
+
+	// var ibcStakeDenom string
```
glnro: Looks like a bunch of commented out variables...not sure if these need to be commented in later for something or they're left over

okwme: good catch, used them in previous versions of the test, removed now.

glnro (approved):

tests/e2e/e2e_ibc_test.go
```diff
@@ -237,9 +237,10 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
 	s.Run("send_uatom_to_chainB", func() {
 		// require the recipient account receives the IBC tokens (IBC packets ACKd)
 		var (
-			ibcStakeDenom string
 			balances      sdk.Coins
 			err           error
+			beforeBalance int64
```
yaruwangway, line 242: beforeBalance math.Int type ?

tests/e2e/e2e_ibc_test.go
```diff
@@ -276,6 +293,187 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
...
+		chainBAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainB.id][0].GetHostPort("1317/tcp"))
+
+		var (
+			beforeSenderUAtomBalance sdk.Coin
+			beforeMiddleIBCBalance   sdk.Coin
+		)
+
+		s.Require().Eventually(
+			func() bool {
+				beforeSenderUAtomBalance, err = getSpecificBalance(chainAAPIEndpoint, sender, "uatom")
```
yaruwangway: "uatom" -> `uatomDenom`

tests/e2e/e2e_ibc_test.go
```diff
@@ -276,6 +293,187 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
...
+
+				returned := beforeSenderUAtomBalance.Sub(tokenAmount).Sub(standardFees).IsEqual(afterSenderUAtomBalance)
+
+				return returned
+			},
+			1*time.Minute,
+			1*time.Second,
+		)
+
+		// since the forward receiving account is invalid, it should be refunded to the original sender (minus the original fee)
```
yaruwangway, line 457: this is a failure test ? 
is there a successful one  ?

okwme: yes this is the failure test (`send_failed_multihop_uatom_to_chainA_from_chainA`) the successful test is above and called `send_successful_multihop_uatom_to_chainA_from_chainA`

tests/e2e/e2e_ibc_test.go
okwme: because it is not clear what the IBC denom will be so only the int64 value is needed to confirm the difference

tests/e2e/e2e_ibc_test.go
okwme: updated

tests/e2e/e2e_ibc_test.go
```diff
@@ -250,11 +251,27 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
...
+				s.Require().NoError(err)
+				return balances.Len() != 0
+			},
+			time.Minute,
+			5*time.Second,
+		)
+		for _, c := range balances {
+			if strings.Contains(c.Denom, "ibc/") {
+				beforeBalance = c.Amount.Int64()
+				break
```
yaruwangway, line 268: so if we use the original c.Amount type, which is `sdkmath.Int`, maybe we do not need to convert to int64

yaruwangway (approved):

faddat:
This looks good thank you

tests/e2e/e2e_ibc_test.go
Pantani: IMHO is always better to use `sdk.Dec` to manage balances values so we can't lose the precision when we do math operations

tests/e2e/e2e_ibc_test.go
```diff
@@ -276,6 +293,187 @@ func (s *IntegrationTestSuite) TestIBCTokenTransfer() {
...
+5. Check Balance of Account 1 on Chain 2, confirm it is original plus x tokens
+*/
+func (s *IntegrationTestSuite) TestFailedMultihopIBCTokenTransfer() {
+	time.Sleep(30 * time.Second)
+
+	s.Run("send_failed_multihop_uatom_to_chainA_from_chainA", func() {
+		// require the recipient account receives the IBC tokens (IBC packets ACKd)
+		var (
+			err error
+		)
```
Pantani:
```suggestion
```

Pantani (commented):
awesome!! Nice improvements

okwme:
let's include the int changes in a refactor @Pantani @yaruwangway .
i copied it from somewhere else in the tests so if it's fixed in this PR it should be fixed elsewhere and at that point would be better suited as a general refactor.

PR #1819, opened 2022-10-14
chore(docs): add code guidelines

Pantani:
## Description

Add code guidelines so we can review and create this document together.


docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+make test-e2e
+```
+
+# Guidelines
+
+These guidelines are the conventions that govern our code. These conventions cover far more than just source file formatting. Can `gofmt` and `goimports` handle that for us.
+
+The goal of this guide is to manage this complexity by describing in detail the Dos and Don'ts of writing Go code. These rules keep the code base manageable while allowing engineers to use Go language features productively.
+
+Try to avoid extensive methods and always test your code. All PRs should have unless 95% of code coverage.
```
okwme:
```suggestion
Try to avoid extensive methods and always test your code. All PRs should have at least 95% of code coverage.
```

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+- Run the unit tests with the race condition flag on.
+
+```shell
+make test-race
+```
+
+- Run end-to-end integration tests (Docker needed).
+
+```shell
+make docker-build-hermes && \
```
okwme: isn't hermes docker built as part of `make docker-build-debug` ?

okwme (commented):
One question about the test command already including the hermes build but otherwise so great to see a doc like this!

There are some redundancies with the current https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md. This will be included in the hub.cosmos.network as it's part of the doc directory, should it link to the other file or the other way around? should we consider merging the content or use this one for website rendered audiences versus github markdown audiences?

docs/code-guidelines.md
yaruwangway: it is not part of `make-docker-build-debug`. and we are not using this hermes image for e2e test actually? e2e is using a hermes docker image [here](https://github.com/cosmos/gaia/blob/0b2781064be85a96c113776ba119df5084bef8c0/tests/e2e/e2e_ibc_test.go#L48).

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+    - /generate_release_note
+    - /scripts
+    - /testnets
+    - /githooks
+
+- /pkg: Library code that's to be reusable.
+    - /address
+    - /genesis
+
+- /internal: Private application and library code.
```
yaruwangway: do you mean we structure like `/internal/app`, `/internal/app` ?

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+    - /ante
+    - /app
+
+- /tools: Supporting tools for this project.
+    - /proto
+
+- /x: Cosmos Modules
```
yaruwangway: also
- /docs:

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+
+```go
+t := []string{}
+```
+
+The former declares a nil slice value, while the latter is non-nil but zero-length. They are functionally equivalent—their `len` and `cap` are both zero—but the nil slice is the preferred style.
+
+Note that there are limited circumstances where a non-nil but the zero-length slice is preferred, such as when encoding JSON objects (a `nil` slice encodes to `null`, while `[]string{}` encodes to the JSON array `[]`).
+
+When designing interfaces, avoid distinguishing between a nil slice and a non-nil, zero-length slice, as this can lead to subtle programming errors.
```
yaruwangway: and also distinguish if a map key exist from if its value is zero/nil/false.

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+	// normal code
+}
+```
+
+Instead, write:
+
+```go
+if err != nil {
+	// error handling
+	return // or continue, etc.
```
yaruwangway: why this is bad ? I feel this sometimes makes the code look cleaner

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+All references to names in your package will be done using the package name so that you can omit that name from the identifiers. For example, if you are in package chubby, you don't need to type ChubbyFile, which clients will write as `chubby.ChubbyFile`. Instead, name the type `File`, which clients will write as `chubby.File`. Avoid meaningless package names like util, common, misc, API, types, and interfaces. See https://go.dev/doc/effective_go#package-names and https://go.dev/blog/package-names for more.
+
+When naming packages, choose a name that is:
+
+- All lowercase. No capitals or underscores.
+- Does not need to be renamed using named imports at most call sites.
+- Short and succinct. Remember that the name is identified in full at every call site.
+- Not plural. For example, `net/url`, not `net/urls`.
+- Not "common", "util", "shared", or "lib". These are bad, uninformative names.
+
```
yaruwangway: to distinguish sdk and gaia with same package name, add `sdk` or `gaia` or `modulename` as prefix.
for example: `sdk/types`, `gaia/types` and `gaia/x/globalfee/types`, can use `sdktype`, `gaiatype`, `globalfeetype`

yaruwangway:
can we also have a guide for writing testing code, including test file naming, structure etc

docs/code-guidelines.md
Pantani: yes!!! WDYT?

docs/code-guidelines.md
Pantani: Agree!!!

docs/code-guidelines.md
Pantani: nice!!! I will add it

docs/code-guidelines.md
Pantani: It's not so bad. But the code keeps cleaner if we try to keep it more intended to the left instead to the right. for simple cases, it's good, but the code will be too nested for more complicated conditions

Pantani:
> One question about the test command already including the hermes build but otherwise so great to see a doc like this!
> 
> There are some redundancies with the current https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md. This will be included in the hub.cosmos.network as it's part of the doc directory, should it link to the other file or the other way around? should we consider merging the content or use this one for website rendered audiences versus github markdown audiences?

Yes, sure!! Let's merge them. It's nice to have clear documentation in one place and not massive documentation for everyone to read quickly; WDYT?

Pantani:
- [x] add semantic PR title guidelines
- [x] add branch name guidelines
- [x] add gitflow guidelines
- [x] add commit description guideline

Pantani:
- Avoid writing directly in the stdout. Eg:

https://github.com/cosmos/gaia/blob/fb95f508648fdd64b8a5ff2593be9e5720de032b/app/sim_test.go#L111-L111

Pantani:
- avoid panic
- handle error on the top level

Pantani:
- use subtests

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+```go
+var (
+  // The following two errors are exported
+  // so that users of this package can match them
+  // with errors.Is.
+
+  ErrBrokenLink = errors.New("link is broken")
+  ErrCouldNotOpen = errors.New("could not open")
+
```
yaruwangway: shall we create  a `types/errors/errors.go` to collect all err types together or put errors in each file where they were used?

docs/code-guidelines.md
```diff
@@ -0,0 +1,1141 @@
...
+  v.Send()
+}
+```
+
+## Performance
+
+Performance-specific guidelines apply only to the hot path.
+
+#### Prefer strconv over fmt
+
```
yaruwangway: can we add 
#### avoid "+" for frequent string concatenation
- bad
```
func strConcat(n int, str string) string {
	s := ""
	for i := 0; i < n; i++ {
		s += str
	}

	return s
}
```

- good
a better option is  `strings.Builder` or `[]byte`
```
func strConcat(n int, str string) string {
	b := []byte{}
	for i := 0; i < n; i++ {
		buf = append(buf, str...)
	}

	return string(b)
}
```

In golang, string is immutable. when use "+", every time, there has to be a new memory allocation. For `[]byte`, the  memory of the slice []byte is allocated in multiples.

yaruwangway:
can we add a section about importable and unimportable struct field and funcs ?

docs/code-guidelines.md
Pantani: Yes, this is a very nice one! sometimes the `+` can be more performative. But definitely is not so clear to read

okwme:
- [x] extract relevant parts of contibuting.md and include in this pr and remove from the other

okwme:
- [ ] add instructions from upgrade test here to run manually
They're here: https://github.com/cosmos/gaia/pull/1892#issuecomment-1317266413

docs/code-guidelines.md
Pantani: I think we can follow the current pattern and keep it inside the types package for each module, but we can create a `pkg/errors` for more generic errors.

Pantani:
> * [ ]  add instructions from upgrade test here to run manually

can you share this info for I can add?

okwme (approved):

PR #1824, opened 2022-10-14
Fix: use `bondDenom` as default globalfee denom

### LINKED ISSUE #1823: test issues of hard coded global fee denom
## Summary of Bug
The default global fee is using a hardcoded denom `uatom`.
This works only for the hub, but for users who want to use gaiad for tests, tests will fail if the chain is not inited with a denom `uatom`.

## Steps to Reproduce
- init a chain by`gaiad init` (the default chain is using denom `stake`)
- submit transactions with paying fee in `stake`, test will fail.

## Fix

config `bondDenom` in staking params as default global fee denom if globalfee is empty.

____

#### For Admin Use

- [ ] Not duplicate issue

yaruwangway:
## Fix: use `bondDenom` in staking module of the denom of defalut global fee.
closes: #1823

To fix issue #1823, stakingParams will be added to `FeeDecorator` for antehandler to access bondDnom. Default globalfee is configured the same as the chain's bonddenom, rather than hardcoded `uatom`

### DIFF
--- a/ante/ante.go
+++ b/ante/ante.go
@@ -18,6 +18,7 @@ type HandlerOptions struct {
 	IBCkeeper            *ibckeeper.Keeper
 	BypassMinFeeMsgTypes []string
 	GlobalFeeSubspace    paramtypes.Subspace
+	StakingSubspace      paramtypes.Subspace
 }
 
 func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
@@ -36,6 +37,9 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 	if opts.GlobalFeeSubspace.Name() == "" {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
 	}
+	if opts.StakingSubspace.Name() == "" {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
+	}
 
 	sigGasConsumer := opts.SigGasConsumer
 	if sigGasConsumer == nil {
@@ -49,7 +53,7 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 		ante.NewTxTimeoutHeightDecorator(),
 		ante.NewValidateMemoDecorator(opts.AccountKeeper),
 		ante.NewConsumeGasForTxSizeDecorator(opts.AccountKeeper),
-		gaiafeeante.NewFeeDecorator(opts.BypassMinFeeMsgTypes, opts.GlobalFeeSubspace),
+		gaiafeeante.NewFeeDecorator(opts.BypassMinFeeMsgTypes, opts.GlobalFeeSubspace, opts.StakingSubspace),
 		// if opts.TxFeeCheck is nil,  it is the default fee check
 		ante.NewDeductFeeDecorator(opts.AccountKeeper, opts.BankKeeper, opts.FeegrantKeeper, opts.TxFeeChecker),
 		ante.NewSetPubKeyDecorator(opts.AccountKeeper), // SetPubKeyDecorator must be called before all signature verification decorators
--- a/ante/ante.go
+++ b/ante/ante.go
@@ -35,10 +35,10 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "IBC keeper is required for AnteHandler")
 	}
 	if opts.GlobalFeeSubspace.Name() == "" {
-		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "globalfee param store is required for AnteHandler")
+		return nil, sdkerrors.Wrap(sdkerrors.ErrNotFound, "globalfee param store is required for AnteHandler")
 	}
 	if opts.StakingSubspace.Name() == "" {
-		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "staking param store is required for AnteHandler")
+		return nil, sdkerrors.Wrap(sdkerrors.ErrNotFound, "staking param store is required for AnteHandler")
 	}
 
 	sigGasConsumer := opts.SigGasConsumer
--- a/ante/ante.go
+++ b/ante/ante.go
@@ -29,16 +29,16 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "bank keeper is required for AnteHandler")
 	}
 	if opts.SignModeHandler == nil {
-		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "sign mode handler is required for ante builder")
+		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "sign mode handler is required for AnteHandler")
 	}
 	if opts.IBCkeeper == nil {
-		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "IBC keeper is required for middlewares")
+		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "IBC keeper is required for AnteHandler")
 	}
 	if opts.GlobalFeeSubspace.Name() == "" {
-		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
+		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "globalfee param store is required for AnteHandler")
 	}
 	if opts.StakingSubspace.Name() == "" {
-		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
+		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "staking param store is required for AnteHandler")
 	}
 
 	sigGasConsumer := opts.SigGasConsumer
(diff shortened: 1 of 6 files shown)


ante/ante.go
```diff
@@ -36,6 +37,9 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 	if opts.GlobalFeeSubspace.Name() == "" {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
 	}
+	if opts.StakingSubspace.Name() == "" {
+		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
```
Pantani:
```suggestion
		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "staking param store is required for ante builder")
```

ante/ante.go
```diff
@@ -36,6 +37,9 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
 	if opts.GlobalFeeSubspace.Name() == "" {
 		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "param store is required for ante builder")
```
Pantani:
```suggestion
		return nil, sdkerrors.Wrap(sdkerrors.ErrLogic, "global fee param store is required for ante builder")
```

x/globalfee/ante/antetest/fee_test_setup.go
```diff
@@ -54,6 +55,12 @@ func (s *IntegrationTestSuite) SetupTestGlobalFeeStoreAndMinGasPrice(minGasPrice
 	return subspace
 }
 
+// in the fee test, set uatom as bond denom.
```
Pantani:
```suggestion
// SetupTestStakingSubspace sets uatom as bond denom for the fee tests.
```

x/globalfee/ante/fee_utils.go
```diff
@@ -31,6 +32,24 @@ func (mfd FeeDecorator) getGlobalFee(ctx sdk.Context, feeTx sdk.FeeTx) sdk.Coins
 	return requiredGlobalFees.Sort()
 }
 
+func (mfd FeeDecorator) DefaultZeroGlobalFee(ctx sdk.Context) []sdk.DecCoin {
+	bondDenom := mfd.getBondDenom(ctx)
+	if bondDenom == "" { //todo: check bonddenom is only one denom
```
Pantani: The lint is complaining, maybe we should add the TODO comment above the `if`.

Pantani (approved):

ante/ante.go
Pantani: also, instead return `sdkerrors.ErrLogic` we can create new custom errors for Gaia, like `gaiaerrors.ErrSubspaceNotFound`

yaruwangway: shall we use `sdk.ErrNotFound` instead?

x/globalfee/ante/antetest/fee_test_setup.go
yaruwangway: shall we use `sdk.ErrNotFound` instead?

ante/ante.go
yaruwangway: actually sdk uses `ErrLogic` in [ this situation](https://github.com/cosmos/cosmos-sdk/blob/18d8b64ce596131b986a96bfa44524cb9625970b/x/auth/ante/ante.go#L27).

x/globalfee/ante/fee_utils.go
```diff
@@ -31,6 +32,25 @@ func (mfd FeeDecorator) getGlobalFee(ctx sdk.Context, feeTx sdk.FeeTx) sdk.Coins
 	return requiredGlobalFees.Sort()
 }
 
+func (mfd FeeDecorator) DefaultZeroGlobalFee(ctx sdk.Context) []sdk.DecCoin {
+	bondDenom := mfd.getBondDenom(ctx)
+	// todo: check bonddenom is only one denom
+	if bondDenom == "" {
+		panic("staking bond denom is empty")
```
Pantani: Can we return an error and use `s.T().Fatalf(err)` in the parent method?

Pantani (approved):
LGTM. Only a one suggestions

x/globalfee/ante/fee_utils.go
yaruwangway: Hi, you meant not panic in `DefaultZeroGlobalFee`, but return error `func (mfd FeeDecorator) DefaultZeroGlobalFee(ctx sdk.Context) ([]sdk.DecCoin, error)`, and then panic [here](https://github.com/cosmos/gaia/blob/b0b59461682a10c920646e65faed2c82d019f066/x/globalfee/ante/fee_utils.go#L21) if err ?

Pantani: yess!!! WDYT? This will be more clear to debug in the test logs

x/globalfee/ante/antetest/fee_test.go
```diff
@@ -19,6 +19,22 @@ func TestIntegrationTestSuite(t *testing.T) {
...
+func (s *IntegrationTestSuite) TestGetDefaultGlobalFees() {
+	bondDenom := "uatom"
+	// set globalfees and min gas price
+	globalfeeSubspace := s.SetupTestGlobalFeeStoreAndMinGasPrice([]sdk.DecCoin{}, &globfeetypes.Params{})
+	stakingSubspace := s.SetupTestStakingSubspace(bondDenom)
+	// setup antehandler
+	mfd := gaiafeeante.NewFeeDecorator(gaiaapp.GetDefaultBypassFeeMessages(), globalfeeSubspace, stakingSubspace)
+
+	defaultGlobalFees, err := mfd.DefaultZeroGlobalFee(s.ctx)
+	s.Require().NoError(err)
```
Pantani, line 37:
```suggestion
	s.Require().NoError(err)
	s.Require().Greater(len(defaultGlobalFees), 0)
```

x/globalfee/ante/fee.go
```diff
@@ -97,3 +101,47 @@ func (mfd FeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, ne
...
+	var globalMinGasPrices sdk.DecCoins
+	var err error
+	if mfd.GlobalMinFee.Has(ctx, types.ParamStoreKeyMinGasPrices) {
+		mfd.GlobalMinFee.Get(ctx, types.ParamStoreKeyMinGasPrices, &globalMinGasPrices)
+	}
+	// global fee is empty set, set global fee to 0uatom
+	if len(globalMinGasPrices) == 0 {
+		globalMinGasPrices, err = mfd.DefaultZeroGlobalFee(ctx)
+		if err != nil {
+			panic(err)
```
Pantani: IMHO we should avoid panic in the helper method, the panic should stay in the first layer of the code or use test fail. But it is something we should create an issue to try to handle panics from the code and merge this PR as is

x/globalfee/ante/fee.go
```diff
@@ -97,3 +101,47 @@ func (mfd FeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, ne
 
 	return next(ctx, tx, simulate)
 }
+
+// ParamStoreKeyMinGasPrices type require coins sorted. getGlobalFee will also return sorted coins (might return 0denom if globalMinGasPrice is 0)
+func (mfd FeeDecorator) getGlobalFee(ctx sdk.Context, feeTx sdk.FeeTx) sdk.Coins {
+	var globalMinGasPrices sdk.DecCoins
+	var err error
```
Pantani:
```suggestion
	var (
		globalMinGasPrices sdk.DecCoins
		err error
	)
```

x/globalfee/ante/fee.go
```diff
@@ -97,3 +101,47 @@ func (mfd FeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, ne
...
+		requiredGlobalFees[i] = sdk.NewCoin(gp.Denom, fee.Ceil().RoundInt())
+	}
+
+	return requiredGlobalFees.Sort()
+}
+
+func (mfd FeeDecorator) DefaultZeroGlobalFee(ctx sdk.Context) ([]sdk.DecCoin, error) {
+	bondDenom := mfd.getBondDenom(ctx)
+	if bondDenom == "" {
+		return []sdk.DecCoin{}, errors.New("empty staking bond denomination")
```
Pantani: avoid allocating a new slice into the memory if we are not using it

```suggestion
		return nil, errors.New("empty staking bond denomination")
```

x/globalfee/ante/antetest/fee_test_setup.go
```diff
@@ -56,8 +56,8 @@ func (s *IntegrationTestSuite) SetupTestGlobalFeeStoreAndMinGasPrice(minGasPrice
 }
 
 // SetupTestStakingSubspace sets uatom as bond denom for the fee tests.
-func (s *IntegrationTestSuite) SetupTestStakingSubspace(bondDenom string) types.Subspace {
-	s.app.GetSubspace(stakingtypes.ModuleName).Set(s.ctx, stakingtypes.KeyBondDenom, bondDenom)
+func (s *IntegrationTestSuite) SetupTestStakingSubspace(params stakingtypes.Params) types.Subspace {
+	s.app.GetSubspace(stakingtypes.ModuleName).SetParamSet(s.ctx, &params)
 	return s.app.GetSubspace(stakingtypes.ModuleName)
```
yaruwangway, line 61: Hi  @Pantani,  I changed `set` to `SetParamSet `. in sdk, when using `setParamSet`, it will call a validate  function , which will validate if the `bondDenom` is empty, if empty, it will panic. 
this means in our app, when `init`, the staking bondDenom is not empty because `setParamSet` in the call stack of `init`.

so I m not sure now if I still check if bondDenom empty [here](https://github.com/cosmos/gaia/blob/ce9762e3bcbb69d7e08811767da4d522cfd39ae7/x/globalfee/ante/fee.go#L133)
extra checking is always ok ?

x/globalfee/ante/fee.go
yaruwangway: I changed the panic to antehandler now

okwme (approved):

PR #1826, opened 2022-10-15
change clone command

tacolopo:
with current command I was getting error: 
```git@github.com: Permission denied (publickey).```
Which was resolved by changing the command to the url

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -61,7 +61,7 @@ Next, let's install the latest version of Gaia. Make sure you `git checkout` the
 correct [released version](https://github.com/cosmos/gaia/releases).
 
 ```bash
-git clone -b <latest-release-tag> git@github.com:cosmos/gaia.git
+git clone -b <latest-release-tag> https://github.com/cosmos/gaia.git
 cd gaia && make install
 ```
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by okwme — APPROVED** [review]
thank you @tacolopo !

File: docs/getting-started/installation.md [hunk 1]
```diff
@@ -61,7 +61,7 @@ Next, let's install the latest version of Gaia. Make sure you `git checkout` the
 correct [released version](https://github.com/cosmos/gaia/releases).
 
 ```bash
-git clone -b <latest-release-tag> git@github.com:cosmos/gaia.git
+git clone -b <latest-release-tag> https://github.com/cosmos/gaia.git
```
Pantani, line 64: You must add your SSH key to your GitHub for the command works. It's a common thing

okwme: i agree that it is common for github users to do this. However I also agree that the docs should be set for users with as little contextual knowledge as possible. If someone knows how to set their SSH key in github, they also know that they can change the instructions to `git@github.com...` for the SSH clone command. I think it's a good change to make here and also includes an update to the latest release of v7 family of gaia.

okwme (approved):

PR #1840, opened 2022-10-24
refactor: e2e test

yaruwangway:
This refactor includes:
- [x] refactor ica test into one func to run in `e2e_test.go`
- [x] move the repeated gov proposal of new global fee in global fee test into `govProposeNewGlobalfee`


tests/e2e/e2e_exec_test.go
```diff
@@ -292,6 +292,32 @@ func (s *IntegrationTestSuite) execBankSend(
...
+	amt       string
+	fees      string
+	log       string
+	expectErr bool
+}
+
+func (s *IntegrationTestSuite) execBankSendBatch(
+	c *chain,
+	valIdx int,
+	txs []txBankSend,
```
Pantani:
```suggestion
	txs ...txBankSend,
```

Can we use a variadic function here?

tests/e2e/e2e_exec_test.go
```diff
@@ -292,6 +292,32 @@ func (s *IntegrationTestSuite) execBankSend(
...
+) int {
+	sucessBankSendCount := 0
```
Pantani, line 311: wdyt to use an `errgroup` package here?

tests/e2e/e2e_globalfee_proposal_test.go
```diff
@@ -0,0 +1,44 @@
...
+	govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
+)
+
+func (s *IntegrationTestSuite) govProposeNewGlobalfee(newGlobalfee sdk.DecCoins, proposalCounter int, submitter string, fees string) {
+	s.writeGovParamChangeProposalGlobalFees(s.chainA, newGlobalfee)
+	chainAAPIEndpoint := fmt.Sprintf("http://%s", s.valResources[s.chainA.id][0].GetHostPort("1317/tcp"))
+	// gov proposing new fees
+	s.T().Logf("Proposal number: %d", proposalCounter)
+	s.T().Logf("Submitting, deposit and vote legacy Gov Proposal: change global fee to %s", newGlobalfee.String())
+	s.submitLegacyGovProposal(chainAAPIEndpoint, submitter, fees, "param-change", proposalCounter, configFile("proposal_globalfee.json"))
```
Pantani: can we use const for the json variable names?

```go
const proposalGlobalFee = "proposal_globalfee.json"
```

tests/e2e/e2e_exec_test.go
yaruwangway: run banksend in concurrently ?

yaruwangway: Hi I refactored `execBankSendBatch` to use goroutine.  Since `execBankSend` does not return err, so I do not use `errgroup`, WDYT?

tests/e2e/e2e_exec_test.go
yaruwangway: we can, but why this is better ?

tests/e2e/e2e_exec_test.go
```diff
@@ -292,6 +293,55 @@ func (s *IntegrationTestSuite) execBankSend(
...
+
+			s.T().Logf(txs[idx].log)
+			s.execBankSend(c, valIdx, txs[idx].from, txs[idx].to, txs[idx].amt, txs[idx].fees, txs[idx].expectErr)
+			if !txs[idx].expectErr {
+				sucessBankSendCount.inc()
+			}
+		}(i)
+
+		wg.Wait()
+	}
```
Pantani: The wait call should be called outside the loop. If not, the `waitgroup` will run synchronously

```suggestion
	}
	wg.Wait()
```

tests/e2e/e2e_exec_test.go
Pantani: Because it can be reusable to send only one transaction without creating a slice and pass to the function. But it is more of a suggestion. Both are good

Pantani (approved):
lgmt

PR #1849, opened 2022-10-26
Update validator-faq.md

khaliduddin:
This word "higher change" on line 297 must be updated to "higher chance" as per my understanding. Please ignore if my understanding is wrong.


okwme (approved):
thank you @khaliduddin  you're right!

PR #1854, opened 2022-10-27
ignore workspace sum

okwme:
as per recommendation from @tac0turtle and @julienrbrt to act the same as Cosmos SDK


julienrbrt:
You may want to delete too https://github.com/cosmos/gaia/blob/main/go.work.sum

PR #1855, opened 2022-10-28
omnnibus patch for gaia v8

faddat:
Omnibus cosmos ecosystem patches for Gaia v8.

I haven't made a changelog entry because I don't view this as particularly signficant.


okwme:
thanks @faddat 
I've already got these changes as part of https://github.com/cosmos/gaia/pull/1834/
the only difference here is the Changelog which is missing a number of other updates so would rather treat as a dedicated PR with all missing entries at the same time.
Going to close this in place of https://github.com/cosmos/gaia/pull/1834/ but appreciate the turnaround from when IBC v5.0.1 was released, i've been eagerly awaiting as well.

PR #1858, opened 2022-10-28
test: Unit tests for global fee module

yaruwangway:


okwme (approved):

Pantani:
I think we can also add tests to all methods from [x/globalfee/ante/fee_utils.go](https://github.com/cosmos/gaia/blob/b0289b656957fa1cfcefd777f89a1405f10b7b24/x/globalfee/ante/fee_utils.go#L1-L1) file

Pantani:
Also, the CLI query functions can be tested. Take a look at this example:

https://github.com/tendermint/spn/blob/develop/x/campaign/client/cli/query_campaign_chains_test.go

yaruwangway:
hi @Pantani , the fee_util_test is in `x/globalfee/ante/antetest`

x/globalfee/types/params_test.go
```diff
@@ -0,0 +1,58 @@
...
+			},
+			true,
+		},
+	}
+
+	for name, test := range tests {
+		t.Log(name)
+		err := validateMinimumGasPrices(test.coins)
+
+		require.Equal(t, err == nil, !test.expectErr)
```
Pantani:
```suggestion
		t.Run(name, func(t *testing.T) {
			err := validateMinimumGasPrices(test.coins)
			if test.expectErr {
				require.Error(t, err)	
				return
			}
			require.NoError(t, err)
		})
```

Subtests make the test logs more clear. And using the `require.NoError` is better because this will print the error if it fails

PR #1862, opened 2022-11-01
Updated and new links

ZorroZ77:
Current Telegram link directs to a chat that no longer exists, replaced it with the main Cosmos Telegram group. Also added link to Mintscan's page of on-chain proposals


okwme (approved):
thank you @ZorroZ77 !

PR #1877, opened 2022-11-04
refactor: simplify the logic in fee antehandler

### LINKED ISSUE #1872: x/globalfee/ante: return early if in simulate mode or if !ctx.IsCheckTx() inside FeeDecorate.AnteHandle
## Summary

The code inside https://github.com/cosmos/gaia/blob/842f231058803329851bf4f5ac4c9710f3e08928/x/globalfee/ante/fee.go#L52 could simplify and return early when in simulate mode or if !ctx.IsCheckTx() to avoid the heavy conditioning
## Type
Code debt

## Impact

Ease of readability and more idiomatic Go

## Proposed Solution

image
____

#### For Admin Use

- [x] Not duplicate issue

yaruwangway:
closes:  #1872
# refactor fee antehandler:
- return early when `!ctx.IsCheckTx() && simulate`
```	
if !ctx.IsCheckTx() && simulate {
		return next(ctx, tx, simulate)
	}
```
- move `getGlobalFee` out if condition.
- add one more test case for disable checkTx: tx fee's denom is not in global fees denoms set

### DIFF
--- a/x/globalfee/ante/antetest/fee_test.go
+++ b/x/globalfee/ante/antetest/fee_test.go
@@ -555,6 +555,15 @@ func (s *IntegrationTestSuite) TestGlobalFeeMinimumGasFeeAnteHandler() {
 			txCheck:         false,
 			expErr:          false,
 		},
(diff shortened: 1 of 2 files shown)


Pantani (approved):
lgmt

PR #1887, opened 2022-11-09
feat: Ensure all relevant docker images are in cosmos org

### LINKED ISSUE #1645: Ensure all relevant docker images are in cosmos org
## Summary

- [x] we were using a tendermintdev dockerhub image of gaia for releases but don't have access to that account.
- [x] we should ensure that all relevant docker images for gaia are maintained here.
there's also the `Dockerfile` in the root dir of the project that I have no record of where it's being used, let's confirm where or remove it if it's redundant.
- [x] jacob's also got a docker image for use in gitpod, we should try consolidating that work, either through submodule or as a copy to ensure you don't have to go scrambling all over the place for each instance of docker relevant to the hub
other docker images we may want to create listed in #1619 

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue

mmulji-ic:
Fixes issue https://github.com/cosmos/gaia/issues/1645


okwme (commented):
I get the following error when I try to run this version of build-reproducible:
```bash
> make build-reproducible
/usr/local/bin/docker rm latest-build || true
latest-build
/usr/local/bin/docker run --volume=/Users/billy/GitHub/cosmos/gaia:/sources:ro \
        --env TARGET_PLATFORMS='linux/amd64 darwin/amd64 linux/arm64 darwin/arm64 windows/amd64' \
        --env APP=gaiad \
        --env VERSION=ignore-delete \
        --env COMMIT=e4207cc300c6b552aaaac252c63ee3c3421cab93 \
        --env LEDGER_ENABLED=true \
        --name latest-build ghcr.io/cosmos/rbuilder:latest
/sources/.build.sh: line 14: /usr/local/share/tendermint/buildlib.sh: No such file or directory
make: *** [build-reproducible] Error 1
```

faddat:
This one became shockingly important.

Pamukov81:
Vvhh

Issue #1870, opened 2022-11-03
test: upgrade test


yaruwangway:
## Summary
~~Add upgrade test in workflow for major release/rc-release during release process.
This test is triggered by a label "major_release", and is configured to upgrade from last major release to present version.~~

EDIT: create nightly build that can be passed to Hypha for testing version-specific features and simulating an upgrade from the current mainnet version

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

okwme:
- [x] fix the denom KV storage issue and overwrite with correct data
- [x] make issue on sdk about bank kv store being inconsistent between v045 and v0.46
- [x] convert script to use cosmovisor so that the panic will actually kill gaiad running and allow the upgrade to take place
- [ ] make a new script that uses recent theta testnet genesis instead of modified genesis from scratch
- [x] build last release as part of action CI and move to ./build/gaiad7 to run legacy node
- [x] figure out a way to read the output automatically to determine whether the upgrade succeeded or not
    - [x] check the liveness test logic

MSalopek:
Closing as completed in [#2328](https://github.com/cosmos/gaia/issues/2328)

Please favor reopening [#2328](https://github.com/cosmos/gaia/issues/2328) in case additional work needs to be done or there is more information.

> EDIT: create nightly build that can be passed to Hypha for testing version-specific features and simulating an upgrade from the current mainnet version

Currently the relevant `release/*` branches and `main` are used.

PR #1892, opened 2022-11-10
fix bank migration in upgrade handler

okwme:
part one of issue #1870


.github/workflows/test.yml
```diff
@@ -102,3 +102,44 @@ jobs:
...
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+      - uses: technote-space/get-diff-action@v6.0.1
+        with:
+          PATTERNS: |
+            **/**.go
+            go.mod
+            go.sum
+      - name: Install GaiaV7
```
yaruwangway, line 129: we can make this more  general, we can configure the upgradeFrom version and upgradeTo version 
1. by providing a label of versions, for example, upgradeFrom Version: v7.0.0. 
or 
3. maybe use present release tag -1, so if we tag this release v8.0.0, it will trigger and upgrade test from v7.0.0 to v8.0.0

here is an example of [set tag of release version as env.](https://github.com/cosmos/gaia/blob/719b81d1837c1fa189e44bea6ab12b29ed90d8ea/.github/workflows/release.yml#L16)

okwme: let's keep that as part of a future update, i think it's still good to contain upgrades between various hard-coded versions to ensure backported changes don't break upgrades

app/upgrades/v8/upgrades.go
```diff
@@ -3,29 +3,73 @@ package v8
...
+	if foundMalformed {
+		// save it with the correct denom
+		keepers.BankKeeper.SetDenomMetaData(ctx, atomMetaData)
+
+		// delete the old format
+		key := keepers.GetKey(banktypes.ModuleName)
+		store := ctx.KVStore(key)
+		oldDenomMetaDataStore := prefix.NewStore(store, banktypes.DenomMetadataPrefix)
+		oldKey := make([]byte, len(malformedDenom))
+		copy(oldKey, []byte(malformedDenom))
```
Pantani:
```suggestion
		copy(oldKey, malformedDenom)
```

contrib/denom.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    {
+      "aliases": [],
+      "denom": "atom",
+      "exponent": 6
+    }
+  ],
+  "description": "The native staking token of the Cosmos Hub.",
+  "display": "atom",
+  "name": "asdf",
```
Pantani, line 28: Is this right?

contrib/scripts/run-gaia-v7.sh
```diff
@@ -0,0 +1,71 @@
...
+  echo "has not gaiad7"
+  exit
+fi
+
+
+rm -rf ./build/.gaia
+
+mkdir -p $NODE_HOME/cosmovisor/genesis/bin
+cp ./build/gaiad7 $NODE_HOME/cosmovisor/genesis/bin/gaiad
+$BINARY init upgrader --chain-id $CHAINID --home $NODE_HOME
```
Pantani: double quote to prevent globbing and word splitting

```suggestion
mkdir -p "$NODE_HOME"/cosmovisor/genesis/bin
cp ./build/gaiad7 "$NODE_HOME"/cosmovisor/genesis/bin/gaiad
$BINARY init upgrader --chain-id $CHAINID --home "$NODE_HOME"
```

contrib/scripts/run-gaia-v7.sh
```diff
@@ -0,0 +1,71 @@
...
+
+
+if ! test -f "./build/gaiad8"; then
+  echo "has not gaiad v8"
+  exit
+fi
+
+mkdir -p $NODE_HOME/cosmovisor/upgrades/v8-Rho/bin
+cp ./build/gaiad8 $NODE_HOME/cosmovisor/upgrades/v8-Rho/bin/gaiad
```
Pantani:
```suggestion
mkdir -p "$NODE_HOME"/cosmovisor/upgrades/v8-Rho/bin
cp ./build/gaiad8 "$NODE_HOME"/cosmovisor/upgrades/v8-Rho/bin/gaiad
```

Pantani: same for other usages of `$NODE_HOME` and `$USER_MNEMONIC`

contrib/scripts/run-gaia-v7.sh
```diff
@@ -0,0 +1,71 @@
...
+NODE_HOME=$(realpath ./build/.gaia)
+echo "NODE_HOME = ${NODE_HOME}"
+BINARY=$NODE_HOME/cosmovisor/genesis/bin/gaiad
+echo "BINARY = ${BINARY}"
+CHAINID=cosmoshub-4
+
+USER_MNEMONIC="abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon art"
+
+if ! test -f "./build/gaiad7"; then
+  echo "has not gaiad7"
```
Pantani:
```suggestion
  echo "gaiad v7 does not exist"
```

contrib/scripts/run-gaia-v7.sh
```diff
@@ -0,0 +1,71 @@
...
+
+rm -rf ./build/.gaia
+
+mkdir -p $NODE_HOME/cosmovisor/genesis/bin
+cp ./build/gaiad7 $NODE_HOME/cosmovisor/genesis/bin/gaiad
+$BINARY init upgrader --chain-id $CHAINID --home $NODE_HOME
+
+
+if ! test -f "./build/gaiad8"; then
+  echo "has not gaiad v8"
```
Pantani:
```suggestion
  echo "gaiad v8 does not exist"
```

contrib/denom.json
okwme: this is dummy info because you can't start a chain with an invalid genesis file and the problem that was originally addressed is that we can't export a valid genesis file until we fill these values. So for the upgrade test they have to be filled with something even though we're replacing the content with what _shoud_ have been there all along.

okwme:
Instructions for running locally that can be added to @Pantani 's PR with that kind of info:

## Build current version and move into ./build:
```sh
make build
cp ./build/gaiad ./build/gaiad8
```

 ## Build gaia v7.1.0 and move into ./build:
```sh
git checkout v7.1.0
make build
cp ./build/gaiad ./build/gaiad7
```

## Go back to current branch
```sh
git checkout -
```

## Install cosmovisor
```sh
go install github.com/cosmos/cosmos-sdk/cosmovisor/cmd/cosmovisor@v1.0.0
```

## Run the Chain

This script prepares the chain and starts it using cosmovisor
```sh
./contrib/scripts/run-gaia-v7.sh
```

## Run the upgrade
In another terminal window, run the script that waits 10 seconds for gaia to start then makes gov proposal to perform an upgrade at height 15
```sh
./contrib/scripts/run-upgrade-commands.sh 15
```

## Monitor for success
In a third window run the upgrade monitoring script that will exit without error when the upgrade succeeds.
```sh
./contrib/scripts/test_upgrade.sh 20 5 16 localhost
```

This should show logs that demonstrate a successful upgrade by reaching block height 16 before 100 seconds is reached.

app/upgrades/v8/upgrades.go
```diff
@@ -3,29 +3,73 @@ package v8
...
 	"github.com/cosmos/cosmos-sdk/types/module"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 	icacontrollertypes "github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/controller/types"
 	icahosttypes "github.com/cosmos/ibc-go/v5/modules/apps/27-interchain-accounts/host/types"
 
 	"github.com/cosmos/gaia/v8/app/keepers"
 )
 
+func fixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
```
Pantani: If it is not too hard, maybe we can add some tests to this method

app/upgrades/v8/upgrades.go
```diff
@@ -3,29 +3,73 @@ package v8
...
+	if foundMalformed {
+		// save it with the correct denom
+		keepers.BankKeeper.SetDenomMetaData(ctx, atomMetaData)
+
+		// delete the old format
+		key := keepers.GetKey(banktypes.ModuleName)
+		store := ctx.KVStore(key)
+		oldDenomMetaDataStore := prefix.NewStore(store, banktypes.DenomMetadataPrefix)
+		oldKey := make([]byte, len(malformedDenom))
+		copy(oldKey, malformedDenom)
```
yaruwangway: why copy here?

app/upgrades/v8/upgrades.go
okwme: added a test

app/upgrades/v8/upgrades.go
okwme: removed the copy

app/upgrades/v8/upgrades_test.go
```diff
@@ -0,0 +1,58 @@
...
+func TestFixBankMetadata(t *testing.T) {
+	app := gaiahelpers.Setup(t)
+	ctx := app.BaseApp.NewContext(false, tmproto.Header{
+		ChainID: fmt.Sprintf("test-chain-%s", tmrand.Str(4)),
+		Height:  1,
+	})
+
+	cdc := app.AppCodec()
+
+	malformedDebom := "uatomu"
```
Pantani:
```suggestion
	malformedDenom := "uatomu"
```

app/upgrades/v8/upgrades_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	correctDenom := "uatom"
+
+	_, foundCorrect := app.AppKeepers.BankKeeper.GetDenomMetaData(ctx, correctDenom)
+	require.False(t, foundCorrect)
+
+	err := v8.FixBankMetadata(ctx, &app.AppKeepers)
+	require.NoError(t, err)
+
+	_, foundCorrect = app.AppKeepers.BankKeeper.GetDenomMetaData(ctx, correctDenom)
```
Pantani, line 55: Can we also check if the `malformedDenom` is not there anymore?

okwme: added!

Pantani (approved):

PR #1893, opened 2022-11-13
Delete CODE_OF_CONDUCT.md

faddat:
It's time to choose. You can either admit that you have absolutely no place in having a code of conduct at all, and delete the code of conduct, or you can knock out Jae+aib (thought crime: hatefulness  and/or stirring up ethnoreligious hatred) and me (thought crime: Yes I did gender the hub, and describe her as overweight because she is a relatively slow computer program).

Freedom or censorship?

O and when someone threatens me, I do call em de badsiewordsies.

PR #1894, opened 2022-11-13
Update CODE_OF_CONDUCT.md

faddat:
Make more real

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,7 +1,7 @@
 
 # Contributor Covenant Code of Conduct
 
-## Our Pledge
+## Our Pledge is worthless
 
 We as members, contributors, and leaders pledge to make participation in our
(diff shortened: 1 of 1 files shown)


mmulji-ic:
Morning @faddat, the code of conduct is being reviewed and input is being sought a number of parties. In any incarnation, action should be taken against anyone who breaches it, not a subset from org A or org B. Should universally apply. Am chasing the status and will get back to you asap.

faddat:
No.

The code of conduct is completely useless. It is effectively a lie to the community. We need to delete it or merge the modified version because we should tell the truth or expect that we are going to lose the communities trust.

By the way just because I think that proposal 82 is a good idea doesn't mean I'm your puppet, I'm also not the puppet of AIB.

jackzampolin:
I also concur that the code of conduct no longer serves any purpose

jackzampolin:
Closing this in favor of #1893

PR #1900, opened 2022-11-15
docs: correct `NoWithVeto` condition description

whylee259:
According to [docs](https://docs.cosmos.network/main/modules/gov#threshold) and [code](https://github.com/cosmos/cosmos-sdk/blob/bcff22a3767b9c5dd7d1d562aece90cf72e05e85/x/gov/keeper/tally.go#L115), it includes `abstain` when calculating the threshold of `NoWithVeto`.


hxrts:
oh yeah, good catch. this went into a cosmos hub gov proposal 75 as well.

hxrts (approved):

okwme (approved):
thank you @whylee259 !

PR #1902, opened 2022-11-15
spike to test feasibility of downgrade to v0.45

okwme:


faddat:
Hi, I would really like to know what is driving this spike?

Happy to work on it with you and not my preferred route.

okwme:
ICS still uses v0.45.x and there's consideration for taking a much more conservative upgrade policy for the Cosmos Hub wrt releases of the SDK. v0.45 is the only currently battletested version of the SDK and if the Hub's new premier feature is security it should realistically be one of the last chains to upgrade to new releases of the SDK. It kind of breaks my heart after all the work put into gaia with v0.46 but we want to help ICS meet it's timeline so attempting to downgrade to v0.45 and preserve as many tests as possible, plus global min fee.

mmulji-ic:
@okwme Probably should've backported a subset of the tests first and get that working, rather than everything at once approach.

okwme:
I think having a version of gaia that compiles with v0.45 is a first step and tests and modules that prevent it from compiling should be removed as a first step. This PR has achieved that so far and so i thought to check whether the broken tests are fixable as a next step.

app/app.go
```diff
@@ -196,17 +198,17 @@ func NewGaiaApp(
 				SignModeHandler: encodingConfig.TxConfig.SignModeHandler(),
 				SigGasConsumer:  ante.DefaultSigVerificationGasConsumer,
 				// TxFeeChecker is not the default fee check, it will not check if the fee meets min_gas_price, this is checked in NewFeeWithBypassDecorator already.
-				TxFeeChecker: func(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
-					feeTx, ok := tx.(sdk.FeeTx)
-					if !ok {
-						return nil, 0, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
-					}
+				// TxFeeChecker: func(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
```
glnro, line 201: Why keep this in but commented out?

app/app.go
```diff
@@ -348,12 +347,14 @@ func (app *GaiaApp) RegisterTxService(clientCtx client.Context) {
...
 func (app *GaiaApp) RegisterTendermintService(clientCtx client.Context) {
-	tmservice.RegisterTendermintService(
-		clientCtx,
-		app.BaseApp.GRPCQueryRouter(),
-		app.interfaceRegistry,
-		app.Query,
-	)
+	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
+
+	// tmservice.RegisterTendermintService(
```
glnro: Why is this commented out as well?

cmd/gaiad/cmd/testnet.go
```diff
@@ -30,8 +30,9 @@ import (
 	tmrand "github.com/tendermint/tendermint/libs/rand"
 	"github.com/tendermint/tendermint/types"
 	tmtime "github.com/tendermint/tendermint/types/time"
-	// ibcclienttypes "github.com/cosmos/ibc-go/v5/modules/core/02-client/types"
-	// ibcchanneltypes "github.com/cosmos/ibc-go/v5/modules/core/04-channel/types"
+
+	// ibcclienttypes "github.com/cosmos/ibc-go/v3/modules/core/02-client/types"
```
glnro, line 34: Why leave this in?

docs/modules/gov.md
```diff
@@ -22,7 +22,7 @@ The **[Cosmos Hub Forum](https://forum.cosmos.network/)** is the main discussion
 
 #### Submission Updates
 
-Additionally the new gov module now accommodates proposals with multiple arbitrary message execution. For `v0.46` of the Cosmos SDK, only `MsgSend`, `MsgSoftwareUpgrade`, and `MsgCancelUpgrade` are supported by executing `submit-proposal`. Signaling proposals can also be submitted with the new api by leaving the messages field empty and populating the metadata field with the corresponding proposal text. The remaining proposal types must use `submit-legacy-proposal` until more message types are supported.
+Additionally the new gov module now accommodates proposals with multiple arbitrary message execution. For `v0.46` of the Cosmos SDK, only `MsgSend`, `MsgSoftwareUpgrade`, and `MsgCancelUpgrade` are supported by executing `submit-proposal`. Signaling proposals can also be submitted with the new api by leaving the messages field empty and populating the metadata field with the corresponding proposal text. The remaining proposal types must use `submit-proposal` until more message types are supported.
```
glnro: This is not relevant for sdk v0.45...so we should remove this section for the time being.

tests/e2e/e2e_setup_test.go
```diff
@@ -621,86 +609,75 @@ func noRestart(config *docker.HostConfig) {
 }
 
 func (s *IntegrationTestSuite) writeGovProposals(c *chain) {
```
glnro, line 611: Why comment this out instead of removing its usage?

tests/e2e/e2e_setup_test.go
```diff
@@ -621,86 +609,75 @@ func noRestart(config *docker.HostConfig) {
...
+	// upgradeProposalBody, err := cdc.MarshalJSON(proposalSendMsg)
+	// s.Require().NoError(err)
+
+	// path := filepath.Join(c.validators[0].configDir(), "config", "proposal_3.json")
+	// err = writeFile(path, upgradeProposalBody)
+	// s.Require().NoError(err)
+	// fmt.Println("saved proposal_3.json to ", path)
 }
 
 func (s *IntegrationTestSuite) writeGovCancelUpgradeSoftwareProposal(c *chain) {
```
glnro, line 669: Same from above

tests/e2e/query.go
```diff
@@ -13,12 +13,13 @@ import (
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	disttypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
 	evidencetypes "github.com/cosmos/cosmos-sdk/x/evidence/types"
-	govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
-	"github.com/cosmos/cosmos-sdk/x/group"
+
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	// govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
+	// "github.com/cosmos/cosmos-sdk/x/group"
```
glnro:
```suggestion
```

app/app.go
```diff
@@ -123,12 +121,16 @@ func NewGaiaApp(
...
+
+	moduleAccountAddresses := app.ModuleAccountAddrs()
+
 	// Setup keepers
 	app.AppKeepers = keepers.NewAppKeeper(
 		appCodec,
 		bApp,
 		legacyAmino,
 		maccPerms,
+		moduleAccountAddresses,
```
glnro, line 133: Call is duplicated on [line 291](https://github.com/cosmos/gaia/pull/1902/files#diff-0f1d2976054440336a576d47a44a37b80cdf6701dd9113012bce0e3c425819b7L289)

app/app.go
yaruwangway: because fee antehandler sdk 0.45 is different from sdk 0.46

docs/modules/gov.md
okwme:
```suggestion
```

tests/e2e/e2e_setup_test.go
okwme: because it should be added back in some format rather than removed completely as other things that aren't going to make it into Rho with v0.45

glnro (approved):

PR #1908, opened 2022-11-17
WIP: feat(build): use distroless ko image to build the app

### LINKED ISSUE #1620: build: use ko to build Go applications instead of the Alpine image to statically link against CGO for better security when building gaia binaries
## Summary

A security assesment from Chainguard for supply chain security yielded a recommendation that given we use CGO in some of this code, that we should use ko directly to build gaiad images e.g when using clevel_db

## Proposal

from the Chainguard report
>ko is a tool
for building go applications and distributing them as OCI containers. The default base image for
images built with ko is a distroless base image, which will be distroless.dev/static:latest in the
next ko release., which is sufficient for supporting the gaiad binary.
We have noticed that there are variants of the gaiad binary which contain components built with
cgo, notably when building gaiad with support for cleveldb using the cleveldb build tag. For
stability and reduced binary size, Chainguard strongly recommends building these components
statically linked against musl. To facilitate this, the Distroless project provides a
`distroless.dev/ko` image which we will use to build the final gaiad image.

distroless.dev/ko image to build a gaiad image using docker from the root of the cosmos/gaia
repository by running:
$ KO_DOCKER_REPO=... ko build ./cmd/gaiad
The KO_DOCKER_REPO variable should be set to the prefix of the repository you want to publish
to, for example, ghcr.io/cosmos for Cosmos Network. When ko completes, it will print the
sha256 reference that can be used with commands such as docker pull, etc.

/cc @kirbyquerby from Orijtech Inc to help out with addressing this
____

#### For Admin Use

- [x] Not duplicate issue

Pantani:
closes #1620

### DIFF
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -9,7 +9,7 @@ COPY . .
 ENV KO_DOCKER_REPO=ghcr.io/cosmos
 ENV PACKAGES curl make git libc-dev bash gcc linux-headers eudev-dev python3
 RUN apk add --no-cache $PACKAGES
-RUN make install
+RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
 FROM distroless.dev/ko:$IMG_TAG
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -8,10 +8,10 @@ RUN go mod download
 COPY . .
 ENV PACKAGES curl make git libc-dev bash gcc linux-headers eudev-dev python3
 RUN apk add --no-cache $PACKAGES
-RUN CGO_ENABLED=0 make install
+RUN make install
 
 # Add to a distroless container
-FROM distroless.dev/static:$IMG_TAG
+FROM distroless.dev/ko:$IMG_TAG
 ARG IMG_TAG
 COPY --from=gaiad-builder /go/bin/gaiad /usr/local/bin/
 EXPOSE 26656 26657 1317 9090
--- a/e2e.Dockerfile
+++ b/e2e.Dockerfile
@@ -6,6 +6,7 @@ WORKDIR /src/app/
 COPY go.mod go.sum* ./
 RUN go mod download
 COPY . .
+ENV KO_DOCKER_REPO=ghcr.io/cosmos
 ENV PACKAGES curl make git libc-dev bash gcc linux-headers eudev-dev python3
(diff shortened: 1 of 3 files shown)


e2e.Dockerfile
```diff
@@ -6,12 +6,13 @@ WORKDIR /src/app/
...
 RUN go mod download
 COPY . .
+ENV KO_DOCKER_REPO=ghcr.io/cosmos
 ENV PACKAGES curl make git libc-dev bash gcc linux-headers eudev-dev python3
 RUN apk add --no-cache $PACKAGES
 RUN CGO_ENABLED=0 make install
 
 # Add to a distroless container
-FROM distroless.dev/static:$IMG_TAG
+FROM distroless.dev/ko:$IMG_TAG
```
imjasonh, line 15: This won't have the intended effect. This still uses `docker build` to build the image, this change only updates the base image the resulting image would be based on.

If you want to use `ko` to build instead of `docker build`, you'll need to replace invocations of `docker build` with `ko build ./cmd/gaiad`, which will replicate the steps of this Dockerfile (e.g., `go build`) and put the resulting binary on top of the same distroless static image you're already using.

As a benefit, `ko` doesn't require Docker to build, and also generates an SBOM by default. If you want some help with this, let me know and I'd be happy to help with a PR. https://ko.build/advanced/migrating-from-dockerfile/ might also be helpful (if not, let me know and I'll fix it!)

yaruwangway: @imjasonh , can you help make a pr to fix this ? Thank you!

e2e.Dockerfile
```diff
@@ -6,12 +6,13 @@ WORKDIR /src/app/
 COPY go.mod go.sum* ./
 RUN go mod download
 COPY . .
+ENV KO_DOCKER_REPO=ghcr.io/cosmos
```
mmulji-ic, line 9: Think we should explicitly specify gaia repo, otherwise will create something else
export KO_DOCKER_REPO=ghcr.io/cosmos/gaia

yaruwangway: Hi, I m not sure about this, because you see here does not have a repo name.https://github.com/orgs/cosmos/packages

yaruwangway:
Hi @odeke-em , for https://github.com/cosmos/gaia/issues/1620, if you can have a look if this pr correctly addressed the issue, thank  you !

PR #1912, opened 2022-11-17
chore(deps) Dependency Bumps [chain-chores]

reecepbcups:
Generated with [chain-chores](https://github.com/notional-labs/chain-chores)

Changes:
- tendermint/tendermint v0.34.22 -> tendermint/tendermint v0.34.23


okwme:
I believe Tendermint and SDK don't do a minor release if it could impact the determinism of the downstream application. However it's always unclear to me whether other dependencies will have adverse effects. Is the chain-chore tool built in a way to determine dependencies that might impact determinism?

reecepbcups:
@okwme Not yet. This is in the pipeline for me to do this upcoming week

Flow will be:
- Bump versions to latest minor revisions
- go test
- state sync to mainnet & confirm
- commit, push, PR

Right now statesync is not yet done to confirm this, but will be the next time I push to any repos :)

okwme:
thanks @Reecepbcups !
we'll wait for the statesync tests to merge.

PR #1913, opened 2022-11-18
test re-org

okwme:
moved tests around so that they're simply listed inside of `e2e_test.go`. This is where we should be adding and removing the `s.T().Skip()` statements. It makes it easier to isolate a specific test when developing as well so that you don't have to run all tests to check that one is passing.


tests/e2e/e2e_staking_test.go
```diff
@@ -1,17 +1,32 @@
...
 
 import (
+	"fmt"
 	"time"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func (s *IntegrationTestSuite) testStaking(chainEndpoint string, delegatorAddress string, validatorAddressA string, validatorAddressB string, fees sdk.Coin, homePath string) {
+func (s *IntegrationTestSuite) testStaking() {
```
yaruwangway, line 10:
```suggestion
func (s *IntegrationTestSuite) TestStaking() {
```

yaruwangway (approved):

tests/e2e/e2e_staking_test.go
okwme: bby keeping it lower case it allows it to be called along with all the other tests in `e2e_test.go`.

Issue #1909, opened 2022-11-17
v0.45 compatible Rho Meta Issue


okwme:
In order to improve chances of ICS launching on time we're considering using v0.45.x for the next upgrade Rho. This may or may not include ICS itself but nonetheless should decrease the work needed to integrate ICS. Since the Cosmos Hub is becoming a security critical chain (security as the primary product offering) we are also considering a much more conservative timeline to updating to the latest release of the Cosmos SDK. To follow this path the following work needs to be done:

- [x] Downgrade gaia to v0.45.x (@okwme https://github.com/cosmos/gaia/pull/1902)
  - [x]  downgrade dependencies (IBC, Liquidity, IBC-Packet-Forward Middleware, Tendermint)
  - [x] Remove ICA Mauth
  - [x] Remove Groups
  - [x] Downgrade Gov to old gov
  - [x] Remove Global Fee Antehandler
  - [x] Remove broken tests
- [x] Add Back Global fee Antehandler (@yaruwangway https://github.com/cosmos/gaia/pull/1918)
  - [x] Needs to be fixed
  - [x] Needs tests added back
- [x] Add back IBC Packet Forward Middleware tests (@okwme #1941 )
  - [x] Needs to confirm that IBC packet forward middleware has a working version compatible with IBC v3.x.x
- [x] Fix Fee Grant E2E tests (@yaruwangway https://github.com/cosmos/gaia/pull/1940)
- [x] Add back `TestGaiaApp_Export` (@okwme #1955)
- [x] Check what `GetStakingKeeper` was doing and why it's needed
- [x] Regenerate swagger files without mauth (@yaruwangway https://github.com/cosmos/gaia/pull/1933 )
- [x] Update proto dependencies to sdk0.45, change from using buf registry to copy third_party to our repo (@yaruwangway https://github.com/cosmos/gaia/pull/1933)
- [ ] ensure all docs are up to date with changes (@yaruwangway @glnro )
  - [ ] gov is not legacy proposal anymore
  - [x] no mauth module
  - [x] no groups module
- [x] Confirm why `initTendermintConfig` was needed in `cmd/gaiad/cmd/root.go`
- [x] Confirm vesting does not have `execCreatePermanentLockedAccount` or `execCreatePeriodicVestingAccount` (@okwme  )
- [x] Add back gov tests `GovSoftwareUpgrade` , `GovCancelSoftwareUpgrade`, (@glnro https://github.com/cosmos/gaia/pull/1951 )
[...]

PR #1918, opened 2022-11-23
 fix: antehandler to fit sdk v0.45

yaruwangway:
issue description: https://github.com/cosmos/gaia/issues/1909


ante/ante.go
```diff
@@ -47,20 +48,19 @@ func NewAnteHandler(opts HandlerOptions) (sdk.AnteHandler, error) {
...
 		ante.NewDeductFeeDecorator(opts.AccountKeeper, opts.BankKeeper, opts.FeegrantKeeper),
-		// ante.NewDeductFeeDecorator(opts.AccountKeeper, opts.BankKeeper, opts.FeegrantKeeper, opts.TxFeeChecker),
 		ante.NewSetPubKeyDecorator(opts.AccountKeeper), // SetPubKeyDecorator must be called before all signature verification decorators
 		ante.NewValidateSigCountDecorator(opts.AccountKeeper),
 		ante.NewSigGasConsumeDecorator(opts.AccountKeeper, sigGasConsumer),
 		ante.NewSigVerificationDecorator(opts.AccountKeeper, opts.SignModeHandler),
 		ante.NewIncrementSequenceDecorator(opts.AccountKeeper),
+		ibcante.NewAnteDecorator(opts.IBCkeeper),
 		// TODO: check what this is
 		// ibcante.NewRedundantRelayDecorator(opts.IBCkeeper),
```
glnro: I think `NewRedundantRelayDecorator` is the API for ibc v5...probably need to revert to old API https://github.com/cosmos/ibc-go/blob/ed71af861e6dfe5fa5d942fbdc7eaa5e88767727/docs/migrations/v4-to-v5.md#ante-decorator

yaruwangway: `ibcante.NewAnteDecorator(opts.IBCkeeper)` is for sdk0.45 and ibc v4, right ? so this is correct, right?

glnro: Yeah so we can just remove the commented out line and we're good to go.

glnro (dismissed):

tests/e2e/e2e_test.go
```diff
@@ -18,8 +18,6 @@ func (s *IntegrationTestSuite) TestGov() {
 
 // globalfee in genesis is set to be "0.00001uatom"
 func (s *IntegrationTestSuite) TestQueryGlobalFeesInGenesis() {
```
glnro, line 20: Getting a failure on the zero coin fee...`feeInGenesis` is ` 0.000010000000000000uatom`, and `fees` is `0.000000000000000000photon,0.000001000000000000uatom`

yaruwangway: this is due to the globalfee last proposal of changing globalfee back to original fee failed, and query globalfee is after global fee test.

app/app.go
```diff
@@ -197,18 +197,6 @@ func NewGaiaApp(
 				FeegrantKeeper:  app.FeeGrantKeeper,
 				SignModeHandler: encodingConfig.TxConfig.SignModeHandler(),
 				SigGasConsumer:  ante.DefaultSigVerificationGasConsumer,
-				// TxFeeChecker is not the default fee check, it will not check if the fee meets min_gas_price, this is checked in NewFeeWithBypassDecorator already.
-				// TxFeeChecker: func(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
```
okwme, line 201: @yaruwangway can you confirm that it's no longer necessary to have any handler options related to bypass min fee or global min fee? Having the antehandlers (`BypassMinFeeMsgTypes` and  `GlobalFeeSubspace`) are enough to enforce the functionality?

yaruwangway: in sdk 0.45 `DeductFeeDecorator()` does not need to pass `TxFeeChecker ` as argument while sdk 0.46`DeductFeeDecorator()` need.
-  in sdk 0.46, the TxFeeChecker  will get the fees, and later a call of `checkDeductFee` will deduct these fees, if `TxFeeChecker ` is nil, it is the default feecheck, which is actually using `min_gas_price` as fees, so that we passed a custom TxFeeChecker  which will not check fees before deduct, because we have already checked the fees ok in `NewFeeDecorator(opts.BypassMinFeeMsgTypes, opts.GlobalFeeSubspace, opts.StakingSubspace)`.
https://github.com/cosmos/cosmos-sdk/blob/8cce7480917c90326c0c8a3bc2d26d1d3c29a33f/x/auth/ante/fee.go#L55-L64
- in sdk 0.45, there is no extra logic of checking fees before deduct.
https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/x/auth/ante/fee.go#L112-L117

okwme (approved):

PR #1920, opened 2022-11-23
test: disable upgrade-test

yaruwangway:
The upgrade test is targeting upgrade from v7 to v8.
The downgrade-v0-45 branch is actually v6. 
This PR disables the upgrade-test to allow all workflow to run on PRs.


okwme:
the v0.45 is not V6. it is V8. It's just a V8 that does include a change to Cosmos SDK.

PR #1933, opened 2022-12-01
feat: proto dependencies for downgrade to 0.45

yaruwangway:
sdk v0.45 does not have buf proto registry. So we have to use the previous style: 
- [x] copying all third_party proto dependencies files to our repo.
- [x] update the script generating swagger and go files

reviewers might skip  review `third_party`, the `third_party/proto/cosmos` proto files are from sdk 0.45.11


okwme (approved):
thanks @yaruwangway !

PR #1939, opened 2022-12-05
push e2e image and use released versions in docker push action

faddat:
This PR:

* uses tagged releases of github actions
* pushes the e2e docker image to ghcr


okwme (approved):
thanks @faddat

Issue #1883, opened 2022-11-07
remove usage of cosmos-sdk/simapp


tac0turtle:
## Summary of Bug

We should remove all usages of simapp we are able to, in 0.47 the dependence of simapp should be non existent. This quick and easy migration now should help the upgrade later on

mpoke:
Closing as this work is tracked by https://github.com/cosmos/gaia/issues/2540

sainoe:
Sovled by #2744

PR #1955, opened 2022-12-12
add back TestGaiaApp_Export 

okwme:
Closes items in #1909  related to
 - TestGaiaApp_Export
 - Investigate the `app/helpers/test_helpers.go` setup function that used to pre-populate with an account and a validator.


app/helpers/test_helpers.go
```diff
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/tendermint/tendermint/crypto"
+
 	// "github.com/tendermint/tendermint/crypto/secp256k1"
```
mmulji-ic: Can we take this out?

okwme: Nice catch. Removed

app/app_test.go
```diff
@@ -5,11 +5,11 @@ import (
 
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	gaia "github.com/cosmos/gaia/v8/app"
```
glnro, line 8: why move these up here?

glnro (approved):
Looks fine so far, in the future we should try to make this as consistent as possible with the [sdk](https://github.com/cosmos/cosmos-sdk/blob/c0d85e30c35921aa0104091f9c45517cd01315ae/simapp/test_helpers.go#L92)...especially when we eventually play #1883

app/helpers/test_helpers.go
```diff
@@ -161,11 +140,7 @@ func setup(withGenesis bool, invCheckPeriod uint) (*gaiaapp.GaiaApp, gaiaapp.Gen
 		encCdc,
 		EmptyAppOptions{},
 	)
-	if withGenesis {
-		return gaiaApp, gaiaapp.NewDefaultGenesisState()
-	}
-
-	return gaiaApp, gaiaapp.GenesisState{}
+	return gaiaApp, gaiaapp.NewDefaultGenesisState()
```
mmulji-ic, line 143: We use NewDefaultGenesisState, so MakeTestEncodingConfig is setup ... what does this give us?

PR #1956, opened 2022-12-12
uncomment still existing test

okwme:
`create-periodic-vesting-account` exists in v0.45.x, `create-permanent-locked-account` does not.


okwme:
apparently `create-periodic-vesting-account` exist in v0.45 but there is no CLI command to create them.
Added a TODO to remind us to add the tests back if/when the CLI is added to v0.45.x

mmulji-ic (approved):

mmulji-ic:
We should bug the SDK folks about consistency.

PR #1938, opened 2022-12-05
feat: Integrate ICS Provider

glnro:
This PR fulfills the minimum requirement #1909 for integrating ICS into Gaia as a [provider chain](https://github.com/cosmos/interchain-security/tree/main/app/provider). 

Outstanding Issues
- [x] Replace Cosmos SDK when ICS changes are merged
- [x] Replace ICS with tagged version when IBC changes are merged (blocked by https://github.com/cosmos/interchain-security/issues/585)
- [x] [Add Gaia Provider test to ICS](https://github.com/cosmos/interchain-security/pull/559)

PR #1959, opened 2022-12-13
chore: remove unneeded replace tags

tac0turtle:
was playing around and saw these, I don't see the point to them currently

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -284,15 +284,12 @@ require (
 replace (
 	// dragonberry
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
-	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.11
 
 	// force all dependecies to use the same versions of ibc, tendermint and cosmos-sdk
 	github.com/cosmos/ibc-go/v3 => github.com/cosmos/ibc-go/v3 v3.4.0
-	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.11
 
 	// use cosmos style protobufs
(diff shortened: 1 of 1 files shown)


okwme:
@tac0turtle these replace tags enforce the same version of tendermint, ibc and sdk for all dependencies (liquidity module and packet forward middleware in particular). This way we're not blocked on them updating their go.mod and publishing a new version or working from a fork.

tac0turtle:
you at least don't need two sdk replaces lol, also are you sure you need them? have you run into issues or something?

okwme:
oh shit i think the double SDK snuck in when i was discussing the same topic on a recent PR that removed them too. thanks for pointing that out!

there should be no real issues if the underlying application uses a patch release version but why take the chance? ideally we eventually get away from redundant dependencies on modules maintained my third parties.

glnro:
For ICS we will need [these specific replace statements ](https://github.com/cosmos/gaia/pull/1938/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6)until changes are upstreamed

tac0turtle:
I don't think replace tags are needed with how go.mod works. Not sure if there was an issue with other modules when using specific heights though?

okwme (approved):
Will keep for now to ensure consistency. There may not be much danger in having slightly different versions of sub-dependencies but it feels much cleaner to ensure they're all the same.

tac0turtle:
> There may not be much danger in having slightly different versions of sub-dependencies but it feels much cleaner to ensure they're all the same.

no this is not how go dependencies work. they grab the latest compatible version. so the parent influences what the dependencies import as well. Would recommend reading up on go mods here https://go.dev/ref/mod

PR #1961, opened 2022-12-14
feat: Change MaxBypassMinFeeMsgGasUsage via NewFeeDecorator

reecepbcups:
This PR allows an importer of the x/globalfee module to change the MaxGas to bypass the minimum fee requirement for select messages.

We need this for Juno so we don't have to fork ante to change this 1 variable.


okwme (approved):
Looks good @Reecepbcups

PR #1963, opened 2022-12-15
upgrade test with (almost) mainnet genesis

okwme:


okwme:
today's my last day so leaving info on where this PR stands.
It's very close but I wasn't able to complete it before heading out.
i can get this to run locally by following the instructions from: https://github.com/cosmos/gaia/blob/main/docs/guidelines/code-guidelines.md#upgrade-test

the only difference is you use `./contrib/scripts/run-gaia-v7-mainnet.sh` instead of `./contrib/scripts/run-gaia-v7.sh` and the parameters are different for the other two commands.
They're in the .github file but listed below as well:

```
          ./contrib/scripts/run-upgrade-commands.sh 12010100 1500
```
and
```
          ./contrib/scripts/test_upgrade.sh 50 5 12010101 localhost
```

the thing is you have to wait ~25 minutes for the command from the `run-gaia-v7-mainnet.sh` to start producing blocks. and only then should the other two commands be run.

probably there should be a new action that just checks for blocks getting produced before proceeding to the final two steps.

PR #1974, opened 2022-12-27
Bring v8 release candidate up to date with main

glnro:

PR #1976, opened 2022-12-30
Migrations for quicksilver issue

stana-miric:
Migrations for stuck funds on quicksilver are added. These changes contain:
1. Sending funds from address where the funds are stuck to the refund address
2. Closing channels related to the incompletely registered quicksilver ica accounts 

@glnro @jtremback This PR should be merged before [this PR](https://github.com/cosmos/gaia/pull/1974) is merged into v.8.0.x release


app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +52,65 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+		return errors.New("invalid destination address")
+	}
+
+	// Get balance from stuck address and subtract 1 uatom sent by bad actor
+	sourceBalance := keepers.BankKeeper.GetBalance(ctx, sourceAddress, "uatom")
+	refundBalance := sourceBalance.SubAmount(sdk.NewInt(1))
+	keepers.BankKeeper.SendCoins(ctx, sourceAddress, destinationAddress, sdk.NewCoins(refundBalance))
+
+	// Get connection to quicksilver chain
+	connectionId, err := getConnectionIdForChainId(ctx, keepers, "quicksilver-1")
```
crodriguezvega: Instead of looping through all connections, would it be possible to just use directly the connection ID that we're interested in? [Is it `connection-707`](https://rest.cosmos.directory/cosmoshub/ibc/core/channel/v1/connections/connection-707/channels)?

app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +52,65 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Close channels
+	closeChannel(keepers, ctx, connectionId, "icacontroller-cosmoshub-4.deposit")
+	closeChannel(keepers, ctx, connectionId, "icacontroller-cosmoshub-4.withdrawal")
+	closeChannel(keepers, ctx, connectionId, "icacontroller-cosmoshub-4.performance")
+	closeChannel(keepers, ctx, connectionId, "icacontroller-cosmoshub-4.delegate")
```
crodriguezvega: Maybe similar comment as with the connection ID. If we know for sure that the channels that we want to close have channel IDs `channel-462`, `channel-463`, `channel-464` and `channel-465` then we could use those IDs directly instead. I see that [all those channels are in connection `connection-707`](https://rest.cosmos.directory/cosmoshub/ibc/core/channel/v1/connections/connection-707/channels), are `OPEN` and have the counterparty port IDs mentioned here (there's also a `channel-466` with counterparty port ID `icacontroller-cosmoshub-4.delegate` in state `TRYOPEN`, but I think that would not be picked up by the call to `GetActiveChannelID`).

app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +52,65 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+
+	return nil
+}
+
+func closeChannel(keepers *keepers.AppKeepers, ctx sdk.Context, connectionId string, port string) {
+	activeChannelId, found := keepers.ICAHostKeeper.GetActiveChannelID(ctx, connectionId, port)
+	if found {
+		channel, found := keepers.IBCKeeper.ChannelKeeper.GetChannel(ctx, icatypes.PortID, activeChannelId)
```
crodriguezvega: Maybe an extra check you could add here is to make sure that the channel is indeed in state `OPEN`.

app/upgrades/v8/upgrades.go
crodriguezvega: Another question: why do the channels with counterparty port IDs `icacontroller-cosmoshub-4.deposit`, `icacontroller-cosmoshub-4.withdrawal` and `icacontroller-cosmoshub-4.performance` also need to be closed if their ICA acccounts on the Hub were not affected?

crodriguezvega (commented):
Thank you! 👍 I left a few comments for now.

app/upgrades/v8/upgrades.go
joe-bowman: We want to return to a state where Quicksilver can re-register the zone from scratch (i.e. channels closed on both ends) so return to the desired state.

app/upgrades/v8/upgrades.go
joe-bowman: What happens here with `icacontroller-cosmoshub-4.delegate` - it is currently in `TRYOPEN`. Presumably it won't be `CLOSED` by this. Do we need it to be to be able to reopen @crodriguezvega ?

app/upgrades/v8/upgrades.go
joe-bowman: Yes, it is `connection-707`.

app/upgrades/v8/upgrades.go
joe-bowman: Oh - we have both 465 and 466 that are both `icacontroller-cosmoshub-4.delegate`. I guess this is okay?

crodriguezvega: Yes, I think it is fine the way it is because [`channel-465` is `OPEN` state](https://rest.cosmos.directory/cosmoshub/ibc/core/channel/v1/channels/channel-465/ports/icahost) and [`channel-466` is in `TRYOPEN` state](https://rest.cosmos.directory/cosmoshub/ibc/core/channel/v1/channels/channel-466/ports/icahost), but 465 is the one that got stored as active channel in state because that's the channel that completed the handshake (storing the active channel happens in the [`OnChanOpenConfirm` callback](https://github.com/cosmos/ibc-go/blob/v3.0.0/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L102)). Related to the [comment above](https://github.com/cosmos/gaia/pull/1976#discussion_r1059655284), maybe we could use the specific channel IDs that we want to close, to be extra sure.

app/upgrades/v8/upgrades.go
stana-miric: We, can use the hard-coded connection id. I'll change it to that one.

app/upgrades/v8/upgrades.go
stana-miric: We'll change the code to use the specific channel IDs, instead of retrieving the channel with GetActiveChannelID

app/upgrades/v8/upgrades.go
stana-miric: Should we also close channel-466?

glnro: @AdityaSripal

mpoke (approved):
Great work @stana-ethernal

app/upgrades/v8/upgrades.go
yaruwangway: Hi @damiannolan, if you or someone in ibc team can do a final review of this ? We plan to include it in the next release if possible, this week. 

Thank you so much !

damiannolan:
Hey, been trying to catch up on the context of the migrations needed here. Thanks for handling this @stana-ethernal, nice work so far! 

Is the idea that quicksilver will reopen the interchain accounts channels with the same port IDs? 
i.e. the following:

```
icacontroller-cosmoshub-4.deposit
icacontroller-cosmoshub-4.withdrawal
icacontroller-cosmoshub-4.performance
icacontroller-cosmoshub-4.delegate
```

Closing the channels is fine, and it will be possible to reopen new ones using the port IDs above. However, I believe that during the handshake the host chain (cosmoshub-4) will still have an interchain account mapping set in the ICA keeper state for the (controller port / host connection) key. Thus, when attempting to reopen new channels you will simply have the same address as used previously. I imagine the desired outcome is to generate a new unique address with ibc-go v3.4.0.

This will occur on `OnChanOpenTry` [here](https://github.com/cosmos/ibc-go/blob/v3.4.0/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L73-L81). In its current form `found` will be `true`, resulting in the same address. I think the desired behaviour is to have `found` return `false` such that a new unique address is created in [`createInterchainAccount`](https://github.com/cosmos/ibc-go/blob/v3.4.0/modules/apps/27-interchain-accounts/host/keeper/account.go#L35). To accomplish this I think the migrations should delete the interchain account mapping in the ICA host keeper. 

For example:
```go
store := ctx.KVStore(k.storeKey) // store key is icahosttypes.StoreKey
store.Delete(icatypes.KeyOwnerAccount(portID, connectionID))
```

cc. @AdityaSripal

stana-miric:
> Hey, been trying to catch up on the context of the migrations needed here. Thanks for handling this @stana-ethernal, nice work so far!
> 
> Is the idea that quicksilver will reopen the interchain accounts channels with the same port IDs? i.e. the following:
> 
> ```
> icacontroller-cosmoshub-4.deposit
> icacontroller-cosmoshub-4.withdrawal
> icacontroller-cosmoshub-4.performance
> icacontroller-cosmoshub-4.delegate
> ```
> 
> Closing the channels is fine, and it will be possible to reopen new ones using the port IDs above. However, I believe that during the handshake the host chain (cosmoshub-4) will still have an interchain account mapping set in the ICA keeper state for the (controller port / host connection) key. Thus, when attempting to reopen new channels you will simply have the same address as used previously. I imagine the desired outcome is to generate a new unique address with ibc-go v3.4.0.
> 
> This will occur on `OnChanOpenTry` [here](https://github.com/cosmos/ibc-go/blob/v3.4.0/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L73-L81). In its current form `found` will be `true`, resulting in the same address. I think the desired behaviour is to have `found` return `false` such that a new unique address is created in [`createInterchainAccount`](https://github.com/cosmos/ibc-go/blob/v3.4.0/modules/apps/27-interchain-accounts/host/keeper/account.go#L35). To accomplish this I think the migrations should delete the interchain account mapping in the ICA host keeper.
> 
> For example:
> 
> ```go
> store := ctx.KVStore(k.storeKey) // store key is icahosttypes.StoreKey
> store.Delete(icatypes.KeyOwnerAccount(portID, connectionID))
> ```
> 
> cc. @AdityaSripal

@damiannolan yes that's right, the ica addresses which are previously successfully registered will remain the same. We've noticed that during testing and I've mentioned this in the slack group related to this problem but we are not sure if those predictable addresses can cause some security risks now when they are already registered. If that is the case, we should add deletion of those addresses to the migration script as mentioned above.

damiannolan:
> @damiannolan yes that's right, the ica addresses which are previously successfully registered will remain the same. We've noticed that during testing and I've mentioned this in the slack group related to this problem but we are not sure if those predictable addresses can cause some security risks now when they are already registered. If that is the case, we should add deletion of those addresses to the migration script as mentioned above.

Okay, I believe the already generated addresses are fine.

To clear up my previous point re. the store deletion for (controller port / host conn) -> ica address. Thanks for the responses on slack which cleared this up. The v3.0.0 code no-ops [here](https://github.com/cosmos/ibc-go/blob/v3.0.0/modules/apps/27-interchain-accounts/host/keeper/account.go#L15) in the event that the account exists and as a result the state entry isn't made in the ICA keeper below on line 26. 

When the handshake takes place post-migrations with v3.4.0 the code which I previously pointed to [here](https://github.com/cosmos/ibc-go/blob/v3.4.0/modules/apps/27-interchain-accounts/host/keeper/handshake.go#L73-L81) will follow the execution path to create a new unique account address.

app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +51,43 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+	if err != nil {
+		return errors.New("invalid source address")
+	}
+	destinationAddress, err := sdk.AccAddressFromBech32("cosmos1jc24kwznud9m3mwqmcz3xw33ndjuufnghstaag")
+	if err != nil {
+		return errors.New("invalid destination address")
+	}
+
+	// Get balance from stuck address and subtract 1 uatom sent by bad actor
+	sourceBalance := keepers.BankKeeper.GetBalance(ctx, sourceAddress, "uatom")
```
damiannolan, line 66: is there a constant for `"uatom"` that could be used?

app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +51,43 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+	if err != nil {
+		return errors.New("unable to refund coins")
+	}
+
+	// Close channels
+	closeChannel(keepers, ctx, "channel-462")
+	closeChannel(keepers, ctx, "channel-463")
+	closeChannel(keepers, ctx, "channel-464")
+	closeChannel(keepers, ctx, "channel-465")
+	closeChannel(keepers, ctx, "channel-466")
```
damiannolan, line 80: I guess you could define the channels in a static array and loop here... but not required.

app/upgrades/v8/upgrades.go
```diff
@@ -69,6 +108,11 @@ func CreateUpgradeHandler(
 			return vm, err
 		}
 
+		err = QuicksilverFix(ctx, keepers)
```
damiannolan, line 113: should this be in its a separate (new) upgrade handler or is it that this PR is targeted towards a particular branch, and this upgrade handler hasn't actually been exec'd yet?

damiannolan (approved):
Thanks @stana-ethernal! 

I just left some minor nits, but nothing major. Feel free to ignore if codeowners here are happy to proceed.

app/upgrades/v8/upgrades.go
glnro: Not outside of testing...we should consider abstracting all instances out in a future refactor.

app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +51,43 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+		return errors.New("invalid source address")
+	}
+	destinationAddress, err := sdk.AccAddressFromBech32("cosmos1jc24kwznud9m3mwqmcz3xw33ndjuufnghstaag")
+	if err != nil {
+		return errors.New("invalid destination address")
+	}
+
+	// Get balance from stuck address and subtract 1 uatom sent by bad actor
+	sourceBalance := keepers.BankKeeper.GetBalance(ctx, sourceAddress, "uatom")
+	refundBalance := sourceBalance.SubAmount(sdk.NewInt(1))
```
glnro: In the [`TestFixBankMetadata` upgrade test](https://github.com/cosmos/gaia/actions/runs/3839514572/jobs/6558146198#step:6:27), there's no error handling for subtracting a zero `sourceBalance`.

glnro: https://github.com/cosmos/gaia/pull/1985 addresses the failing test...we can write a separate test for the whole upgrade.

yaruwangway: this is because the source account does not exist in our test genesis, and also no balance

glnro (approved):

app/upgrades/v8/upgrades.go
```diff
@@ -49,6 +51,45 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
...
+	}
+
+	// Get balance from stuck address and subtract 1 uatom sent by bad actor
+	sourceBalance := keepers.BankKeeper.GetBalance(ctx, sourceAddress, "uatom")
+	if sourceBalance.IsGTE(sdk.NewCoin("uatom", sdk.NewInt(1))) {
+		refundBalance := sourceBalance.SubAmount(sdk.NewInt(1))
+		err = keepers.BankKeeper.SendCoins(ctx, sourceAddress, destinationAddress, sdk.NewCoins(refundBalance))
+		if err != nil {
+			return errors.New("unable to refund coins")
+		}
```
yaruwangway, line 72: @stana-ethernal , since we might skip `sendCoins` if source does not have balance. maybe we can log here that the fund is sent back , so that we can make sure in mainnet and theta-testnet, we have this step executed

PR #2002, opened 2023-01-09
update logs in upgrade v8

yaruwangway:


app/upgrades/v8/upgrades.go
```diff
@@ -40,7 +42,7 @@ func FixBankMetadata(ctx sdk.Context, keepers *keepers.AppKeepers) error {
 	// proceed with the original intention of populating the missing Name and Symbol fields
 	atomMetaData, foundCorrect := keepers.BankKeeper.GetDenomMetaData(ctx, correctDenom)
 	if !foundCorrect {
-		return errors.New("atom denom not found")
+		return errors.New(fmt.Sprintf("denom %s not found", correctDenom))
```
glnro, line 45:
```suggestion
		fmt.Errorf("denom %s not found", correctDenom)
```

glnro: We probably also want to handle this differently at a higher level than the current behavior

glnro:
@yaruwangway going to close this as it's been updated in #2008

PR #2008, opened 2023-01-11
fix: correct ica wiring and update pfm fixes for v8 upgrade

yaruwangway:
During the course of simulated upgrade testing with exported genesis from mainnet it was discovered that the node would panic after upgrading when attempting to prune two different stores, icacontroller and packetforwardmiddleware (PFM).

**Update**:  The fix for ica controller is to remove this as it was accidentally leftover as an artifact from the original v8 release with `v0.46` of the sdk. For PFM, the fix includes a corrected wiring in Gaia, and several updates to PFM itself due to an incorrectly named module and incorrect implementation of an interface necessary to pass to the TransferKeeper.

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -91,7 +91,7 @@ type AppKeepers struct {
 	AuthzKeeper     authzkeeper.Keeper
 	LiquidityKeeper liquiditykeeper.Keeper
 
-	RouterKeeper *routerkeeper.Keeper
+	RouterKeeper routerkeeper.Keeper
 
 	// Modules
 	ICAModule      ica.AppModule
@@ -272,11 +272,12 @@ func NewAppKeeper(
 		govRouter,
 	)
 
+	// TODO: Confirm TransferKeeper wiring correct after strangelove backports https://github.com/strangelove-ventures/packet-forward-middleware/blob/6867bcaac00fef1267c4ea99c3cbb6ee8bc6a025/router/keeper/keeper.go#L49
 	appKeepers.TransferKeeper = ibctransferkeeper.NewKeeper(
 		appCodec,
 		appKeepers.keys[ibctransfertypes.StoreKey],
 		appKeepers.GetSubspace(ibctransfertypes.ModuleName),
-		*appKeepers.RouterKeeper,
+		appKeepers.RouterKeeper,
 		appKeepers.IBCKeeper.ChannelKeeper,
 		&appKeepers.IBCKeeper.PortKeeper,
 		appKeepers.AccountKeeper,
@@ -298,7 +299,7 @@ func NewAppKeeper(
 	appKeepers.ICAModule = ica.NewAppModule(nil, &appKeepers.ICAHostKeeper)
 	icaHostIBCModule := icahost.NewIBCModule(appKeepers.ICAHostKeeper)
 
-	*appKeepers.RouterKeeper = routerkeeper.NewKeeper(
+	appKeepers.RouterKeeper = routerkeeper.NewKeeper(
 		appCodec, appKeepers.keys[routertypes.StoreKey],
 		appKeepers.GetSubspace(routertypes.ModuleName),
 		appKeepers.TransferKeeper,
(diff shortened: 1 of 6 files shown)


glnro (approved):

damiannolan:
Hey @yaruwangway, @glnro  as part of this PR or another can you remove the `ICAControllerKeeper` from `AppKeepers` as well as the store key in `keepers/keys.go`. 

I was going to make one myself but I don't have push access rights on this repo it seems. 
I think it would be worth retesting the upgrade scenario you were having issues with after making these changes. I recall there was errors regarding the `icacontroller` store during some store pruning. I'm not sure at what point you were trying to do pruning but the `icacontroller` store wouldn't have previously existed so may explain why its erroring trying to delete previous iavl versions. The error you're seeing is bubbling up from [here](https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L622-L629), right?

glnro:
> Hey @yaruwangway, @glnro as part of this PR or another can you remove the `ICAControllerKeeper` from `AppKeepers` as well as the store key in `keepers/keys.go`.
> 
> I was going to make one myself but I don't have push access rights on this repo it seems. I think it would be worth retesting the upgrade scenario you were having issues with after making these changes. I recall there was errors regarding the `icacontroller` store during some store pruning. I'm not sure at what point you were trying to do pruning but the `icacontroller` store wouldn't have previously existed so may explain why its erroring trying to delete previous iavl versions. The error you're seeing is bubbling up from [here](https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L622-L629), right?

Yes, that's where the error is bubbling up. We're running an upgrade test with this [test branch](https://github.com/cosmos/gaia/commit/467c8e3ab93e87cc5bf18f44661a368ee31a70e9) and then will assess.

glnro:
> > Hey @yaruwangway, @glnro as part of this PR or another can you remove the `ICAControllerKeeper` from `AppKeepers` as well as the store key in `keepers/keys.go`.
> > I was going to make one myself but I don't have push access rights on this repo it seems. I think it would be worth retesting the upgrade scenario you were having issues with after making these changes. I recall there was errors regarding the `icacontroller` store during some store pruning. I'm not sure at what point you were trying to do pruning but the `icacontroller` store wouldn't have previously existed so may explain why its erroring trying to delete previous iavl versions. The error you're seeing is bubbling up from [here](https://github.com/cosmos/cosmos-sdk/blob/main/store/rootmulti/store.go#L622-L629), right?
> 
> Yes, that's where the error is bubbling up. We're running an upgrade test with this [test branch](https://github.com/cosmos/gaia/commit/467c8e3ab93e87cc5bf18f44661a368ee31a70e9) and then will assess.

@damiannolan the fix of removing the `ICACOntrollerKeeper` as well as the store key seems to have worked in the stateful upgrade test we ran. Now the sdk panics when it tries to prune the `packetforwardmiddleware` store. Not ideal but still at least we can close this up.

faddat:
interesting!

faddat:
hey, FYI I've PR'd into this branch:

https://github.com/cosmos/gaia/pull/2012

Basically just handles the ibc-go v4 jump sooner.

app/keepers/keepers.go
```diff
@@ -277,26 +272,19 @@ func NewAppKeeper(
 		govRouter,
 	)
 
+	// TODO: Confirm TransferKeeper wiring correct after strangelove backports https://github.com/strangelove-ventures/packet-forward-middleware/blob/6867bcaac00fef1267c4ea99c3cbb6ee8bc6a025/router/keeper/keeper.go#L49
 	appKeepers.TransferKeeper = ibctransferkeeper.NewKeeper(
 		appCodec,
 		appKeepers.keys[ibctransfertypes.StoreKey],
 		appKeepers.GetSubspace(ibctransfertypes.ModuleName),
-		appKeepers.IBCKeeper.ChannelKeeper,
+		appKeepers.RouterKeeper,
```
damiannolan: This looks like its not going to work as expected because the `RouterKeeper` is initialised below on line 314. 
At this point it looks like we're going to be passing an empty struct.

yaruwangway: yeah, TransferKeeper need routerKeeper, routerKeeper need transferKeeper to initialize, one must be empty

glnro: @damiannolan Strangelove have added [a setter on `RouterKeeper` for the `TransferKeeper`.](https://github.com/strangelove-ventures/packet-forward-middleware/commit/4fb516ae522e1a36376ca856b82d9d6e75284213)

glnro:
@Mergifyio backport release/v8.0.x

PR #2012, opened 2023-01-12
ibc v4 for v8

faddat:
I think it might make a lot of sense to jump to ibc v4 in v8.

This way, we're using "fresher" upstream code.

Note: this may require an ibc-go migration / upgrade handler, but I'm headed to bed now :)


glnro:
Hey @faddat this would be great if v8 was happening in isolation, but the catch is it's going to be followed immediately by v9 with ICS which is still dependent on v3.4.0... tagging @mpoke @jtremback for input.

faddat:
Ah so I think that @yaruwangway is working on a v9 with v4.2.0

My thinking is that this will get an upgrade out of the way in V8, so we can have fewer changes in v9

faddat:
I'll also be having some pokes at ccv this week, and looking into consumer chains.  I'd like to use 47 when prototyoing them... Will see what I see :).

Probably next week though.

glnro:
@faddat closing this just to consolidate as there is already a spike to look at 4.2 in V9 though the likely scenario is 4.2 in v10. V8 will launch with 3.4

PR #2013, opened 2023-01-12
fix: jwt, cosmos keyring, and gin

faddat:
today while I was working on Juno, the gin library that the hub is using 
flagged a critical security vulnerability.

Looking upstream, I found that these were replaced in the SDK's main 
branch.

Doubt it's the super-critical-critical, but figure better safe than sorry.


glnro:
@julienrbrt  will these be backported to v45?

julienrbrt:
> @julienrbrt  will these be backported to v45?

I think v0.45 have these replace already. The versions probably need to be bumped, but we are not using vulnerable versions.
We'll release a point version soon, so I can have a look.

go.mod
```diff
@@ -282,12 +282,21 @@ require (
...
+	// Use cosmos keyring
+	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
+
 	// dragonberry
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
 
 	// enforce same SDK, Tendermint and IBC on all dependencies
 	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.45.11
 	github.com/cosmos/ibc-go/v3 => github.com/cosmos/ibc-go/v3 v3.4.0
+	// dgrijalva/jwt-go is deprecated and doesn't receive security updates.
```
mmulji-ic: Do we still need these comments?

glnro (approved):
@faddat we're trying to cut a release tonight so might just push this with signed commits unless you're available to amend 🤞

go.mod
glnro:
```suggestion
```

faddat:
Ok, do you want signed commits?

I am generally against that practice.

faddat:
This this has been addressed.

PR #2016, opened 2023-01-13
chore: use go1.19 to prevent mixed runtimes

faddat:
The alternative to this is to have the Makefile check the Go version, as 
Osmosis did for their v13 release.

This is the lower-effort, safer way to address the issue.  Mixing the go 
1.18 and 1.19 runtimes has been known to cause issues with state sync and 
consensus.  Choosing the higher version ensures that no one can use the 
lower version.

closes: 2015


glnro:
Neither ibc-go or sdk versions that v8 will use are opting for 1.19. The [docs explicitly mention which version of go is compatible with gaia](https://hub.cosmos.network/main/getting-started/installation.html#install-go). Perhaps best to open an issue for future consideration instead.

faddat:
Then I implore that a check be added to the Makefile for the go version.

Normally this does not matter.  Between 18 and 19, it matters.

faddat:
Hey I would just like to say once again, that 1.19 will build releases that specify 1.18. this leads to a mixed runtime situation. When there are mixed run times, state sync does not work, and there are rumors that this can break consensus that are being explored in the osmosis validators discord.

The way to entirely prevent this issue, is to specify 1.19 in the go.mod file of version 8 of the cosmos hub. Even adding a check to the makefile like cosmosis did, can still have the result of validators building with the wrong runtime, like:

go install ./...

I know this from experience.

glnro:
Would you be able add the check to the makefile then instead in this pr? @jtremback

jtremback:
@faddat here's how I understand this issue: 

- There are potentially problems related to having some validators compile with 1.18 and some compile with 1.19
- By specifying 1.19 in the go.mod, nobody will be able to compile with 1.18
- But we don't to specify 1.19 because we don't want to do a version upgrade right now because we are trying to freeze the code and get a release out, and our dependencies use 1.18
- So we can add a check in the makefile to refuse to compile with 1.19, solving the potential mixed runtimes problem.

Is that correct?

faddat:
Yeah, but it's way better to just use 1.19, otherwise there can be a bunch of issues.

faddat:
The commentary around state sync here:

* https://github.com/cosmos/gaia/issues/2073
* 
could be solved by this.

PR #2077, opened 2023-01-25
v8.0.x cleanups

faddat:
This PR:

* updates to sdk v0.45.12
* adds replaces from cosmos-sdk repository
* adds the fast node flag, default is fastnodedisable=true, same as sdk upstream


mpoke:
Note that SDK 0.45.12 still uses go 1.18. Thus, we should not update to go 1.19.

faddat:
@mpoke If we use 1.18 then it is possible for validators to have mixed binaries.

If we use 1.19 then it is impossible for validators to have mixed binaries. We studied this at osmosis. 1.19 good. 1.18 good. Mixed bad.

Thus 1.19

faddat:
@mpoke I've reverted to 1.18 -- please know that I do not recommend this, and urge that we use 1.19.


Note: cosmos-sdk team recommends taking measures to ensure that we use only one version of go across the validator set:

* https://github.com/cosmos/cosmos-sdk/issues/13976

When we did this at Osmosis, we used Makefile + 1.18 -- however, after testing, and after a few validators having 1.19 related explosions because they install the chain in the logical manner, eg:

```bash
go install ./...
```

the go version was bumped to 1.19 because otherwise (present continuous verb) validators like myself will run into problems.

There is no need to lower our go runtime version to match the SDK's, life is great with the higher one.  The only thing we need to do is be at 1.18+.  And... 1.18 puts validators into pickles.

go.mod
```diff
@@ -289,16 +288,19 @@ replace (
...
 	// TODO: remove it: https://github.com/cosmos/cosmos-sdk/issues/13134
 	github.com/dgrijalva/jwt-go => github.com/golang-jwt/jwt/v4 v4.4.2
+
 	// TODO Remove it: https://github.com/cosmos/cosmos-sdk/issues/10409
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.8.1
 
 	// use cosmos style protobufs
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+
+	github.com/jhump/protoreflect => github.com/jhump/protoreflect v1.9.0
```
yaruwangway, line 321: Hi @faddat , where this is used ? and why we want v1.9.0 ?  Thanks!

faddat: I slavishly follow upstream recommendations, and that's recommended in the sdk.

faddat:
Will return to this after we've got 47 in the main branch, but closing for now.

Issue #2078, opened 2023-01-25
State sync update


yaruwangway:
## Summary
Regarding to hub state-sync, the followings need to be fixed:
- [ ] The state sync rpc server is unavailable. https://rpc.cosmos.network/
- [ ] Find a reliable source to curl live peers, ref: https://polkachu.com/live_peers/agoric
- [ ] Update the [tutorial](https://hub.cosmos.network/main/getting-started/quickstart.html)

## Proposed Solution

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

faddat:
hey :)

I wanna suggest a whole lot that we do not attempt to fix this until v8.0.x is out and then we never again follow the policy of simply not updating existing releases. 

This is ludicrously difficult to fix before v8 is running on mainnet, and I suggest that we devote time elsewhere

faddat:
* https://github.com/cosmos/gaia/pull/2079

Next is update state sync script then will take out of draft

faddat:
The PR is now ready for review. 

I made an omnibus update to v7.0.x that tidies all the things.

State sync from my laptop succeeds.

faddat:
https://github.com/cosmos/gaia/pull/2079

yaruwangway:
present tutorial was updated and tested working

Issue #1450, opened 2022-04-25
Speed Runs


faddat:
## Summary

Archive nodes are hard to maintain and hard to create.  We should have a system in place testing our release branches sequentially, and timing it.

## Problem Definition

* It takes a very long time to sync from v4-7
* In order to approach this problem, we need to test release branches sequentially and get feedback on how long that takes
* github actions is not the right tool for the job

## Proposal

Create a gitlab CI server for this purpose

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

mmulji-ic:
@faddat Interesting to do, but will check back nearer the end of the qtr once we've cleared Rho.

mpoke:
@faddat is this issue still relevant? If yes, could you please provide more context?

mpoke:
Closing this issue as outdated. If it still relevant, please reopen and provide more context.

PR #2079, opened 2023-01-25
v7.0.x-final

faddat:
This PR updates gaia v7.0.x.  It is an omnibus. 

Closes:

* #2078 

Changes:
* use informal's tendermint fork and bump to v0.34.26
* bump sdk to v0.45.12
* bump ibc (not state breaking) to v3.0.2
* fix state sync script 
* don't use a t-word docker image registry
* markdownlint
* fix golangci-lint 
* use iavl fast node by default 

These are combined here, because this is to be a final cut of release/v7.0.x

I do wish to politely request that we begin to maintain all pieces of cosmoshub-4, beginning with v4.2.x, once we have v8 shipping

To test state sync:

bash contrib/statesync.bash

test-liveness is expected to fail until this is merged.  At that time, we can add to readme.md, the invocation:

```bash
bash contrib/statesync.bash
```

and we shall witness its passage. 

This PR should sync from the first block of v7 to the last, to ensure that it is good.


faddat:
@yaruwangway this is now complete, could you have a look? 

I can confirm:

* state sync works
* we use iavl fast node
* state sync is dramatically faster because of fast node

contrib/testnets/local/gaiadnode/Dockerfile
```diff
@@ -7,7 +7,7 @@
 # Changed from Alpine to Ubuntu because the keyring PR is linking to libc
 # Alpine uses muslc instead of libc
 
-FROM ubuntu:18.04
+FROM ubuntu:jammy
```
yaruwangway, line 10: Hi why use this one ?

cmd/gaiad/cmd/root.go
```diff
@@ -82,6 +82,7 @@ func initAppConfig() (string, interface{}) {
 	srvCfg := serverconfig.DefaultConfig()
 	srvCfg.StateSync.SnapshotInterval = 1000
 	srvCfg.StateSync.SnapshotKeepRecent = 10
+	srvCfg.BaseConfig.IAVLDisableFastNode = false // enable fastnode by default
```
yaruwangway: Hi, @faddat, can you describe a bit more about the  `IAVLDisableFastNode ` ? what is the benefit and concerns if we set it `true` or `false`.  Is there some docs i can check? Thanks!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,31 @@
+name: ci
```
yaruwangway: maybe call it "Publish Docker image"

.github/workflows/docker.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches:
+      - "main"
```
yaruwangway: this action will be triggered on every of push to main ?

yaruwangway:
Hi @faddat, I am not sure this will be included in v7, depends if we still release v7 line, also sdk0.45.12 might need a [coordinate upgrade](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.12). 
 but this PR has some changes which can improve v8, we can include it in v8 (the docker, the state-sync docs and other docs improvement) ? @mmulji-ic @mpoke @jtremback

contrib/testnets/local/gaiadnode/Dockerfile
faddat: old one is obsolete, this caused glibc issues

.github/workflows/docker.yml
faddat: great catch :D

.github/workflows/docker.yml
faddat: You know it shouldn't be 🙂 thank you for your review I will be making the changes now

faddat:
I also think that given the release notes on 45.12, that it probably makes sense to stay with 45.11. what say you?

faddat:
@yaruwangway I need to inform you that the ICA issue with qs happened precisely because we did not maintain previous versions. We were only accepting pull requests at the tip, and this harmed maintenance.

Therefore, I advocate maintaining all versions of Gaia that are necessary to sync Gaia, that would be v4.2.0.x-v8.

All of those should be maintained. Otherwise, verifiability of archive state becomes incredibly difficult.

Essentially, ever since I wrote a certain issue that anthropomorphized the hub, I have wanted to roll through the hub's releases like:

* v4
* v5
* v6
* v7
* v8

And format / optimize each of these.  This would allow us to run speed races, and ensure that the chain syncs as rapidly as possible.

cmd/gaiad/cmd/root.go
faddat: If we set it to false here, then we should set it to false in v8.  I recommend false.

faddat: false means we will use fast node.  I recommend using it.

yaruwangway: `fast node` means [fast sync](https://docs.tendermint.com/v0.34/tendermint-core/fast-sync.html) ? Do you know why the fafault is set to be true ? If we change it, what would be the impact on a node ?

yaruwangway:
Hi @faddat, if you can split this PR into two PRs? one only updates fast node, the rest goes to another PR, and make both PRs against main ?

or you can change your present PR against [this branch](https://github.com/cosmos/gaia/tree/update-docs-workflow) and I can make some changes before it merged ?

Thank you!

mmulji-ic:
> @yaruwangway I need to inform you that the ICA issue with qs happened precisely because we did not maintain previous versions. We were only accepting pull requests at the tip, and this harmed maintenance.
> 
> Therefore, I advocate maintaining all versions of Gaia that are necessary to sync Gaia, that would be v4.2.0.x-v8.
> 
> All of those should be maintained. Otherwise, verifiability of archive state becomes incredibly difficult.
> 
> Essentially, ever since I wrote a certain issue that anthropomorphized the hub, I have wanted to roll through the hub's releases like:

Thanks @faddat for this PR. Currently we're following the [Major Release Maintenence policy](https://github.com/cosmos/gaia/blob/main/RELEASING.md#major-release-maintenance), so the current plan is to stop updates for v6 and lower when Rho is out. We should schedule a discussion if you strongly believe that we should support older versions to understand the impact for node operators.

We will review this, as @yaruwangway suggested splitting this into 2 would be helpful.

As a sidenote, the qs issue was caused by constant upstream delays and also the time squeeze that happened because of ICS. In the future we will target smaller releases and release whats ready rather than wait for an agreed feature set which might be delayed. Hopefully this will help with maintenance and security issues fixed sooner.

faddat:
We should not "freeze" old gaias. 


@yaruwangway I've removed the iavl fast node by default here, I think this PR is good to go. 

The qs issue was caused by a lack of work on the hub, not by upstream delays.  There is no other production chain that I'm aware of using that version of ibc-go.

yaruwangway:
Hi @faddat, how about updating v8 first so we can include this pr to the upcoming releases ? this might need some update of the go.mod in this pr to match v8. 
let me know if you prefer to do it, otherwise, i can merge this pr to a branch i checked out and change the version there. Thank you!

faddat:
@yaruwangway 

Hi, so this is not the kind of thing that we would merge into the tip, because many of these changes are already in the tip.

Thing is, we do not backport them, and they are version-sensitive.  With state sync, for example, we will need to rewrite that after the upgrade. 

In fact I'm interested in going even more backwards.

faddat:
@mmulji-ic 

https://github.com/cosmos/ibc-go/releases

Please count the number of compatible releases that upstream shipped while hub used v3.0.0

I came up with a number that makes your claim impossible, unfortunately. 

I would like to suggest that we try to keep historical gaias up to date with all of our latest standards and practices, while not breaking state.

yaruwangway:
Hi, @faddat , can try to get this pr into v7.1.1, can you update this pr to resolve the conflicts ?

yaruwangway:
a release of v7.1.2 ? @mmulji-ic

faddat:
@yaruwangway 

I'll break out few very old issues:

* https://github.com/cosmos/gaia/issues/1450
* https://github.com/osmosis-labs/osmosis/issues/3518
* https://github.com/cosmos/cosmos-sdk/issues/14426
* https://github.com/osmosis-labs/osmosis/issues/3519

You'll note that I've brought in 2 osmosis issues here, and 1 sdk issue. 

Currently it is kinda nearly impossible to make archive nodes so everyone just uses quicksync.io but now that's the canonical archive state of the hub, eg: possible long range attack or just plain old state corruption.  We also know that we can dramatically improve sync speed.  So the idea is to bundle these, and actually begin from v4.2.0, and work our way upwards.

cosmoshub-4 is:

v4.x
v5.x
v6.x
v7.x
v8.x
v9.x

-- not only the tip.  So we can't really just make PR's into the tip because for reproducibility we need to walk through the different versions.  Ultimate issue: that is simply too time consuming right now so no one does it.

cmd/gaiad/cmd/root.go
faddat: no, it refers to iavl fast node

faddat:
Note the v0.45.12 update -- interestingly, we don't know if it is state breaking or not -- it maybe is, according to its release notes.  So then we would want to walk through prior releases, and do a full sync, so we can know for sure. 

Does it make sense?

glnro:
> Note the v0.45.12 update -- interestingly, we don't know if it is state breaking or not -- it maybe is, according to its release notes. So then we would want to walk through prior releases, and do a full sync, so we can know for sure.
> 
> Does it make sense?

The advice is to bump to `v0.45.12` in a coordinated upgrade, which the hub will be doing with `v8.0.0` https://github.com/cosmos/gaia/releases/tag/v8.0.0

faddat:
@glnro indeed.  I want to test this with the full state of v7 before merging. 

Then we **know**. 

There have been a bunch of cases where we guess, and I figure it's better for us to know for sure.

faddat:
Sorry, since we want to use this, and v0.45.12 would delay that, I've reverted that commit.  Can test that at a later time.

faddat:
@glnro -- I think that we should bump to v0.45.13 in v8

* #2183

mpoke:
Closing as v7.0.x is no longer supported.

Issue #2074, opened 2023-01-24
Rest endpoints appear to have gone away


jtremback:
## Summary of Bug

- Some change in some rest endpoints, for example /blocks/latest, has made them not work
- This would be a big issue for block explorers so we can’t release without addressing this
- The odd thing is that the SDK version bump is a minor bump from 45.9 to 45.11, and was just supposed to add some RS related code to prepare for the release of Gaia v9
- This needs to be summed up and presented to Marko or someone from the SDK team

julienrbrt:
I have tried with simapp and these endpoints do work.

~~The only issue I can think about is https://github.com/cosmos/cosmos-sdk/issues/13695, but it gives another error.
The above issue is due to an issue in the way a chain load its config. This had popped up in a point released because we shadowed a variable (which we've reverted in v0.45.12).~~ EDIT: I don't see an error with your config loading, so I don't think it's that.

ryujimin220:
We cannot provide services due to the same issue.

## Tested Port
- 1317 : not working
- 26657 : good
- 9090 grpc : good

## Environment
- OS : Ubuntu-2004-focal-64-minimal
- Gaia : v8.0.0 RC3
- Go : go1.19.5 linux/amd64

Both our archive and rpc nodes are facing this issue. 

I do agree that this version should have been blocked from releasing.

yaruwangway:
tested with sdk 0.45.12 for v8:

the followings are there:
- block?height=_  
- validators?height=_&page=_&per_page=_

the following are still  missing:
- blocks/latest 
- validatorsets/latest
- node_info/synching

jtremback:
@julienrbrt sorry for assuming this was a Cosmos-SDK thing when it was indeed in our app

PR #2080, opened 2023-01-26
Reintroduce missing code for rest endpoint wiring

mmulji-ic:
Fixes issue https://github.com/cosmos/gaia/issues/2074


mmulji-ic:
Note integration tests are not added at this point.

faddat (approved):
I strongly approve of this. Thank you so much for catching it.

tests/e2e/e2e_rest_interfaces_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// Helper function to read the response body and unmarshal it into a map
+func readJSON(resp *http.Response) (map[string]interface{}, error) {
```
glnro: We have a similar helper here...https://github.com/cosmos/gaia/blob/main/tests/e2e/http_util.go and similar functionality in https://github.com/cosmos/gaia/blob/main/tests/e2e/query.go#L23...maybe there's a way to refactor

tests/e2e/e2e_rest_interfaces_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+		for _, endpoint := range testEndpoints {
+
+			// Call the required endpoint
+			resp, err := http.Get(endpointURL + endpoint.Path)
+			if err != nil {
+				testOk = false
+				s.T().Logf("failed to get endpoint: %s, %v", endpointURL+endpoint.Path, err)
+				continue
+			}
```
glnro:
```suggestion
			s.NoError(err, fmt.Sprintf("Failed to get endpoint: %s%s, %s", endpointURL, endpoint.Path, err.Error()))
```

tests/e2e/e2e_rest_interfaces_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+			// Decode the JSON resopnse
+			jsonBody, errJSON := readJSON(resp)
+			if errJSON != nil {
+				testOk = false
+				s.T().Logf("failed to read body of endpoint: %s, %v", endpointURL+endpoint.Path, errJSON)
+				continue
+			}
```
glnro:
```suggestion
			jsonBody, err := readJSON(resp)
			s.NoError(err, fmt.Sprintf("Failed to read body of endpoint: %s%s, %s", endpointURL, endpoint.Path, err.Error()))
```

tests/e2e/e2e_rest_interfaces_test.go
```diff
@@ -0,0 +1,83 @@
...
+				{"/node_info", false},
+				{"/syncing", false},
+				{"/missing_endpoint", true},
+			}
+		)
+
+		for _, endpoint := range testEndpoints {
+
+			// Call the required endpoint
+			resp, err := http.Get(endpointURL + endpoint.Path)
```
glnro: For failure cases, if the endpoint doesn't exist, there's not really a reason for the final test clause, you might as well just assert `s.EqualValues(resp.StatusCode, 200)`, which will fail as the `missing_endpoint` test case will return a `501`

glnro (approved):

tests/e2e/e2e_rest_interfaces_test.go
```diff
@@ -0,0 +1,83 @@
...
+			c             = s.chainA
+			endpointURL   = fmt.Sprintf("http://%s", s.valResources[c.id][valIdx].GetHostPort("1317/tcp"))
+			testEndpoints = []struct {
+				Path         string
+				ExpectedFail bool
+			}{
+				{"/node_info", false},
+				{"/syncing", false},
+				{"/missing_endpoint", true},
+			}
```
yaruwangway: we can check every endpoint, not limited to these 3, but can do this later

tests/e2e/e2e_rest_interfaces_test.go
```diff
@@ -0,0 +1,83 @@
...
+			jsonBody, errJSON := readJSON(resp)
+			if errJSON != nil {
+				testOk = false
+				s.T().Logf("failed to read body of endpoint: %s, %v", endpointURL+endpoint.Path, errJSON)
+				continue
+			}
+
+			if endpoint.ExpectedFail == false && jsonBody["message"] == "Not Implemented" {
+				testOk = false
+				s.T().Logf("Encountered a Not implemented endpoint: %s", endpointURL+endpoint.Path)
```
yaruwangway: is it should be "Not Implemented""? otherwise, can  just fail here, log might not make people realize if this test fails. maybe can use `s.Require.False(jsonBody["message"] == "Not Implemented")` or use `s.Require.True` to check how the `jsonBody["message"]` should be might be better.

tests/e2e/e2e_rest_interfaces_test.go
mmulji-ic: This is good. I also wanted to check the structure strictly (will leave that for later).

tests/e2e/e2e_rest_interfaces_test.go
mmulji-ic: +1. atm just want to make sure that the module is wired ok.

mpoke (approved):

tests/e2e/e2e_rest_regression_test.go
```diff
@@ -0,0 +1,89 @@
...
+2. Auth params
+3. Distribution for Community Pool
+4. Evidence
+5. Gov proposals
+6. Mint params
+7. Slashing params
+8. Staking params
+*/
+
+var (
```
MSalopek: Nit: maybe consider using `const`.
```suggestion
const (
```

MSalopek (approved):
LGTM!

Left some nits.

sainoe (approved):

tests/e2e/e2e_rest_regression_test.go
mmulji-ic: Fixed nit

Issue #2083, opened 2023-01-26
Update Rho dependencies (SDK & Tendermint)


mmulji-ic:
## Summary

- [x] Bump Cosmos SDK to v0.45.12
- [x] Bump Tendermint to use Informal Systems repo v0.34.24
~~- [ ] Remove unneeded references to Tendermint in Gaia~~

## Type

Code Debt

## Impact

Version conflict / Source stability

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

mmulji-ic:
Fixed in #2088

PR #2088, opened 2023-01-27
chore: bump SDK v0.45.12

mmulji-ic:
Closes issue #2083


go.mod
```diff
@@ -29,6 +28,7 @@ require (
 require (
 	cosmossdk.io/math v1.0.0-beta.4
 	github.com/ory/dockertest/v3 v3.9.1
+	github.com/tendermint/tendermint v0.34.24
```
glnro:
```suggestion
```

go.mod
```diff
@@ -299,6 +295,8 @@ replace (
 
 	// use cosmos style protobufs
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+
+	// Use Informal Systems fork of Tendermint
 	github.com/tendermint/tendermint => github.com/informalsystems/tendermint v0.34.24
```
mpoke, line 303: Do we still need this?

go.mod
```diff
@@ -20,7 +20,6 @@ require (
 	github.com/spf13/viper v1.14.0
 	github.com/strangelove-ventures/packet-forward-middleware/v3 v3.1.1
 	github.com/stretchr/testify v1.8.1
-	github.com/tendermint/tendermint v0.34.24
```
glnro: No need to remove here

glnro (changes requested):

go.mod
mpoke: Why? I though that we don't need to directly import TM.

go.mod
glnro: We need to keep the replace at minimum...gaia doesn't need tendermint directly.

go.mod
glnro: We actually do, other dependencies require it.

go.mod
```diff
@@ -285,11 +284,8 @@ replace (
 	// Use cosmos keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
 
-	// dragonberry
-	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
```
glnro, line 289: Per the [SDK's release notes](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.45.12), it's advised to also bump `github.com/confio/ics23/go v0.9.0`...while I see this has been indirectly required, given the nature of the version bump and the SDK's recommendation we should explicitly require this.

glnro (changes requested):

go.mod
mmulji-ic: This was the go mod action, but have updated this time. 
We shouldn't mangle too much what go mod kicks out.

go.mod
mmulji-ic: SDK says:
# Can be deleted from go.mod
replace github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
Instead, github.com/confio/ics23/go must be bumped to v0.9.0.

Which is what I've done.

go.mod
mmulji-ic: This was the go mod action, but have updated this time. 
We shouldn't mangle too much what go mod kicks out.

go.mod
glnro: It's being required indirectly [here](https://github.com/cosmos/gaia/pull/2088/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R72)...we should explicitly require this version.

glnro (approved):

mpoke (approved):

sainoe (approved):

PR #2092, opened 2023-01-27
change bypass-min-fee-types parsing; change defaults

### LINKED ISSUE #2087: Config params not properly parsed
## Summary of Bug

Configuration params with the `key = [ ]` structure are not properly passed. The expected value should be an empty list, however, the outcome is [a nil value](https://github.com/cosmos/gaia/blob/983a9fe4288d6e474aea7402afc640fcbc67e842/app/app.go#L190).  

In general, we should be able to differentiate between the following scenarios:

- no key in the config file 
  - Log that default value is used instead
- `key =`
  - Log a warning that the config is incorrect. Use default value instead.
- `key = [ ]` 
  - The value for the key is an empty list 
- `key = ["value1", "value2"]`
  - The value for the key is this list. 

## Version

v8, v9

____

#### For Admin Use

- [x] Not duplicate issue

MSalopek:
Closes: #2087 

Parsing of `bypass-min-fee-types` is changed to allow operators to use empty bypass list.

The default was changed to be an empty list (no bypass allowed).

Removing the `bypass-min-fee-types` from `app.toml` will cause the default list of types to be used.

Validation was added so strings not corresponding to any proto message types will cause a panic (since this is a misconfiguration, manual action should be required)

Logging was added to inform operators about bypass usage.

### Behaviour
`no key in the config file`
- use defaults; log that default values were applied
 
`key = `
- node panics; config cannot be parsed

`key = [ ]`
- The value for the key is an empty list 

`key = ["value1", "value2"]`
- the value for the key is this list.


cmd/gaiad/cmd/root.go
```diff
@@ -83,7 +83,7 @@ func initAppConfig() (string, interface{}) {
 
 	return params.CustomConfigTemplate(), params.CustomAppConfig{
 		Config:               *srvCfg,
-		BypassMinFeeMsgTypes: gaia.GetDefaultBypassFeeMessages(),
+		BypassMinFeeMsgTypes: []string{},
```
yaruwangway: I think we still want set default bypass msg when do `gaiad init` ?

yaruwangway:
we can add some e2e tests, let me know if you want me to add it

yaruwangway:
this pr will be refactored, the global fee will finally goes to store, in case some decision changes later, lets still get this merged first.

yaruwangway:
this pr will be finally refactored. please see this issue. 
https://github.com/cosmos/gaia/issues/2228#issuecomment-1458437415

but lets still get it merged first, in case any decision changes later

sainoe (approved):

cmd/gaiad/cmd/root.go
MSalopek: The default with this change means that no bypass messages will be set for you. You would need to change the config yourself.

Do you think it shouls be pre-populated?

yaruwangway: it is pre-populated, and nodes can change it.

MSalopek: Done!

yaruwangway (approved):

yaruwangway:
changelog need to be added before merge.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * 
 * (feat) Add two more msg types `/ibc.core.channel.v1.MsgTimeout` and `/ibc.core.channel.v1.MsgTimeoutOnClose` to default `bypass-min-fee-msg-types`.
 * (feat) Change the bypassing gas usage criteria. Instead of requiring 200,000 gas per `bypass-min-fee-msg`, we will now allow a maximum total usage of 1,000,000 gas for all bypassed messages in a transaction. Note that all messages in the transaction must be the `bypass-min-fee-msg-types` for the bypass min fee to take effect, otherwise, fee payment will still apply.
+* (fix) [#2087](https://github.com/cosmos/gaia/issues/2087) Fix `bypass-min-fee-msg-types` parsing in `app.toml`. Parsing of `bypass-min-fee-types` is changed to allow node operators to use empty bypass list. Removing the `bypass-min-fee-types` from `app.toml` applies the default message types. See [#2092]( https://github.com/cosmos/gaia/pull/2092) for details.
```
sainoe:
```suggestion
* (fix) [#2087](https://github.com/cosmos/gaia/issues/2087) Fix `bypass-min-fee-msg-types` parsing in `app.toml`. Parsing of `bypass-min-fee-types` is changed to allow node operators to use empty bypass list. Removing the `bypass-min-fee-types` from `app.toml` applies the default message types. See [#2092](https://github.com/cosmos/gaia/pull/2092) for details.
```

PR #2095, opened 2023-01-29
feat: Add go version check

glnro:
This PR addresses concerns raised in #2016 by adding a check in the Makefile to ensure the system is using `1.18` to build Gaia. 

AC
```
$ go version
go version go1.19.5 darwin/arm64

$ make build
ERROR: Go version 1.18 is required for glnro/version-warning-983a9fe4288d6e474aea7402afc640fcbc67e842 of Gaia.
exit 1
make: *** [check_version] Error 1

// Downgrade to 1.18

$ go version
go version go1.18.10 darwin/arm64

$ make build
--> Ensure dependencies have not been modified
all modules verified
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,8 @@ DOCKER := $(shell which docker)
 BUILDDIR ?= $(CURDIR)/build
 TEST_DOCKER_REPO=cosmos/contrib-gaiatest
 
-GO_MAJOR_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f1)
-GO_MINOR_VERSION = $(shell go version | cut -c 14- | cut -d' ' -f1 | cut -d'.' -f2)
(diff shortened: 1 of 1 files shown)


mpoke (approved):
Could we pull the required go version from the SDK version that we are using? :)

Makefile
```diff
@@ -89,16 +92,22 @@ endif
 include contrib/devtools/Makefile
 
 ###############################################################################
-###                              Documentation                              ###
+###                              Build                                      ###
 ###############################################################################
 
+check_version:
+ifneq ($(GO_MINOR_VERSION),18)
```
MSalopek: Would there be a problem with checking against `1.18` or even `>= 1.18`?

Also, would it be possible to make the hardcoded `18` const into a var?
```bash
REQUIRE_GO_VERSION = '1.18'
```

MSalopek (approved):

PR #2164, opened 2023-02-09
chore: update CHANGELOG.md

MSalopek:


glnro (approved):

yaruwangway (approved):

yaruwangway:
@glnro , in the past, we do not keep rc release changelog, we can keep it from v9, that means we need also do it for future releases.

PR #2187, opened 2023-02-14
fix: link checking failed

yaruwangway:


mmulji-ic:
@yaruwangway is this not fixed by the mlc_config.json changes?

yaruwangway:
it seems not, the markdown link check failed before this pr still.

Issue #2249, opened 2023-02-24
genesis export goes to stderr rather than stdout


yaruwangway:
## Summary of Bug
`gaiad export` and `gaiad init` export the genesis to stderr rather than stdout.

## Concerns
Users has to export genesis by redirect stderr to stdout: `gaiad export > genesis.json  2>&1 `. 
This fix is api breaking.

## Version
gaiad v8.0.0, gaiad v9.0.0
`simd` from sdk 0.45.10 and sdk 0.45.13-ics do not have the issue with `export` but only `init`

## Steps to Reproduce
`gaiad export  > genesis.json` ,  genesis.json is empty
`gaiad export > genesis.json  2>&1 `, get exported genesis into the file
The same with `gaiad init`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

yaruwangway:
Fix export: https://github.com/cosmos/gaia/pull/2250

yaruwangway:
issue for `binary init` outputs to stderr in cosmos-sdk https://github.com/cosmos/cosmos-sdk/issues/15167

yaruwangway:
`simd` commands output to stderr is expected. https://github.com/cosmos/cosmos-sdk/issues/15167 
nevertheless, `simd export` can output to stdout, if gaia does the same, here is the fix. https://github.com/cosmos/gaia/pull/2250  otherwise, can close this issue.

mmulji-ic:
@yaruwangway could we also have a warning when executing that command that this function has changed? Just to make sure that people using this function get a reminder immediately that something has changed when they use it.

mpoke:
@yaruwangway what's the impact of this issue? Is anybody affected by this?

yaruwangway:
the impact is when export, has to redirect stderr. 
close for now since no users complain about it.

PR #2250, opened 2023-02-24
fix: gaiad export to stdout, not stderr

yaruwangway:
Please check https://github.com/cosmos/gaia/issues/2249 for issue description.
 - [x] fix `gaiad export` to output stdout
 - [x] fix parsing block info from stdout, not stderr 
 - [ ] add to changelog (need to decide which version to include it)


yaruwangway:
Need to decide: which version to include this fix.

cmd/gaiad/cmd/root.go
```diff
@@ -52,6 +52,10 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 		Use:   "gaiad",
 		Short: "Stargate Cosmos Hub App",
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			// set the default command outputs
+			cmd.SetOut(cmd.OutOrStdout())
+			cmd.SetErr(cmd.ErrOrStderr())
+
```
mpoke, line 58: What's the impact of this change on users that already use stderr?

mpoke (changes requested):
Not sure whether we want to fix this in a patch release.

cmd/gaiad/cmd/root.go
yaruwangway: if users use `gaiad export > genesis.json 2>&1` before, it will still work.

mpoke:
Closing due to lack of activity.

PR #2263, opened 2023-03-07
Update validator-faq.md

jiseongnoh:
change the number of validators from 150 to 175


mmulji-ic (approved):
Thanks for catching @jiseongnoh

mpoke (approved):

PR #2264, opened 2023-03-07
Update delegator-faq.md

jiseongnoh:
top 125 to 175 validators


mmulji-ic (approved):
Thanks for catching that @jiseongnoh

mpoke (approved):

PR #2209, opened 2023-02-16
State Sync & t*-db

faddat:
This PR adds a state sync script for gaia v8, and avoids the t* github org, while adding support for pebble-db.

to experience velocity:

```bash
bash contrib/statesync.bash
```

PR #2298, opened 2023-03-16
Update statesync.bash

faddat:
This PR fixes state sync only, as opposed to #2209 which approached other matters.  This way, validators and infrastructure providers will be able to more easily state sync the cosmos hub. 

This PR can be tested by:

```bash
bash contrib/statesync.bash
```


MSalopek (approved):
Thank you for your contribution!

Looks good!

mmulji-ic:
Hi @faddat thanks for the script can you take a look at the comments in https://github.com/cosmos/gaia/pull/2209 , which also apply to this. Once the commit to main is ok, would close https://github.com/cosmos/gaia/pull/2209

mmulji-ic (approved):

PR #2317, opened 2023-03-22
Fix typos and Consumer vs Child chain naming

p-offtermatt:
Sometimes, Consumer Chains were called child chains. I changed it to Consumer Chains.
Also, I fixed a few minor typos.


mpoke (approved):

mpoke:
@mmulji-ic @jtremback is this document still needed?

mmulji-ic:
> @mmulji-ic @jtremback is this document still needed?

Think this was added as there was no other ICS documentation, now there is, so this could be removed. Check any references to this doc and point to the new ICS site.

mmulji-ic (approved):

mpoke:
> > @mmulji-ic @jtremback is this document still needed?
> 
> Think this was added as there was no other ICS documentation, now there is, so this could be removed. Check any references to this doc and point to the new ICS site.

I opened this issue https://github.com/cosmos/gaia/issues/2348 to deal with this. In the meantime, there is no reason to update the doc. As a result, I'll close this PR.

PR #2336, opened 2023-03-24
additional linter configuration

faddat:
This PR adds explicit configuration to the majority of linters that Gaia uses


faddat:
Related to:

* https://github.com/cosmos/cosmos-sdk/issues/15546

But also more broadly, the challenge that faces maintainers.

jtremback:
Does this linter have default settings? Is there a reason not to use the defaults?

faddat:
this isn't needed, thanks

PR #2343, opened 2023-03-28
Update tutorial examples on minting

deniseyu:
I noticed that the `minting` CLI namespace is `mint` in the latest released version (`v9.0.1`). Opening a PR to update the tutorial examples accordingly. Thanks!


mmulji-ic:
Good catch @deniseyu thanks!

mmulji-ic (approved):

mpoke (approved):
Thanks @deniseyu

PR #2356, opened 2023-04-03
chore: update mergify

mpoke:
- remove v7 & v10 backports
- add author as backport assignee

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -42,3 +42,12 @@ pull_request_rules:
       backport:
         branches:
           - release/v9.0.x
+
+  - name: Backport patches to the release/v10.0.x branch
+    conditions:
+       - base=main
+       - label=A:backport/v10.0.x
+     actions:
+       backport:
+         branches:
+           - release/v10.0.x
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -45,12 +42,3 @@ pull_request_rules:
       backport:
         branches:
           - release/v9.0.x
-
-  - name: Backport patches to the release/v10.0.x branch
```
glnro: Why remove 10 if we'll need to add it back soon?

mpoke: Good point. I added it back.

sainoe (approved):

.mergify.yml
```diff
@@ -18,15 +24,6 @@ pull_request_rules:
...
-  - name: Backport patches to the release/v7.0.x branch
-    conditions:
-      - base=main
-      - label=A:backport/v7.0.x
-    actions:
-      backport:
-        branches:
-          - release/v7.0.x
-
   - name: Backport patches to the release/v8.0.x branch
```
mmulji-ic, line 27: How about the following condition for all backports?
```
    conditions:
      - "#approved-reviews-by>=1"
```

mpoke: It's already part of the branch protection rule. That will overwrite whatever you add here.

mmulji-ic (approved):

PR #2377, opened 2023-04-06
chore: Fix makefile linting targets

glnro:
Update Makefile linting targets to be consistent with the Cosmos SDK

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,14 +218,24 @@ docker-build-all: docker-build-debug docker-build-hermes
 ###############################################################################
 ###                                Linting                                  ###
 ###############################################################################
+golangci_lint_cmd=golangci-lint
+golangci_version=v1.50.1
 
 lint:
 	@echo "--> Running linter"
-	@go run github.com/golangci/golangci-lint/cmd/golangci-lint run --timeout=10m
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golangci_version)
+	@$(golangci_lint_cmd) run --timeout=10m
+
+lint-fix:
+	@echo "--> Running linter"
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golangci_version)
+	@$(golangci_lint_cmd) run --fix --out-format=tab --issues-exit-code=0
 
 format:
+	@go install mvdan.cc/gofumpt@latest
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golangci_version)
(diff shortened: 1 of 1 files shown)


Makefile
mpoke, line 1: Is it necessary to update `.golangci.yml`?

Makefile
```diff
@@ -218,14 +218,24 @@ docker-build-all: docker-build-debug docker-build-hermes
 ###############################################################################
 ###                                Linting                                  ###
 ###############################################################################
+golangci_lint_cmd=golangci-lint
+golangci_version=v1.50.1
```
mpoke, line 222: In go.mod we use `github.com/golangci/golangci-lint v1.52.1`. Does it have to match?

sainoe (approved):

Makefile
```diff
@@ -218,14 +218,24 @@ docker-build-all: docker-build-debug docker-build-hermes
...
 
 lint:
 	@echo "--> Running linter"
-	@go run github.com/golangci/golangci-lint/cmd/golangci-lint run --timeout=10m
+	@go install github.com/golangci/golangci-lint/cmd/golangci-lint@$(golangci_version)
+	@$(golangci_lint_cmd) run --timeout=10m
+
+lint-fix:
```
mpoke, line 229: What's the difference between `make lint-fix` and `make format`?

Makefile
glnro: No because it requires go `1.19`. If you bump it on your machine and run `make lint` you'll encounter the following error:
```
--> Running linter
# honnef.co/go/tools/unused
../../../go/pkg/mod/honnef.co/go/tools@v0.4.3/unused/unused.go:419:14: obj.Origin undefined (type *types.Var has no field or method Origin)
../../../go/pkg/mod/honnef.co/go/tools@v0.4.3/unused/unused.go:421:14: obj.Origin undefined (type *types.Func has no field or method Origin)
note: module requires Go 1.19
# honnef.co/go/tools/staticcheck
../../../go/pkg/mod/honnef.co/go/tools@v0.4.3/staticcheck/lint.go:4025:29: righti.Method(i).Origin undefined (type *types.Func has no field or method Origin)
../../../go/pkg/mod/honnef.co/go/tools@v0.4.3/staticcheck/lint.go:4030:36: sel.Obj().(*types.Func).Origin undefined (type *types.Func has no field or method Origin)
note: module requires Go 1.19
make: *** [lint] Error 2
```

glnro: I believe this is why @MSalopek had to update the linter workflow in #2324.

Makefile
glnro: What updates would you want to make to `.golangci.yml` ? This PR is specific to updating the makefile linting targets, not changing anything about golangci configuration.

Makefile
glnro: `make lint-fix` is just running the golanci linter, while `make format` is running `gofumpt` on everything that's not excluded by `-not -path`, and then it runs golanci linter afterwards.

mpoke (approved):

PR #2399, opened 2023-04-13
chore: bump sdk and replace comet

glnro:
Bump to sdk `v0.45.15-ics` and replace tendermint with comet


mmulji-ic (approved):

mmulji-ic:
~~@glnro shouldn't this target `main` first~~ 
main already updated

PR #2472, opened 2023-05-07
enforce the use of go 1.20, making mixed runtimes impossible

faddat:
THis PR makes it impossible to use any version of Go below 1.20 with Gaia. 

Go has a builtin install command like:

```bash
go install ./...
```

many validators use it.  It fully bypasses the go version check in Makefile.


mpoke:
@faddat Thanks for your contribution. This PR was closed automatically when https://github.com/cosmos/gaia/pull/2474 was merged. Since this emergency release was not tested with go 1.20, we decided to not risk it and add it to the upgrade. The next planed Gaia upgrade will bump go to 1.20; `main` already uses go 1.20.

PR #2595, opened 2023-06-15
Use IAVL fast node

faddat:
This PR enables IAVL fast node, and will become more important as more consumer chains are being added.


jtremback:
Does this introduce any risks of consensus breaking or can it be put out as a non-coordinated upgrade?

Can you post some stats about what the performance increase is?

faddat:
The best way is to test it out for yourself, here:

https://twitter.com/gadikian/status/1669600575053508613?s=20

As for stats, I can ask Roman on the osmosis team, but I'm not fully sure.  There may have been some stats, from a long time ago, the thing I can say for sure is that there was a dramatic improvement in iterator performance.  

I'll make a PR that demonstrates using it as seen in the video, against the v9.x branch...

* #2600 

- please note that this has in the past caused a consensus break on osmosis but that was several iavl versions ago.  I think that everything ought to be fine.  You'll notice much better performance from the get go though.

faddat:
This can also help with the set of issues outlined in #2635 

Keep in mind: when a node is pounded to death over RPC, one thing that happens is that p2p goes down.

faddat:
@jtremback -- the benchmarks are... not great or formal.  

To my knowledge, no one has them.  Another reason that this PR is a good idea though, is the fact that it will ensure that we don't run into downtime at the 47 upgrade.  Basically, would be good to roll this out across validators while we are still on v10.

mpoke:
@faddat At the recommendation of the SDK team, we decided not to go ahead with this change. The reason is that node operators are free to change this and many of them already do.

PR #2596, opened 2023-06-15
chore: remove depguard and bump golangci-lint

faddat:
The trouble with depguard is that it recently suddenly started working, but is not the tool for the job we are trying to accomplish. 

The bump of golangci-lint is just routine maintenance.


mpoke (approved):
Thanks @faddat. Approval.

ldez:
Hello,

FYI, the latest version of golangci-lint is v1.53.2, and depguard require to be configured to be used.

- https://github.com/golangci/golangci-lint/releases
- https://golangci-lint.run/usage/linters/#depguard

faddat:
Hey @ldez thank you! 

We'd been using depguard for a long time without issue but I think it was likely doing nothing / vestigial. 

@mpoke & @ldez would you prefer to see it configured or removed?

Happy to do either.

sainoe (approved):

mmulji-ic:
> Hey @ldez thank you!
> 
> We'd been using depguard for a long time without issue but I think it was likely doing nothing / vestigial.
> 
> @mpoke & @ldez would you prefer to see it configured or removed?
> 
> Happy to do either.

Hey @faddat , I'd keep it out for the time being, we don't need it at the moment.

mmulji-ic:
@faddat we had a bump to golangci-lint here https://github.com/cosmos/gaia/pull/2562
I closed that because upgrading to that version caused the following error:

```sh
❯ make lint
--> Running linter
ante/ante.go:29:10: opts.AccountKeeper undefined (type HandlerOptions has no field or method AccountKeeper) (typecheck)
	if opts.AccountKeeper == nil {
	        ^
ante/ante.go:32:10: opts.BankKeeper undefined (type HandlerOptions has no field or method BankKeeper) (typecheck)
	if opts.BankKeeper == nil {
```
If I update the `go.mod` file, with the version version you specify, that change results in the above error.
Whats the different between the `make lint` target and changing the `go.mod` file?

Makefile
```diff
@@ -242,7 +242,7 @@ docker-build-all: docker-build-debug docker-build-hermes
 ###                                Linting                                  ###
 ###############################################################################
 golangci_lint_cmd=golangci-lint
-golangci_version=v1.52.2
+golangci_version=v1.53.2
```
mmulji-ic: @faddat we had a bump to golangci-lint here https://github.com/cosmos/gaia/pull/2562
I closed that because upgrading to that version caused the following error:

❯ make lint
--> Running linter
ante/ante.go:29:10: opts.AccountKeeper undefined (type HandlerOptions has no field or method AccountKeeper) (typecheck)
	if opts.AccountKeeper == nil {
	        ^
ante/ante.go:32:10: opts.BankKeeper undefined (type HandlerOptions has no field or method BankKeeper) (typecheck)
	if opts.BankKeeper == nil {
If I update the go.mod file, with the version version you specify, that change results in the above error.
Whats the different between the make lint target and changing the go.mod file?

faddat:
hey, I think I can clean this up tonight and also.... maybe... we are doing something twice or such

Makefile
faddat: it's odd.  I don't see golangci-lint in the go.mod file anymore?

Issue #2347, opened 2023-04-01
~80 peers, but config has been long set to 40 + 10


gaia:
## Summary of Bug

More peers than allowed in config.

## Version

9.0.1

## Steps to Reproduce

How is this possible? Note config.toml last modification date, last time the service was restarted and how many peers it should have VS how many it has.

```
$ cat config.toml | grep _num_ && curl -s http://localhost:26657/net_info | jq -r .result.n_peers && /home/ubuntu/bin/gaiad version && systemctl status gaiad | grep 'Active:' &&
 ps aux | grep [g]aiad && ll config.toml
max_num_inbound_peers = 40
max_num_outbound_peers = 10
82
v9.0.1
     Active: active (running) since Sat 2023-04-01 19:39:22 UTC; 2h 58min ago
ubuntu       140  0.0 50.6 9698672 7915472 ?     Ssl  19:39 102:39 /home/ubuntu/bin/gaiad start
-rw-r--r-- 1 ubuntu ubuntu 19K Mar 20 23:38 config.toml
```

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned
- [ ] Is a spike necessary to map out how the issue should be approached?

faddat:
Do we think that this is a gaia specific bug, or do we think that this is an issue in tenderment or comet?

gaia:
> 

Noticed it on Juno v14.0.0 but not Juno v14.1.0. Maybe something got fixed upstream

adizere:
hi @gaia 
how many nodes do you have configured in your `unconditional_peer_ids` ? See [spec/p2p](https://github.com/cometbft/cometbft/blob/ecd5ee112533cda28900cbd75afb349f67da3fa5/spec/p2p/config.md#unconditional-peers):

> Unconditional Peers
These are IDs of the peers which are allowed to be connected by both inbound or outbound regardless of max_num_inbound_peers or max_num_outbound_peers of user's node reached or not.

adizere:
> Do we think that this is a gaia specific bug, or do we think that this is an issue in tenderment or comet?

Likely Comet.

gaia:
[...]

Issue #2253, opened 2023-02-27
lost peers 


infine8:
v 8.0.0

after several hours of working the node lose its peers. nothing helps

![image](https://user-images.githubusercontent.com/2751504/221557708-1a7f1aa0-afea-4e84-ac79-0447cb1c1a7e.png)

mmulji-ic:
Hello @infine8 , we recommend using the v8.0.1 version will have some tendermint updates to help with this.

mpoke:
@infine8 could you please let us know if the issue still exists after upgrading to v8.0.1?

yj0x0x:
@mpoke Still lost peers, sync speed too slow.

mmulji-ic:
Hi @infine8 could you share your config as well as your hardware spec.

mmulji-ic:
Hi @infine8 mainnet has moved to Gaia v9.0.1. Let us know if you are experiencing the same issues, otherwise will close this issue in a week, as v8 is now only run for archive nodes.

mmulji-ic:
Hi @infine8 let us know if there is an update for this issue, otherwise will close next week.

Issue #2073, opened 2023-01-24
Full node syncing issue


jtremback:
## Summary of Bug

- There is some kind of issue in the theta release testnet resulting in full nodes not being able to sync
  - This is not affecting the validators, just full nodes
  - Important detail is that all full nodes were set up by our team. So it may not be an issue with full nodes, but it may be an issue with our configuration
  - However, this did not happen with Gaia v7, so something strange is going on.
  - Ways to fix this:
    - Review full node settings
    - Get one of the validators to set up a full node

yaruwangway:
Hi, also state-sync tutorial is not working.

faddat:
Hi there,

Can you describe the symptoms better?

faddat:
About state sync, @yaruwangway are you trying to ss for V8?

That won't work till about 1500 blocks after v8 is running and may not work at all on testnet because the script pulls from chain registry. 

Recommendations:

* fix state sync after release
* use go 1.19 in release to ensure there aren't compatibility issues with state sync
* move information on testnets to this repository
* define full node

I am going to define full node here, as archival with all history, and attempt to sync now.

yaruwangway:
Hi about state sync, I am syncing mainnet. it is working now, I opened an issue [here](https://github.com/cosmos/gaia/issues/2078). It is not a bug, just the peers are not working. We need to think a way to curl live peer from a reliable source. 

I also tried to sync v8 Rho-testnet, which is not working.

faddat:
@yaruwangway if we are dealing with v7.0.x, I can fix that.  Previously, it was frowned upon, for reasons that go beyond me, to update anything other than the tip of the repo.  

Thus, state sync stuff diverged.

yaruwangway:
Hi, @faddat , i see your pr, will test the sync, lets move to talk in the other issue, this is probably not so related this one here.

faddat:
Ok, in order to work this issue better, I'll need the exact configuration of the nodes that are failing please

faddat:
let's define a test here for this:

```bash
git clone https://github.com/cosmos/gaia
cd gaia
gh pr checkout 2077
go install ./...
gaiad init test
wget https://github.com/cosmos/testnets/raw/master/public/genesis.json.gz
gzip -d genesis.json.gz
mv genesis.json ~/.gaia/config/genesis.json
[...]

Issue #965, opened 2021-09-13
Gaiad sync issue


rocky2015aaa:
## Summary of Bug

I have two issues.

1. When I follow gaia tutorials, I get these kinds of errors several times but syncing works.
`9:18AM ERR dialing failed (attempts: 1): dial tcp 162.55.132.230:2010: connect: connection refused`
`9:18AM ERR dialing failed (attempts: 1): dial tcp 106.12.73.242:26656: i/o timeout`

2. According to quick start on document, At the `6910000` block height with panic, I need to change gaiad version.
I want to test it but for syncing I want to use quick sync.
https://cosmos.quicksync.io/
If I use quick sync, running v5.0.5 is enough? No need to run from v4.2.1 and change gaiad version?
## Version

`gaiad version` -> v4.2.1

## Steps to Reproduce

On macOS Big Sur v11.5.2

quick start
https://hub.cosmos.network/main/gaia-tutorials/join-mainnet.html#quickstart

quick sync
https://cosmos.quicksync.io/
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

tac0turtle:
> 1. When I follow gaia tutorials, I get these kinds of errors several times but syncing works.
>    `9:18AM ERR dialing failed (attempts: 1): dial tcp 162.55.132.230:2010: connect: connection refused`
>    `9:18AM ERR dialing failed (attempts: 1): dial tcp 106.12.73.242:26656: i/o timeout`

this is normal, if you wait a period of time they will pass and you will have a peer. Tendermint spits out a lot of noise, but they are benign 

> 2\. According to quick start on document, At the `6910000` block height with panic, I need to change gaiad version.
>     I want to test it but for syncing I want to use quick sync.
>     https://cosmos.quicksync.io/
>     If I use quick sync, running v5.0.5 is enough? No need to run from v4.2.1 and change gaiad version?

if you are using quick sync then you only need to use the latest binary since you already have all the old state.

PR #2635, opened 2023-07-02
fix: possible netsplits and chokes

faddat:
This PR removes depguard and cleans up root.go

Related issues:

* #704 (originally called fat gaia)
* #678 
* #2253 
* #876 
* #2347 
* #2073 
* #965 

(I'm serious about the related issues -- they all relate back to the reality that the cosmos hub network is choked by default settings)

Recent tweet that could be related unless all of these nodes are run by the same VaaS:

* https://twitter.com/Zdeadex/status/1675566228398546944?s=20

## Complimented by

* #2595 

This PR is complimented by iavl fast node because enabling iavl allows for better query handling and fixes a condition that is well known in tendermint: you can make p2p stop by hammering on the rpc endpoint, demonstrated here:

* https://twitter.com/gregosuri/status/1436181842336444416?s=20

There's another related issue I opened on the tendermint repository before it was moved to comet, please inquire privately about that. 

Combined, these could be used to destabilize the hub's p2p network. 

## Issues discovered while working on this

* https://github.com/cosmos/gaia/pull/2644
* https://github.com/cosmos/gaia/issues/2645

## note

I am unsure why this is making the E2E tests fail. It seems like it's just one of them, and if somebody could have a peek that would be great, because I am back on baby duty 🙂.

## other possibilities 

It's perfectly possible that all of this was caused by a failure at a single VAS provider or by an interruption in the internet service links between the validators who experience downtime, and the rest of the world.  In all cases, increasing the default number of inbound peers will improve performance on Gaia.


faddat:
@mmulji-ic sir, something is wierd here. 

The changes I've made in this branch should not affect the tests that it seems to break.  Is it possible you can take a look?

I may have found it -- I added the other debug commands, but the CI output above may be worth having a peek at, because I can't figure out why adding other debug commands would ever make that happen.

PR #2675, opened 2023-07-27
fix: Fix pointer to store upgrades inside loop

outofforest:
## Description

Due to how `for` loop works, if many upgrade plans are specified, the store upgrades are taken from the last one on the list, not the one matched by the upgrade name. To fix this, the copy of the object is made, before taking the pointer.

---

### Author Checklist

I have...

* [ ] Provided a link to the relevant issue or specification - not applicable
* [ ] Followed the guidelines for [building SDK modules](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules) - not applicable
* [ ] Included the necessary unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing) - not applicable
* [ ] Added a changelog entry in `.changelog` (for details, see [contributing guidelines](../../CONTRIBUTING.md#changelog)) - not applicable
* [ ] Included comments for [documenting Go code](https://blog.golang.org/godoc) - not applicable
* [ ] Updated the relevant documentation or specification - not applicable
* [ ] Reviewed "Files changed" and left comments if necessary  - not applicable

### Reviewers Checklist

I have...

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -357,7 +357,8 @@ func (app *GaiaApp) setupUpgradeStoreLoaders() {
 
 	for _, upgrade := range Upgrades {
 		if upgradeInfo.Name == upgrade.UpgradeName {
-			app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &upgrade.StoreUpgrades))
+			storeUpgrades := upgrade.StoreUpgrades
+			app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 		}
 	}
 }


MSalopek:
Thank you for opening this PR!

Would you mind clarifying why would the situation described happen?

The `Upgrades` slice is always 1 element long (we set it to be so).

It sounds like you are describing a race condition, but the code is executed sequentially (as far as I can tell) - I could be wrong about this.

outofforest:
@MSalopek 

In `go`, if you have loop like this: `for _, el := range someSlice {}` the memory where `el` exists is not reallocated on each step. Instead, new value is stored in the same memory location. It means that if I do sth like this:

```go
for _, el := range someSlice {
  pointer := &el
}
```
the address stored in `pointer` is the same on every iteration.

It means that If I do sth like this:
```go
pointers := []*someType{}
for _, el := range someSlice {
  pointers = append(pointers, &el)
}
```
after the loop is completed, surprisingly all the pointers in the `pointers` point to the same element, which is the last element in `someSlice`.

In the code I fix in this PR you have this situation, where pointer is stored for later use and at the end of the day all of them point to the last value in the iterated slice. In your case you have only one element in the slice so bug does not affect you but it's still a bug and one day might cause troubles. In my case, I discovered it when I wrote a test for our upcoming upgrade.

MSalopek:
@outofforest 
I see now, thank you very much!
It returns the address of the for loop iterator, intead of the object pointer (which is expected).

This is safe to merge. Thanks a lot!

MSalopek (approved):
LGTM!
Thank you!

yaruwangway (approved):

PR #2687, opened 2023-08-07
Fix staking & distribution e2e test with LSM

sampocs:
credit to @jstr1121 for the fix!

A delegation from a previous test was causing this check to fail - the resolution is to consider any previous delegated totals before comparing the change


MSalopek (approved):
Awesome job!

Thanks!
LGTM

PR #2692, opened 2023-08-11
Test: automate upgrade test in ci

### LINKED ISSUE #2255: Test: upgrade test in ci
## Summary

In gaia upgrade CI, the ugprade test will test the upgrade from last released major version to present version. Each major upgrade, the [test script](https://github.com/cosmos/gaia/tree/main/contrib/scripts/upgrade_test_scripts) and [test.yml](https://github.com/cosmos/gaia/blob/93cde2335fddca7c5e929f4bd042c4fb508cdaf4/.github/workflows/test.yml#L133) need to be modified.

## Proposal
Automatic test: automate the upgrade test without each time re-write the testing script and change the test.yml.  The test will upgrade a node from `[last major release]` to `[present branch version], with upgrade name taken from [upgrade file](https://github.com/cosmos/gaia/blob/189d0175b983559cd1daa8deb441579dfbf044a4/app/upgrades/v9/constants.go#L12).

____

#### For Admin Use

- [ ] Not duplicate issue

yaruwangway:
## Description

Closes: #2255

- The upgrade version is configured from reading `app/upgrade/ ` and get the latest upgrade. The upgrade version will be used to propose upgrade proposal and setup cosmovisor dir. For example, if `app/upgrade/ `  dir has two sub dir , v10 and v11, v11 will be used to propose ugprade and setup `cosmovisor/upgrades/v11/bin/gaiad`, and v11 will be used as upgrade name in the upgrade gov proposal.

- the upgradeFrom version is hardcoded (for each major release, please change [here in the test.yml](https://github.com/cosmos/gaia/blob/4a813072b191e6273b565eafc6b440ced45a474e/.github/workflows/test.yml#L151) file.), the upgradeTo version is the PR branch.

~~The upgrade test workflow steps:~~
~~1. checkout pre-release if exist, otherwise checkout latest release.~~
~~2. checkout present pr branch ,build binary~~
~~3.  checkout -,(back to latest release branch), build binary. (in step1, the binary is not build because if build there, it is lost in step 2)~~
~~4. checkout -, again back to pr branch to use the bash script in contrib for setting up the chain.~~

~~Pros:~~
~~for each major release, no need to update `https://github.com/cosmos/gaia/tree/main/contrib/scripts/upgrade_test_scripts`scripts, and no need to update upgrade test in test.yml file.~~

~~Cons:
This PR contains many steps of setup, upgrade name is configured, this might have a higher chance of err when some setup breaks this configuration.~~

---

### Author Checklist

I have...

* [ ] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [ ] Added `!` to the type prefix if API, client, or state breaking change (i.e., requires minor or major version bump)
* [ ] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
* [ ] Added a changelog entry in `.changelog` (for details, see [contributing guidelines](../../CONTRIBUTING.md#changelog))

### Reviewers Checklist

I have...


.github/workflows/test.yml
```diff
@@ -1,178 +1,208 @@
...
+#            -   uses: technote-space/get-diff-action@v6.0.1
+#                with:
+#                    PATTERNS: |
+#                        **/**.go
+#                        go.mod
+#                        go.sum
+#            -   name: Build Docker Image
+#                run: make docker-build-debug
+#            -   name: Test E2E
+#                run: make test-e2e
```
yaruwangway: this will be brought back later when PR is reviewed.

.github/workflows/test.yml
```diff
@@ -1,178 +1,208 @@
...
+            #          if: env.GIT_DIFF
+            -   name: Checkout PR branch
+                uses: actions/checkout@v3
+                with:
+                    ref: ${{ github.event.pull_request.head.ref }}
+            -   name: Install new Gaiad binary
+                run: |
+                    make build
+                    cp ./build/gaiad ./build/gaiadnew
+            #          if: env.GIT_DIFF
```
yaruwangway: if: env.GIT_DIFF will be added later when PR is reviewed.

contrib/scripts/upgrade_test_scripts/run_gaia.sh
```diff
@@ -0,0 +1,90 @@
...
+  fi
+done
+
+if [ -n "$UPGRADE_VERSION_NUMBER" ]; then
+  echo "Upgrade to version: $UPGRADE_VERSION_NUMBER"
+else
+  echo "No upgrade version found in app/upgrades."
+fi
+
+UPGRADE_VERSION=v$UPGRADE_VERSION_NUMBER
```
yaruwangway, line 26: here upgrade version is configured from the latest upgrade in `app/upgrades`.

.github/workflows/test.yml
```diff
@@ -1,178 +1,208 @@
...
+            
+                    if [ -n "$PRERELEASE_TAG" ]; then
+                      echo "Using pre-release tag: $PRERELEASE_TAG"
+                      RELEASE_TAG="$PRERELEASE_TAG"
+                    else
+                      echo "No pre-release found, using latest release tag: $LATEST_RELEASE_TAG"
+                      RELEASE_TAG="$LATEST_RELEASE_TAG"
+                    fi
+                    
+                    git checkout "$RELEASE_TAG"
```
yaruwangway: only checkout and not build the old binary here. if build the old binary here, in the next step "checkout PR branch", the binary somehow is not there anymore, store the binary in $GITHUB_WORKSPACE also does not work.

.github/workflows/test.yml
mpoke, line 1: @yaruwangway Could you please fix the indentation to avoid having the entire file modified (hard to see the diff)?

.github/workflows/test.yml
```diff
@@ -146,33 +146,63 @@ jobs:
...
         run: |
-          git checkout -
-          make build
-          cp ./build/gaiad ./build/gaiad12
-        if: env.GIT_DIFF
+            PRERELEASE_TAG=$(curl -s "https://api.github.com/repos/${{ github.repository }}/releases" | jq -r '.[] | select(.prerelease) | .tag_name' | head -n 1)
+            echo "Pre-release tag: $PRERELEASE_TAG"
+            echo "::set-output name=pre_release_tag::$PRERELEASE_TAG"
+        continue-on-error: true
+      - name: Checkout old binary branch
```
yaruwangway: maybe here we can do here manually with pointing out which branch to checkout for upgradeFrom version for each major release.

Right now ,the workflow will figure out the upgradeFrom version itself: it will use pre-release if that exists otherwise use lastest-release.
if we have pre-release v12.0.0-rc0 , latest release v11.0.0, when testing upgrade to v13, the upgrade test will use v12.0.0-rc0  which is correct.
if we have pre-release v12.0.0-rc0 , latest release v12.0.0,  when upgrade to v13, the upgrade test will use v12.0.0-rc0  which is ok, but not the best , the best is to use v12.0.0 to test upgrade to v13

a third option is to compare the string and try to figure out the latest release.

.github/workflows/test.yml
```diff
@@ -146,33 +146,34 @@ jobs:
...
-          ./contrib/scripts/upgrade_test_scripts/v12/run_gaia_v11.sh
-        if: env.GIT_DIFF
+          ./contrib/scripts/upgrade_test_scripts/run_gaia.sh
+          #          if: env.GIT_DIFF
       - name: Submit Upgrade Commands
         run: |
-          ./contrib/scripts/upgrade_test_scripts/v12/run_upgrade_commands_v12.sh 15
-        if: env.GIT_DIFF
+          ./contrib/scripts/upgrade_test_scripts/run_upgrade_commands.sh 15
+          #          if: env.GIT_DIFF
```
yaruwangway: ` if: env.GIT_DIFF` will bring back later when pr get approved.

contrib/scripts/upgrade_test_scripts/run_gaia.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/sh
+
+set -o errexit -o nounset
+set -x
+UPGRADES_DIR=$(realpath ./app/upgrades)
+UPGRADE_VERSION_NUMBER=0
+
+for dir in "$UPGRADES_DIR"/*; do
```
p-offtermatt, line 10: might be worth adding a comment in the script what this loop does

p-offtermatt (approved):
Nice, thanks for the work. Looks good to me.
I don't know where the appropriate place would be, but I feel like the behaviour should be documented somewhere, e.g. which folder is searched for releases, then the largest version number is used.
I'm only worried that if this behaviour is implicit, it might be confusing/silently break something at some point in the future.
If you know where an appropriate place would be, could you add a brief sentence about that?

Other than that, looks good!

.github/workflows/test.yml
```diff
@@ -146,33 +146,34 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: 1.20.x
-      - name: Install GaiaV11
+      - name: Install Old Gaiad
         run: |
           git checkout v11.0.0
```
insumity: Does the hardcoded `v11.0.0` version imply we have to make a change in a new release?

yaruwangway: yes, here need to be changed every time when we release a new major release, I can add a comment here. presently this test only runs on PR trying to merge to main. This means the test is targeting at upgrading from last major release to present PR branch binary version. I also tried:
1. config v11.0.0 by latest release or pre-release, it is messy depending on  latest release is before or after pre-release.
2. by using latest tag, somewhow this does not work in our repo, the latest tag is not the latest.

yaruwangway:
> Nice, thanks for the work. Looks good to me. I don't know where the appropriate place would be, but I feel like the behaviour should be documented somewhere, e.g. which folder is searched for releases, then the largest version number is used. I'm only worried that if this behaviour is implicit, it might be confusing/silently break something at some point in the future. If you know where an appropriate place would be, could you add a brief sentence about that?
> 
> Other than that, looks good!

good point, I also dont know where to add it, maybe in the internal repo release process ? cc @mpoke

MSalopek (approved):
Approve.

Maybe we should consider just downloading binaries instead of building from source. Using docker images with both binaries available could also be a good option.

PR #2750, opened 2023-09-22
chore: update mergify and dependabot

mpoke:
## Description

Closes: NA

---

### Author Checklist

I have...

- [x] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Provided a link to the relevant issue or specification
- [x] Reviewed "Files changed" and left comments if necessary 
- [ ] Confirmed all CI checks have passed

### Reviewers Checklist

I have...

- [ ] Confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Confirmed all author checklist items have been addressed
- [ ] Confirmed that this PR does not change production code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,16 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+    target-branch: "release/v13.x"
+    # Only allow automated security-related dependency updates on release branches.
+    open-pull-requests-limit: 0
+    labels:
+      - dependencies
+
+  - package-ecosystem: gomod
+    directory: "/"
+    schedule:
+      interval: daily
     target-branch: "release/v12.x"
     # Only allow automated security-related dependency updates on release branches.
     open-pull-requests-limit: 0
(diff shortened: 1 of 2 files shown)


yaruwangway (approved):

.github/dependabot.yml
```diff
@@ -41,7 +41,7 @@ updates:
     directory: "/"
     schedule:
       interval: daily
-    target-branch: "release/v10.0.x"
+    target-branch: "release/v11.x"
```
p-offtermatt, line 44: Should we add a comment that this is the last 3 releases, with a reference to e.g. https://github.com/cosmos/gaia/blob/main/RELEASING.md#major-release-maintenance?

.mergify.yml
```diff
@@ -24,29 +24,29 @@ pull_request_rules:
           {{ title }} (#{{ number }})
           {{ body }}
 
-  - name: Backport patches to the release/v10.0.x branch
+  - name: Backport patches to the release/v11.x branch
```
p-offtermatt, line 27: Same as above, maybe we should put a comment that this is supposed to be the last 3 releases included here

p-offtermatt (approved):
LGTM, just a comment. I think it probably shouldn't be addressed in this PR anyways, but I'll let you decide how to proceed/if you think it's worth doing

.github/dependabot.yml
mpoke: We should keep it as at the moment the latest two _in-production_ releases are v11 and v12.

PR #2924, opened 2024-02-05
fix: add ics4 wrapper (channel keeper) to icahost

MSalopek:
Assign `ChannelKeeper` as `ics4wrapper` when registering ICAHost.

Not providing `ics4wrapper` is not an option in IBCv7. Providing `nil` as `ics4wrapper` results in runtime panics during ICA packet handling.


sainoe (approved):
LGTM. Nice catch!

dusan-maksimovic (approved):

stana-miric (approved):

PR #3011, opened 2024-03-19
fix: exit if golang version is wrong when building

freak12techno:


MSalopek (approved):
Thanks for your contribution!

The removal was intentional, we should have echoed a `Warning: ` instead of the `Error: `.

Approving so we restore previous behavior (hard fail on go version mismatch).

freak12techno:
@MSalopek I've seen validators in Cosmos discord complaining about it being misleading (it says wrong Golang version, yet compiles), so I made it fail hard here. Also seems like other chains also make it fail hard here, for example this fix I basically took from Jackal chain and I saw a few other chains also fail hard here.
Let me know if it's better to stick with the warning approach, I can update it to warn instead, or to leave it as it is.

(IMO it makes sense to fail hard, as I guess a lot of people may miss the warning and end up building with the wrong Golang version and probably cause non-determinism later)

MSalopek:
@freak12techno Let's go with hard failing.

App hashes due to golang version mismatch have been rare after `go >= 1.20`.

We'll have this in the next release.

sainoe (approved):

PR #3033, opened 2024-03-28
chore: add v15.2.x bots targets

MSalopek:
Adds bot targets for v15.2.x

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -33,15 +33,6 @@ pull_request_rules:
         branches:
           - release/v14.2.x
 
-  - name: Backport patches to the release/v15.1.x branch
-    conditions:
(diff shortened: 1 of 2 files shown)


.mergify.yml
```diff
@@ -41,3 +41,12 @@ pull_request_rules:
       backport:
         branches:
           - release/v15.1.x
+
+  - name: Backport patches to the release/v15.2.x branch
```
mpoke, line 36: Just replace the v15.1.x as you did in the dependabot.yml file

mpoke (approved):

.mergify.yml
MSalopek: good point!

PR #3039, opened 2024-04-03
chore: update go.mod to prevent CI build breakages (add ICS v4.1.0-lsm-rc2)

MSalopek:
This PR fixes a CI issue where the build breaks during liveness and upgrade tests.

Part of the fix is updating ICS version to v4.1.0-rc1.


mpoke (approved):

p-offtermatt (approved):

app/upgrades/v16/upgrades.go
```diff
@@ -26,6 +27,11 @@ func CreateUpgradeHandler(
 		// Enable ICA controller
 		keepers.ICAControllerKeeper.SetParams(ctx, icacontrollertypes.DefaultParams())
 
+		// Set default blocks per epoch
+		providerParams := keepers.ProviderKeeper.GetParams(ctx)
```
sainoe, line 31: This will panic since `BlocksPerEpoch` key doesn't exist in the paramstore at this point.

app/upgrades/v16/upgrades.go
```diff
@@ -26,6 +27,11 @@ func CreateUpgradeHandler(
...
+		providerParams.BlocksPerEpoch = providertypes.DefaultBlocksPerEpoch
+		keepers.ProviderKeeper.SetParams(ctx, providerParams)
+
```
sainoe, line 34: We can do this instead:
```
subspace := keepers.GetSubspace(providertypes.ModuleName)
subspace.Set(ctx, providertypes.KeyBlocksPerEpoch, int64(providertypes.DefaultBlocksPerEpoch))
```

MSalopek: The upgrade code is in https://github.com/cosmos/interchain-security/pull/1757

It will should be available in the next release candidate of `ICS@v4.1.x`

app/upgrades/v16/upgrades.go
MSalopek: The upgrade code is in https://github.com/cosmos/interchain-security/pull/1757

It will should be available in the next release candidate of `ICS@v4.1.x`

app/upgrades/v16/upgrades.go
sainoe: Okay, that makes sense to do the upgrade in ICS. So why is the `BlocksPerEpoch` param set to default here (again)?

MSalopek: Making it explicit. Otherwise you have to reference ICS docs - reviewing and tracking changelogs becomes a pain.

We did a similar thing for ICA controller.

MSalopek:
Please wait with the merge.

ICS v4.1.0-rc2 needs to be linked in go.mod.

EDIT:
Done

app/upgrades/v16/upgrades.go
MSalopek: Changes are available here:
* [v4.1.0-rc1](https://github.com/cosmos/interchain-security/releases/tag/v4.1.0-rc1)

Issue #3050, opened 2024-04-09
Integrate Permissioned CosmWasm


reecepbcups:
CosmosHub delegations have now signaled the addition of CosmWasm into the cosmoshub mainnet per https://www.mintscan.io/cosmos/proposals/895

gaiad q gov proposal 895

# notes
- On Upgrade handler, set to permissioned for upload, but permissionless instantiate? (maybe permissioned on both)
- Builds can no longer be done for windows

MSalopek:
Closing as cosmwasm got merged to main in #3051 (targeting gaia v18).

Thanks you for your support in adding cosmwasm to gaia! 🚀

Issue #2950, opened 2024-02-16
Remove outdated scripts from `./contrib`


mpoke:
# Problem

Some of the content in `./contrib` is no longer used, e.g., 

- `contrib/generate_release_note/main.go`
- `contrib/scripts/upgrade_test_scripts/v8/run-gaia-v7.sh`

# Closing criteria

The `./contrib` is cleaned up.

PR #3051, opened 2024-04-09
feat!: integrate cosmwasm

reecepbcups:
ref: #3050

Adds CosmWasm and makes it permissioned in the upgrade handler per [proposal 895](https://www.mintscan.io/cosmos/proposals/895) signaling to add. 

# Test Local node Script
```
# The testnode here is permissionless. On upgrade handler, permissioned is set
GAIAD_NODE=http://localhost:26657
rm -rf ~/.gaia/ && make install
sh contrib/single-node.sh local-1 cosmos10r39fueph9fq7a6lgswu4zdsg8t3gxlqvvvyvn
```

## Upload Wasm Contract
```
sh ./contrib/upload-contract.sh
```

## Local Upgrade
```
go install github.com/cosmos/cosmos-sdk/cosmovisor/cmd/cosmovisor@latest
GAIAD_NODE=http://localhost:26657

git checkout v15.1.0
make build
cp ./build/gaiad ./build/gaiadold
go clean -modcache

git checkout wasmd-integration
make build
cp ./build/gaiad ./build/gaiadnew

./contrib/scripts/upgrade_test_scripts/run_gaia.sh

./contrib/scripts/upgrade_test_scripts/run_upgrade_commands.sh 15

./contrib/scripts/upgrade_test_scripts/test_upgrade.sh 20 5 16 localhost

gaiad q wasm params
```

# TODO:
- [x] Fix Upgrade CI (which works locally)
- [ ] Changelog

---

image


app/upgrades/v16/upgrades.go
```diff
@@ -87,6 +88,13 @@ func CreateUpgradeHandler(
 		} else if addErr != nil {
 			return vm, errorsmod.Wrapf(addErr, "unable to add rate limits")
 		}
+
+		// Set CosmWasm params
+		wasmParams := wasmtypes.DefaultParams()
+		wasmParams.CodeUploadAccess = wasmtypes.AllowNobody
+		// TODO(reece): only allow specific addresses to instantiate contracts or anyone with AccessTypeEverybody?
+		wasmParams.InstantiateDefaultPermission = wasmtypes.AccessTypeAnyOfAddresses
```
reecepbcups: Need input on the team for if this. Should it be 
- `InstantiateDefaultPermission = AccessTypeAnyOfAddresses`  _(whitelisted wallets can instantiate new contracts as much as they want off of approve code ids from gov)_
- or `InstantiateDefaultPermission = AccessTypeNobody` _(only gov only can create contracts off previously gov approved code)_

To me `AccessTypeAnyOfAddresses` makes sense.

app/upgrades/v16/upgrades.go
```diff
@@ -87,6 +89,13 @@
...
```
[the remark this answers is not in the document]
reecepbcups: false, used in keepers.WasmKeeper.SetParams(ctx, wasmParams) & confirmed in a local upgrade

image

app/upgrades/v16/upgrades.go
```diff
@@ -87,6 +89,13 @@
 		} else if addErr != nil {
 			return vm, errorsmod.Wrapf(addErr, "unable to add rate limits")
 		}
+
+		// Set CosmWasm params
+		wasmParams := wasmtypes.DefaultParams()
+		wasmParams.CodeUploadAccess = wasmtypes.AllowNobody
```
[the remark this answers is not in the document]
reecepbcups: same as below

reecepbcups:
Upgrade works locally with versions but not here, odd. 
image

.goreleaser.yml
```diff
@@ -20,10 +20,8 @@ builds:
     goos:
       - darwin
       - linux
-      - windows
     goarch:
       - amd64
-      - arm64
```
webmaster128, line 26: There should not be anything wrong with doing the same for ARM. I'd just not recommend running on ARM CPUs for production systems as Singlepass still has some smaller issues on ARM.

.goreleaser.yml
```diff
@@ -20,10 +20,8 @@ builds:
     goos:
       - darwin
       - linux
-      - windows
```
webmaster128, line 23: On Windows we can still build a client binary (through some build flags). However, the contract execution on Windows is not supported.

Dockerfile
```diff
@@ -4,7 +4,13 @@ ARG IMG_TAG=latest
 FROM golang:1.21-alpine AS gaiad-builder
 WORKDIR /src/app/
 COPY go.mod go.sum* ./
-RUN go mod download
+RUN set -eux; \
+    export ARCH=$(uname -m); \
+    WASM_VERSION=$(go list -m all | grep github.com/CosmWasm/wasmvm | awk '{print $2}'); \
```
webmaster128: I'd call this one `WASMVM_VERSION` as this is really versioning the wasmvm repo

Dockerfile
```diff
@@ -4,7 +4,13 @@ ARG IMG_TAG=latest
...
+    if [ ! -z "${WASM_VERSION}" ]; then \
+      wget -O /lib/libwasmvm_muslc.a https://github.com/CosmWasm/wasmvm/releases/download/${WASM_VERSION}/libwasmvm_muslc.${ARCH}.a; \
+    fi; \
```
webmaster128: Shoudn't this fail hard in the else case?

Dockerfile
```diff
@@ -4,7 +4,13 @@ ARG IMG_TAG=latest
 FROM golang:1.21-alpine AS gaiad-builder
 WORKDIR /src/app/
 COPY go.mod go.sum* ./
-RUN go mod download
+RUN set -eux; \
+    export ARCH=$(uname -m); \
+    WASM_VERSION=$(go list -m all | grep github.com/CosmWasm/wasmvm | awk '{print $2}'); \
+    if [ ! -z "${WASM_VERSION}" ]; then \
+      wget -O /lib/libwasmvm_muslc.a https://github.com/CosmWasm/wasmvm/releases/download/${WASM_VERSION}/libwasmvm_muslc.${ARCH}.a; \
```
webmaster128: This blindly trusts whatever GitHub returns. I think at least the checksum check should be added here even it adds a tiny bit of upgrade work. The other option would be signatures but we don't have a setup for this right now.

Dockerfile
```diff
@@ -4,7 +4,13 @@ ARG IMG_TAG=latest
 FROM golang:1.21-alpine AS gaiad-builder
```
webmaster128: [We currently need a build system with Alpine < 3.19](https://github.com/CosmWasm/wasmvm/issues/523). This seems to do the job for most teams:

```suggestion
FROM golang:1.21-alpine3.18 AS gaiad-builder
```

app/keepers/keepers.go
```diff
@@ -422,6 +431,41 @@ func NewAppKeeper(
...
+
+	wasmCapabilities := []string{
+		"iterator",
+		"staking",
+		"stargate",
+		"cosmwasm_1_1",
+		"cosmwasm_1_2",
+		"cosmwasm_1_3",
+		"cosmwasm_1_4",
+	}
```
webmaster128: Any reason for not using `AllCapabilities()` which returns exactly that list? I mean yeah, it has a misleading doc comment and is in the wrong package. But we'll fix those minor flaws. The function itself is good IMO as it simply returns all capabilities that the current wasmd has.

e2e.Dockerfile
```diff
@@ -4,7 +4,13 @@ ARG IMG_TAG=latest
...
 WORKDIR /src/app/
 COPY go.mod go.sum* ./
-RUN go mod download
+RUN set -eux; \
+    export ARCH=$(uname -m); \
+    WASM_VERSION=$(go list -m all | grep github.com/CosmWasm/wasmvm | awk '{print $2}'); \
+    if [ ! -z "${WASM_VERSION}" ]; then \
+      wget -O /lib/libwasmvm_muslc.a https://github.com/CosmWasm/wasmvm/releases/download/${WASM_VERSION}/libwasmvm_muslc.${ARCH}.a; \
+    fi; \
+    go mod download;
```
webmaster128: Same points as in the other Dockerfile

webmaster128 (commented):
Great stuff 🎉 I'm sure you know the Go integration part better than myself

app/keepers/keepers.go
reecepbcups: I really dislike having to use the wasmd app namespace to import this type. This was why I originally wanted it in wasmvm. tbh I think it makes sense to put in the wasmd/x/wasm/types as a const or something. Maybe I am being too picky idk

Will use `wasmapp.AllCapabilities()`

Dockerfile
reecepbcups: Good call out, I actually forgot to  do the static link build tag in the make file. So in the event this fails, it will fail building and give the correct output that it could not find libwasmvm. Is nice because it keeps bash to a minimum

.goreleaser.yml
reecepbcups: Re-enabled in e9058051f89db82238d6b2e17eccaa41e39adcb5

Dockerfile
reecepbcups: e9058051f89db82238d6b2e17eccaa41e39adcb5

Dockerfile
reecepbcups: ref makefile e9058051f89db82238d6b2e17eccaa41e39adcb5

app/keepers/keepers.go
reecepbcups: e9058051f89db82238d6b2e17eccaa41e39adcb5

.goreleaser.yml
reecepbcups: I'm not sure if goreleaser lets you specify the `GOOS=windows GOARCH=amd64 go build $(BUILD_FLAGS) -o build/gaiad.exe ./cmd/gaiad` for just 1 type of arch. Will have to look more into

Dockerfile
reecepbcups: resolved: 0f5ebcdf0afd3b6e8ecc83b24a99556326a87c93

contrib/single-node.sh
```diff
@@ -26,11 +26,12 @@ gaiad genesis collect-gentxs
 
 # Set proper defaults and change ports
 echo "Setting rpc listen address"
-sed -i '' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.gaia/config/config.toml
-echo 2
-sed -i '' 's/timeout_commit = "5s"/timeout_commit = "1s"/g' ~/.gaia/config/config.toml
-sed -i '' 's/timeout_propose = "3s"/timeout_propose = "1s"/g' ~/.gaia/config/config.toml
-sed -i '' 's/index_all_keys = false/index_all_keys = true/g' ~/.gaia/config/config.toml
+# sed -i '' 's#"tcp://127.0.0.1:26657"#"tcp://0.0.0.0:26657"#g' ~/.gaia/config/config.toml
```
reecepbcups, line 29: These scripts probably need to be redone tbh. Happy to do if the team would like. Maybe another PR

ref what I use: https://github.com/CosmosContracts/juno/blob/main/scripts/test_node.sh

MSalopek:
@Reecepbcups 
Hey, thanks for the your contribution! Clearly, there was quite a lot of work that went into this and we appreciate the effort!

I'd like to help pass all the checks (upgrade tests, e2e etc.) if you don't mind me adding to the PR?

As for merging this, the v16 release is already feature frozen and will be cut tomorrow morning (European time) as we have already prepared all the testing pipelines and made plans for the testnet upgrades. It's a pretty beefy release adding IBC features voted in by governance.

I think it should be pretty safe to merge this into main once we finalize the v16 release branch.

I'm really excited about this :rocket:

.goreleaser.yml
MSalopek: I can take a look at this but I think you're right about the env vars.

contrib/single-node.sh
MSalopek: We have this in the backlob but we haven't started refactoring yet: https://github.com/cosmos/gaia/issues/2950

We can add refactoring this script to that issue and do them all in a different PR.

reecepbcups:
@MSalopek Thanks and Sounds good on the v16 front. Yes, Maintainer edits are enabled.

reecepbcups:
I’ll pull in main Tuesday in prep for v18*

MSalopek:
> I’ll pull in main Tuesday in prep for v17

I already merged main into the branch (maintainer edits are enabled on the `wasm-integration`  branch). I'm fixing tests.

v17 is already cut and the propsal is in voting (it has partial set security features enabled with some other goodies).
This brings a major change for the ICS protocol and all consumer chains so it was made into its own release.

v18 is next with cosmwasm, ibc-go v7.5.0 and some other upgrades. The PR will be merged as soon as all tests are passing and we provide a definite answer regarding this discussion: https://github.com/cosmos/gaia/pull/3051#discussion_r1558435243.

The relatively fast upgrade cadence enables us to push out more features in distinct releases so we offset some risks around the upgrade.

app/upgrades/v16/upgrades.go
MSalopek: We want the governance to handle both code upload and instantiation. 

IIUC, to use permissiond wasm we would need to use these settings:
```golang
wasmParams := wasmtypes.Params{
	CodeUploadAccess:             wasmtypes.AllowNobody,
	InstantiateDefaultPermission: wasmtypes.AllowNobody,
}
```

~~I don't think that code upload is followed by an instantiation operation by default. But, that can be can be achieved through a gov prop that has 2 Txs to execute: `[MsgStoreCode, MsgInstantiateContract]`. This way, the uploaded contracts would immediately get instantiated.~~

We can use `MsgStoreAndInstantiateContract` to store and instantiate a contract via gov proposal.

Do you have any comments on this @Reecepbcups ?

e2e.Dockerfile
```diff
@@ -1,20 +1,32 @@
...
+RUN sha256sum /lib/libwasmvm_muslc.x86_64.a | grep 465e3a088e96fd009a11bfd234c69fb8a0556967677e54511c084f815cf9ce63
+RUN cp "/lib/libwasmvm_muslc.$(uname -m).a" /lib/libwasmvm_muslc.a
+
 COPY go.mod go.sum* ./
 RUN go mod download
 COPY . .
-ENV PACKAGES curl make git libc-dev bash gcc linux-headers eudev-dev python3
-RUN apk add --no-cache $PACKAGES
-RUN CGO_ENABLED=0 make install
+RUN LEDGER_ENABLED=false LINK_STATICALLY=true BUILD_TAGS=muslc make build
```
MSalopek, line 20: Had issues with this in e2e.

CGO needs to be enabled (don't know why).
I've also had to add `RUN apk add --no-cache build-base
` in the runner container to make CGO available.

It seems that we cannot run the tests without CGO enabled.

I'm not sure what the implications of this are. While checking other projects, they all seem to be running with CGO enabled (or at least they don't explicitly disable it using `CGO_ENABLED=0`.

reecepbcups: > This repo contains both Rust and Go code. The Rust code is compiled into a library (shared .dll/.dylib/.so or static .a) to be linked via cgo and wrapped with a pleasant Go API

CGO is a requirement for CW yes https://github.com/CosmWasm/wasmvm

app/upgrades/v16/upgrades.go
reecepbcups: This makes total sense yes, wasmtypes.AllowNobody is the desired param for both

[MsgStoreAndInstantiateContract](https://github.com/CosmWasm/wasmd/blob/v0.45.0/proto/cosmwasm/wasm/v1/proposal_legacy.proto#L310) looks like it will allow the uploader (gov) to override the AccessConfig though for instantiates.  `InstantiateContractProposal` does not have this. 

I would assume this overrides the defaults since gov approved, which is a feature but actually a bug in our case for expected behavior. i.e. need to ante block a proposal if `MsgStoreAndInstantiateContract.instantiate_permission` is set

MSalopek: I think I'm getting it, thanks!

We will address those concerns before testnet and write new docs before cosmwasm lands on the hub mainnet.

MSalopek (approved):
Thanks again for the PR!

All tests were cleaned up, the upgrade is passing.

There's a bit of work to do later.
We still need to check the goreleaser settings, update Dockerfile and make better upgrade tests.

The build got a bit complicated given the `libwasmvm` dep that can be either statically or dynamically linked. `upgrade-test` is using a dynamic linking approach while dockerized build uses static linking.

We will need to expand the build docs to show how to build with cosmwasm in mind.
* https://github.com/cosmos/gaia/issues/3119

app/upgrades/v16/upgrades.go
dasanchez: Hi! Do you have an example of what a proposal that uses `MsgStoreAndInstantiateContract` would look like? I would like to know how the `wasm_byte_code` field must be encoded:
```
{
 "messages": [
  {
   "@type": "/cosmwasm.wasm.v1.MsgStoreAndInstantiateContract",
   "authority": "cosmos10d07y265gmmuvt4z0w9aw880jnsr700j6zn9kn",
   "wasm_byte_code": null,  <- My understanding is the .wasm file contents will go in here
   "instantiate_permission": null,
   "unpin_code": false,
   "admin": null,
   "label": "label",
   "msg": null,
   "funds": [],
   "source": null,
   "builder": null,
   "code_hash": null
  }
 ],
 "metadata": "ipfs://CID",
 "deposit": "10000000uatom",
 "title": "Test Proposal for Store and Instantiate",
 "summary": "none"
}
```

reecepbcups: @dasanchez It can be raw or gzip compressed from the source file. Base64 encoded

So like `cat artifacts/clock_example.wasm | base64 -w 0` or `cat artifacts/clock_example.wasm | gzip -cf | base64 -w 0` iirc

dasanchez: @Reecepbcups thanks! I ended up using the `submit-proposal` command from the wasm module (rather than the gov draft-proposal template) and managed to 1) store and instantiate, and 2) execute via proposals:
```
gaiad tx wasm submit-proposal store-instantiate my_first_contract.wasm '{"count":100}' \
    --label "my first contract" \
    --no-admin \
    --instantiate-nobody true \
    --title "Store and instantiate CW template" \
    --summary "This proposal will store and instantiate the cw template contract"
```

```
gaiad tx wasm submit-proposal execute-contract $contract_address \
  '{"increment":{}}' --title "Execute CW template" \
   --summary "This proposal will execute an increment on the cw template contract"
```

```
gaiad q wasm contract-state smart $contract_address '{"get_count":{}}'
data:
  count: 101
```

ante/ante.go
```diff
@@ -25,6 +29,10 @@ type HandlerOptions struct {
 	GlobalFeeSubspace paramtypes.Subspace
 	StakingKeeper     *stakingkeeper.Keeper
 	TxFeeChecker      ante.TxFeeChecker
+
+	TxCounterStoreKey storetypes.StoreKey
+	WasmConfig        *wasmtypes.WasmConfig
+	WasmKeeper        wasmkeeper.Keeper
```
dusan-maksimovic: Looks like `WasmKeeper` isn't used anywhere, can we remove it?

app/keepers/keepers.go
```diff
@@ -437,6 +447,31 @@ func NewAppKeeper(
...
+		appKeepers.scopedWasmKeeper,
+		appKeepers.TransferKeeper,
+		bApp.MsgServiceRouter(),
+		bApp.GRPCQueryRouter(),
+		homePath,
+		wasmConfig,
+		strings.Join(wasmapp.AllCapabilities(), ","),
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+	)
+
```
dusan-maksimovic, line 484: Can't add a comment on line 513 since it wasn't changed. We need to wire-up the wasm module with IBC stack, otherwise smart contracts will not be able to create channels etc. See the snippet (and figure out if we want fee in the stack):

```go
var wasmStack porttypes.IBCModule
	wasmStack = wasm.NewIBCHandler(app.WasmKeeper, app.IBCKeeper.ChannelKeeper, app.IBCFeeKeeper)
	wasmStack = ibcfee.NewIBCMiddleware(wasmStack, app.IBCFeeKeeper)
...
ibcRouter := porttypes.NewRouter().
     AddRoute(wasmtypes.ModuleName, wasmStack).
```

reecepbcups: resolved: 56c28cd5efc480a4af1375ed9770010ad8411b59

ante/ante.go
reecepbcups: resolved: b22dcd3859df2495b4e168c1a5566bce80c461f9

stana-miric (approved):

app/modules.go
stana-miric, line 1: Do we need to update maccPerms with wasm module?

MSalopek: I added this permission based on `wasmd` [example app](https://github.com/CosmWasm/wasmd/blob/main/app/app.go).
```golang
wasmtypes.ModuleName:              {authtypes.Burner},
```

We can remove it if it interferes with something.

reecepbcups:
:fire:

PR #3060, opened 2024-04-15
fix: disable ica fee middleware

MSalopek:
Issues were uncovered during integration testing for ICA host/controller.

ICA will not use IBCFee middlaware because it would prevent gaia from interfacing with chains that don't have IBCFee enabled.

More about the issue in question can be found here: https://github.com/cosmos/ibc-go/issues/6171

IBCFee middleware will be enalbed for ICA after the issue above is adressed.


mpoke (approved):

sainoe (approved):
Administrative approval knowing that it has already been thoroughly reviewed.

PR #3083, opened 2024-04-24
fix(goreleaser): missing tendermint->cometbft update

tbruyelle:
This one escaped the migration, probably because it's located in a hidden file.

Note that this omission does not seem to have affected the releases that followed the migration (at least for v15.0.2, which is the only one I've checked), as `cometbft.TMCoreSemVer` is already equal to the correct version (it's apparently updated manually during the cometbft release process).

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,7 +16,7 @@ builds:
       - CGO_ENABLED=0
     ldflags:
       # .Env.TM_VERSION is provided in the workflow runner environment -> see .github/workflows/release.yml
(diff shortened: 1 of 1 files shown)


MSalopek (approved):
Thanks for your contribution!

I think there's a line missing when you run `gaiad version --long`, there's no other impact on the system.

LGTM.

tbruyelle:
> Thanks for your contribution!
> 
> I think there's a line missing when you run `gaiad version --long`, there's no other impact on the system.
> 
> LGTM.

Hey thank you @MSalopek
I think the affected command is more likely `gaiad cometbft version`, but as mentioned in the PR body, `TMCoreSemVer` is manually  set to the correct version during the cometbft release process, so actually, there is no impact at all (until someone forgets to do that update ;)).

MSalopek:
> > Thanks for your contribution!
> > I think there's a line missing when you run `gaiad version --long`, there's no other impact on the system.
> > LGTM.
> 
> Hey thank you @MSalopek I think the affected command is more likely `gaiad cometbft version`, but as mentioned in the PR body, `TMCoreSemVer` is manually set to the correct version during the cometbft release process, so actually, there is no impact at all (until someone forgets to do that update ;)).

Thanks for the clarification.

For completeness:
```shell
$: gaiad cometbft version
ABCI: 1.0.0
BlockProtocol: 11
P2PProtocol: 8
Tendermint: 0.37.5
```

mpoke (approved):

PR #3113, opened 2024-05-24
tests: LastValidatorPower set to voting power value

stana-miric:
When changing app state for local testing LastValidatorPower should be set to voting power. Previously it was set to 0 and this PR fixes that.


p-offtermatt (approved):
LGTM. nit: Could you make that maybe a variable to have the magic number just once in the variable declaration, and then used both when constructing the val and when setting the power here?

stana-miric:
> LGTM. nit: Could you make that maybe a variable to have the magic number just once in the variable declaration, and then used both when constructing the val and when setting the power here?

fixed f0da3dfcac9a610fb3c19c1b8fc91e1827da049d

MSalopek (approved):
LGTM! Thanks for the changes!

PR #3208, opened 2024-07-19
Remove double initialization of ModuleBasics

stana-miric:
Currently, there is a global ModuleBasics which is initialized twice (the second initialization occurs in the app and is done using the module manager). This PR implements module manager creation and initialization to follow SDK guidelines. To achieve this, it:
- removes double initialization of ModuleBasics
- moves the global ModuleBasics to be part of the app
- moves interface registration and encoding configuration creation into the app (removing the encoding configuration parameter when creating the app)


MSalopek (approved):
Awesome, thank you!

LGTM

sainoe (approved):
Administrative approval

PR #3205, opened 2024-07-16
ci: First run at interchaintests.

fastfadingviolets:
## Description

Introduces integration tests using the interchaintest framework.

These include tests for different transaction types, and consumer chain launches. They are triggered by the docker-push workflow and will test the pushed image. They can also be triggered manually and will try to test an image on the branch supplied.

For the gaia team, the main thing to review are the two github workflows.

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Provided a link to the relevant issue or specification
- [ ] Reviewed "Files changed" and left comments if necessary 
- [ ] Confirmed all CI checks have passed

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please add
your handle next to the items reviewed if you only reviewed selected items.*

I have...

- [ ] Confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] Confirmed all author checklist items have been addressed
- [ ] Confirmed that this PR does not change production code

PR #3210, opened 2024-07-21
fix: use client/v2 v2.0.0-beta3; fix broken client.toml usage

MSalopek:
Fixes a bug discovered during testing.

The CLI would ignore the variables from `client.toml` which forced operators to always use `--node <IP:port>` command line options.

The fix consists of changing `root.go` cmd wiring and updating `cosmossdk.io/client/v2` [v2.0.0-beta3](https://github.com/cosmos/cosmos-sdk/releases/tag/client%2Fv2.0.0-beta.3). The application does not use depinject.

Related:
* https://github.com/cosmos/cosmos-sdk/issues/20689
* https://github.com/cosmos/cosmos-sdk/pull/19646
* https://github.com/cosmos/cosmos-sdk/issues/18310
* example: https://github.com/cosmosregistry/chain-minimal/pull/143


p-offtermatt:
Is there any automated testing we have that would help us confirm whether this fixes the issue, or do we have to check manually?

MSalopek:
> Is there any automated testing we have that would help us confirm whether this fixes the issue, or do we have to check manually?

Tests are with Hypha, we're migrating parts of it [here](https://github.com/cosmos/gaia/pull/3205).
Re-wiring our e2e tests it not feasible at this point since new features are already in the pipeline.

Easiest way to test is to spin up a local gaia with non-default API port (`!= 26657`), update the `client.toml` and try making queries without specifying `--node`.

E.g.
```
// client.toml
node = "tcp://localhost:26658"
```

Before the fix:
```shell
gaiad status
// fails with node not available on 127.0.0.1:26657 [the default!] -- node is actually on 26658

gaiad status --node tcp://127.0.0.1:26658
// passes
```

After the fix:
```shell
gaiad status
// passes because `node = "tcp://localhost:26658"` from config.toml gets applied
```

dusan-maksimovic (approved):
Administrative approval

p-offtermatt (approved):
Thanks for the answer. That makes sense to me, so LGTM

PR #3227, opened 2024-07-22
deps: use ibc-apps rate limiter import

MSalopek:
Uses ibc-apps/rate-limiter.

Stride-labs implementation was moved the implementation to ibc-apps repo. This implementation is equivalent to the one from stride, there should be no differences in logic.


mpoke (approved):
Please add a changelog entry

dusan-maksimovic (approved):

PR #3230, opened 2024-07-26
deps!: use fixed cosmwasm version (v0.51.0)

MSalopek:
`wasmd` v0.50.x seems to have a regression that breaks amino signing on execute messages. The issue affects cosmos hub testnets running `gaiad v19.0.0-rc0` - mainnet is unaffected.

It seems that `wamd v0.50.x` is a broken release that did not get retracted so it can still be imported.
Nothing in the release notes suggested that it was a broken.

Context:
* https://github.com/CosmWasm/wasmd/pull/1844
* https://github.com/CosmWasm/wasmd/issues/1829


tests/integration/test_utils.go
```diff
@@ -27,6 +28,28 @@ import (
 
 var app *gaiaApp.GaiaApp
 
+// Some tests require a random directory to be created when runnin IBC testing suite with gaia.
+// This is due to how CosmWasmVM initializes the VM - all IBC testing apps must have different dirs so they don't conflict.
+func GaiaAppIniterRandomDir() (ibctesting.TestingApp, map[string]json.RawMessage) {
```
p-offtermatt: Maybe 
```suggestion
func GaiaAppIniterTempDir() (ibctesting.TestingApp, map[string]json.RawMessage) {
```
instead? The Randomness seems not the important part, it's that it's a dedicated directory

p-offtermatt (approved):
Tiny nit. LGTM otherwise

dusan-maksimovic (approved):

tests/integration/test_utils.go
MSalopek: Applied!

app/keepers/keepers.go
```diff
@@ -469,6 +468,7 @@ func NewAppKeeper(
 	// Must be called on PFMRouter AFTER TransferKeeper initialized
 	appKeepers.PFMRouterKeeper.SetTransferKeeper(appKeepers.TransferKeeper)
 
+	wasmDir := filepath.Join(homePath, "wasm")
```
reecepbcups, line 471: :(

ref: same issue as I did on Juno: https://github.com/CosmosContracts/juno/pull/961

reecepbcups:
New CI test needs to be written to ensure this does not happen again:

Write a CI which uploads, and executes/queries a contract. 
Then upgrade, then try to execute / query again on the SAME contract

This can not be caught if you upload a new contract, since it saved to the new (incorrect) directory. Which is easy to overlook

PR #3259, opened 2024-08-07
feat!: Bump ICS to include inactive validators

p-offtermatt:
## Description

Closes: #XXXX

Upgrades ICS to current main (which includes inactive validators).

Best reviewed by commit.

---

### Author Checklist

I have...

* [ ] Added a changelog entry in `.changelog` (for details, see [contributing guidelines](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#changelog))

### Reviewers Checklist

I have...

### DIFF
--- a/app/keepers/keepers.go
+++ b/app/keepers/keepers.go
@@ -254,16 +254,6 @@ func NewAppKeeper(
 		authcodec.NewBech32Codec(sdk.GetConfig().GetBech32ConsensusAddrPrefix()),
 	)
 
-	appKeepers.MintKeeper = mintkeeper.NewKeeper(
-		appCodec,
-		runtime.NewKVStoreService(appKeepers.keys[minttypes.StoreKey]),
-		appKeepers.StakingKeeper,
-		appKeepers.AccountKeeper,
-		appKeepers.BankKeeper,
-		authtypes.FeeCollectorName,
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
-	)
-
 	appKeepers.DistrKeeper = distrkeeper.NewKeeper(
 		appCodec,
 		runtime.NewKVStoreService(appKeepers.keys[distrtypes.StoreKey]),
@@ -321,21 +311,6 @@ func NewAppKeeper(
 		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
 	)
 
-	// provider depends on gov, so gov must be registered first
-	govConfig := govtypes.DefaultConfig()
-	// set the MaxMetadataLen for proposals to the same value as it was pre-sdk v0.47.x
-	govConfig.MaxMetadataLen = 10200
-	appKeepers.GovKeeper = govkeeper.NewKeeper(
-		appCodec,
-		runtime.NewKVStoreService(appKeepers.keys[govtypes.StoreKey]),
-		appKeepers.AccountKeeper,
-		appKeepers.BankKeeper,
-		appKeepers.StakingKeeper,
-		appKeepers.DistrKeeper,
-		bApp.MsgServiceRouter(),
-		govConfig,
-		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
-	)
 	appKeepers.ProviderKeeper = icsproviderkeeper.NewKeeper(
 		appCodec,
 		appKeepers.keys[providertypes.StoreKey],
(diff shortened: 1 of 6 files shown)


app/keepers/keepers.go
```diff
@@ -350,14 +325,49 @@ func NewAppKeeper(
...
 		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
 		authcodec.NewBech32Codec(sdk.GetConfig().GetBech32ValidatorAddrPrefix()),
 		authcodec.NewBech32Codec(sdk.GetConfig().GetBech32ConsensusAddrPrefix()),
 		authtypes.FeeCollectorName,
 	)
 
-	appKeepers.ProviderModule = icsprovider.NewAppModule(&appKeepers.ProviderKeeper, appKeepers.GetSubspace(providertypes.ModuleName))
+	// gov depends on provider, so needs to be set after
+	govConfig := govtypes.DefaultConfig()
+	// set the MaxMetadataLen for proposals to the same value as it was pre-sdk v0.47.x
```
p-offtermatt, line 337: @insumity notice that this is already present before my changes

just in general, this is there to just get the same behaviour in terms of metadata length that Gaia has always had; it's just that in 47, something changed that makes it necessary to specify this now

app/upgrades/v20/upgrades.go
```diff
@@ -25,7 +33,53 @@ func CreateUpgradeHandler(
...
 	}
 }
+
+// InitializeMaxValidatorsForExistingConsumers initializes the max validators
+// parameter for existing consumers to the MaxProviderConsensusValidators parameter.
+// This is necessary to avoid those consumer chains having an excessive amount of validators.
+func InitializeMaxValidatorsForExistingConsumers(ctx sdk.Context, providerKeeper providerkeeper.Keeper) {
+	maxVals := providerKeeper.GetParams(ctx).MaxProviderConsensusValidators
+	for _, chainID := range providerKeeper.GetAllRegisteredConsumerChainIDs(ctx) {
+		providerKeeper.SetValidatorSetCap(ctx, chainID, uint32(maxVals))
```
p-offtermatt: proposed chains would not have their validator cap set to 180.
but all chains we have on the horizon are opt-in chains that specify a validator set cap, anyways.
if this is not ultimately the case, then we need to adjust this.

go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	github.com/cosmos/ibc-apps/modules/rate-limiting/v8 v8.0.0
 	github.com/cosmos/ibc-go/modules/capability v1.0.0
 	github.com/cosmos/ibc-go/v8 v8.4.0
-	github.com/cosmos/interchain-security/v5 v5.1.1
+	github.com/cosmos/interchain-security/v5 v5.0.0-20240806104629-29327696b8e6
```
p-offtermatt, line 31: This looks like a regression, but this is just temporary and stems from upgrading to a commit on main instead of teh v5.1 branch

This will be updated when we move to the actual ICS release

app/upgrades/v20/upgrades.go
```diff
@@ -25,7 +46,80 @@ func CreateUpgradeHandler(
...
+
+	err = stakingKeeper.SetParams(ctx, params)
+	if err != nil {
+		panic(err)
+	}
+}
+
+// InitializeLastProviderConsensusValidatorSet initializes the last provider consensus validator set
+// by setting it to the first 180 validators from the current validator set of the staking module.
+func InitializeLastProviderConsensusValidatorSet(ctx sdk.Context, providerKeeper providerkeeper.Keeper, stakingKeeper stakingkeeper.Keeper) {
```
p-offtermatt: Thanks @insumity, yes the `LastProviderConsensusValSet` should be initialized here. Nice catch

app/upgrades/v20/upgrades.go
```diff
@@ -25,7 +46,80 @@ func CreateUpgradeHandler(
...
+		InitializeLastProviderConsensusValidatorSet(ctx, keepers.ProviderKeeper, *keepers.StakingKeeper)
+
 		ctx.Logger().Info("Upgrade v20 complete")
 		return vm, nil
 	}
 }
+
+// InitializeMissingValidatorSetCaps initializes the max validators
+// parameter for existing consumers to the MaxProviderConsensusValidators parameter.
+// This is necessary to avoid those consumer chains having an excessive amount of validators.
```
mpoke: Is this needed? If they don't allow inactive validators, then they cannot have more than 180 anyway, right? My concern is that if new consumers joins before the v20 upgrade, this will set their valset cap to 180 regardless of their power shaping config.

app/upgrades/v20/upgrades.go
```diff
@@ -25,7 +46,80 @@ func CreateUpgradeHandler(
...
+	}
+}
+
+// InitializeMaxProviderConsensusParam initializes the MaxProviderConsensusValidators parameter.
+// It is set to 180, which is the current number of validators participating in consensus on the Cosmos Hub.
+// This parameter will be used to govern the number of validators participating in consensus on the Cosmos Hub,
+// and takes over this role from the MaxValidators parameter in the staking module.
+func InitializeMaxProviderConsensusParam(ctx sdk.Context, providerKeeper providerkeeper.Keeper) {
+	params := providerKeeper.GetParams(ctx)
+	if params.MaxProviderConsensusValidators == 0 {
```
mpoke: Why is this if condition necessary?

app/upgrades/v20/upgrades.go
```diff
@@ -25,7 +46,80 @@ func CreateUpgradeHandler(
...
+// SetMaxValidators sets the MaxValidators parameter in the staking module to 200,
+// which is the current number of 180 plus 20.
+// This is done in concert with the introduction of the inactive-validators feature
+// in Interchain Security, after which the number of validators
+// participating in consensus on the Cosmos Hub will be governed by the
+// MaxProviderConsensusValidators parameter in the provider module.
+func SetMaxValidators(ctx sdk.Context, stakingKeeper stakingkeeper.Keeper) {
+	params, err := stakingKeeper.GetParams(ctx)
+	if err != nil {
+		panic(err)
```
mpoke: Do not panic and return an error. See https://github.com/cosmos/gaia/blob/v18.0.0/app/upgrades/v16/upgrades.go#L94 for an example.

app/upgrades/v20/upgrades.go
```diff
@@ -25,7 +46,80 @@ func CreateUpgradeHandler(
...
+	}
+}
+
+// InitializeLastProviderConsensusValidatorSet initializes the last provider consensus validator set
+// by setting it to the first 180 validators from the current validator set of the staking module.
+func InitializeLastProviderConsensusValidatorSet(ctx sdk.Context, providerKeeper providerkeeper.Keeper, stakingKeeper stakingkeeper.Keeper) {
+	vals, err := stakingKeeper.GetBondedValidatorsByPower(ctx)
+	if err != nil {
+		panic(err)
```
mpoke: ditto: don't panic

app/upgrades/v20/upgrades.go
p-offtermatt: Good point. Yeah, I wanted this as another "line of defense", but it's not strictly necessary anymore, since inactive validators will by default not be allowed.
Seems better to remove this

app/upgrades/v20/upgrades.go
p-offtermatt: Not necessary, just double-checking. Can also remove this

app/upgrades/v20/upgrades.go
p-offtermatt: Done, thanks

mpoke (approved):

app/upgrades/v20/upgrades.go
```diff
@@ -22,10 +45,86 @@ func CreateUpgradeHandler(
...
+// by setting it to the first 180 validators from the current validator set of the staking module.
+func InitializeLastProviderConsensusValidatorSet(
+	ctx sdk.Context, providerKeeper providerkeeper.Keeper, stakingKeeper stakingkeeper.Keeper,
+) error {
+	vals, err := stakingKeeper.GetBondedValidatorsByPower(ctx)
+	if err != nil {
+		return err
+	}
+
+	// cut the validator set to the first 180 validators
```
insumity, line 111: Isn't `vals` going to contain exactly 180 validators during the upgrade?
If yes, maybe we should return an error here because something is off?

insumity (approved):
Left a comment but otherwise looks good to me.
Would it also be possible to have the lint check pass before merging?

app/upgrades/v20/upgrades.go
p-offtermatt: I think it's probably ok like this

PR #3278, opened 2024-08-15
ci: interchain tests for cosmwasm, plus IBC & jailing on consumer chains

fastfadingviolets:
## Description

Porting some of Hypha's tests into gaia. These have basic permissioned cosmwasm, plus IBC transactions, and tests for validator jailing on consumer chains

### Author Checklist

I have...

- [x] Included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] Targeted the correct branch (see [PR Targeting](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#pr-targeting))
- [ ] Confirmed all CI checks have passed


dasanchez (approved):

stana-miric (approved):
LGTM
I have a small suggestion regarding the module name. It would be better to remove the version (v20) from the middle of the module path(module github.com/cosmos/gaia/v20/tests/interchain). Having the version in the middle makes the module name invalid according to Go's module naming conventions and causes issues when trying to import this module into other projects, as it treats interchain as a subpackage of gaia. 
Not mandatory to change it but it would be good just in case if we decide to reuse this setup in ics tests

fastfadingviolets:
will do @stana-miric i'll add that to a future PR

PR #3293, opened 2024-08-26
test: rm upgrade tests; avoid env teardown during exec

MSalopek:
e2e testing refactor.

1. remove software upgrade test
* test is covered by upgrade-test on the CI
* test is covered by interchaintest

Test was removed because it causes the environment to be re-created which is a waste of time given that the case is already covered. The cancel and expedite upgrade are stil valid tests and should not be removed at this point.


dusan-maksimovic (approved):
Administrative approval

mpoke (approved):

Issue #3319, opened 2024-09-06
Security documentation inaccurate


faddat:
## this is what security.md says

![Image](https://github.com/user-attachments/assets/1ba38b4c-f131-447e-a245-2f1e927d2241)

## this is what happens if you follow security.md

https://x.com/gadikian/status/1832105330802921675?t=sw2JE_oJ3SIcNveXHlqfTQ&s=19

## this is Amulet changing the security reporting rules one day after closing Joe Bowmans report 

![Image](https://github.com/user-attachments/assets/59c90f5e-e904-4f69-a378-9dacc38f13ac)

https://hackerone.com/cosmos/policy_versions?change=3736457&type=team

The new reporting rules from amulet state That there are third-party components in Gaia. But they don't describe who is a third party. Informal is a third party? How about all of the various libraries in the go programming language? Third party? 

IBC? Third party? 

Critically, the documentation for making security reports found here does not match what amulet says the process is. 

What is the process?

https://acrobat.adobe.com/id/urn:aaid:sc:AP:f8e9e3d5-bd7e-41a6-958a-ef180329f83f

## who is a third party?

Seems to me that you are routing all security concerns about the Hub to an organization that is not concerned with the security of the Hub as a whole.  But the hub is a whole.  

It is safe or not.

It doesn't care who made the bits and bobs in it. 

## who is deciding who is a third party?

It seems as though the foundation has labeled skip as a third party, and says that their code is often a source of security vulnerabilities. This is not true. One needs to look no further than the recent critical on ICS to understand that everybody's code is frequently a source of security vulnerabilities.

## what are reporters to do from here?

* There's no definition of what is and is not third party 
* There's no incentive to report bugs on the Hub in fact there's disincentive 
* Seems the foundation is being very clear: the security of the cosmos hub is not its concern

faddat:
@mpoke the information in security.md does not contain Amulet's note on gaia, and is therefore surely inaccurate. 

Do you motion to remove the fee market from Gaia?

Do you agree with their asssessment that since it is "third party" (whatever that means) software, it is innately unsafe?  I disagree with them and think that the fee market is the cure p2p storms and related issues always needed.

If not, clearly no one should be taking issues to amulet/icf.

Because they don't support it, and say that its third party nature weakens security on the hub.

So let's next make a PR to remove the fee market, since it seems that is what is advised by Amulet.

Eric-Warehime:
[...]

PR #3320, opened 2024-09-06
docs: Delete SECURITY.md

faddat:
## Description

Having no documentation here is much better than having inaccurate documentation here.

closes:

* https://github.com/cosmos/gaia/issues/3319


faddat:
An interesting alternative to merging this is of course to include a notice that the ICF and amulet reserve the right to reclassify anything at any time of their choosing, without even specifying what is and is not third-party code.

mpoke:
The information in SECURITY.md is accurate.

PR #3430, opened 2024-11-28
feat!: Add upgrade handler for ics consumer infraction params

stana-miric:
Adds an update handler to support changes from [cosmos/interchain-security#2403](https://github.com/cosmos/interchain-security/pull/2403).


MSalopek (approved):

app/upgrades/v22/upgrades.go
```diff
@@ -27,7 +32,37 @@ func CreateUpgradeHandler(
...
+			return vm, errorsmod.Wrapf(err, "running module migrations")
+		}
+
 		ctx.Logger().Info("Upgrade v22 complete")
 		return vm, nil
 	}
 }
+
+func SetConsumerInfractionParams(ctx sdk.Context, pk providerkeeper.Keeper) error {
+	infractionParameters := DefaultInfractionParams()
```
mpoke: nit: Why not use [DefaultConsumerInfractionParameters](https://github.com/cosmos/interchain-security/blob/fe2ec1e9f31eba2b000af00c0165cb043bda3e6a/x/ccv/provider/types/provider.go#L32)?

mpoke (approved):

app/upgrades/v22/upgrades.go
stana-miric: I overlooked that the slashing keeper is available here. It's definitely better to take the default parameters from the provider module. fixed https://github.com/cosmos/gaia/pull/3430/commits/ad71048399a839ed0ecc1dd7297c1c46ce70f746

PR #3445, opened 2024-12-11
fix: export only active consensus validators

fastfadingviolets:
## Description

The genesis import will ask each module which validators to take as active. The provider will answer with only the active consensus validators. If the genesis export file contains a different set of validators, the import will fail.

---

### Author Checklist

I have...

* [ ] Added `!` to the type prefix if API, client, or state breaking change (i.e., requires minor or major version bump)
* [x] Added a changelog entry in `.changelog` (for details, see [contributing guidelines](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#changelog))

### Reviewers Checklist

I have...

### DIFF
--- a/app/export.go
+++ b/app/export.go
@@ -2,6 +2,7 @@ package gaia
 
 import (
 	"encoding/json"
+	"sort"
 
 	tmproto "github.com/cometbft/cometbft/proto/tendermint/types"
 
@@ -43,6 +44,18 @@ func (app *GaiaApp) ExportAppStateAndValidators(
 	}
 
 	validators, err := staking.WriteValidators(ctx, app.StakingKeeper)
+	if err != nil {
+		return servertypes.ExportedApp{}, err
+	}
+	sort.SliceStable(validators, func(i, j int) bool {
+		return validators[i].Power > validators[j].Power
+	})
+	// we have to trim this to only active consensus validators
+	maxVals := app.ProviderKeeper.GetMaxProviderConsensusValidators(ctx)
+	if len(validators) > int(maxVals) {
+		validators = validators[:maxVals]
+	}
+
 	return servertypes.ExportedApp{
 		AppState:        appState,
 		Validators:      validators,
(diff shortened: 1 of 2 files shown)


app/export.go
```diff
@@ -43,6 +44,18 @@ func (app *GaiaApp) ExportAppStateAndValidators(
...
 
 	validators, err := staking.WriteValidators(ctx, app.StakingKeeper)
+	if err != nil {
+		return servertypes.ExportedApp{}, err
+	}
+	sort.SliceStable(validators, func(i, j int) bool {
+		return validators[i].Power > validators[j].Power
+	})
+	// we have to trim this to only active consensus validators
+	maxVals := app.ProviderKeeper.GetMaxProviderConsensusValidators(ctx)
```
MSalopek, line 54: This would delete validators that are allowed to validate consumer chains but are not in the active validator set of the hub.

Would you mind expanding on the issue?

fastfadingviolets: Sure. [This](https://github.com/cosmos/cosmos-sdk/blob/ec58adf8e2fa4149c093d871d2fd398ca508525a/baseapp/abci.go#L111-L116) assertion fails when you're importing a chain right now. This is because [the provider keeper](https://github.com/cosmos/interchain-security/blob/dc06ecaf80970407c8f863fa21635fc90f52f62d/x/ccv/provider/keeper/genesis.go#L85-L96) only answered with the consensus validators.

Would the validators actually be deleted? I was expecting them to still be in the app_state for `staking` so that once the chain starts up they get added back to the set, but I may be wrong

MSalopek: This command if for exporting when creating a new `genesis.json` if I'm not mistaken.

In that case, yes, they would be deleted and not available in the new chain.

fastfadingviolets: I'll do some testing next week and come back to this

fastfadingviolets:
Closing to reopen with a different branch.

PR #3455, opened 2024-12-18
feat! Break out LSM module from SDK

Eric-Warehime:
## Overview
This pull request moves the liquid staking functionality from the SDK fork into its own dedicated module, `x/lsm`. 
The following is a summary of the APIs and types of this module and highlights some of the key differences between 
this implementation and the previously integrated functionality living in the SDK's `features/v0.50.x-lsm` branch.

You can refer to [this](https://github.com/cosmos/cosmos-sdk/compare/release/v0.50.x...feature/v0.50.x-lsm) diff to see the full changeset which the fork requires--it can be useful to refer to this and  compare to the diff in this 
PR, although I've tried to highlight the major differences below since the branch has 63,000+ lines changed.

---

### **Validator Bonds**
The existing lsm implementation made a couple of breaking changes to the staking module.
1. Deprecating the `min_self_delegation` field
2. Implementing `ValidatorBond` and `UnbondValidator` transactions

These changes introduced a new transaction type which specified a validator and a delegator indicating the intent to 
mark a specific delegation as a validator bond. The amount for each validator which was marked specifically as 
Validator Bonds was used along with the `validator_bond_factor` field which was added to the staking module's `Params`.

The combination of these things was used to determine a limit for how many liquid staking tokens a validator could 
issue.  
```
    Tokenize Share Cap = Bonded Tokens * validator_bond_factor
```

The PR removes the concept of `ValidatorBond` and `UnbondValidator` along with the bond factor param. It removes the 
deprecated label from the `min_self_delegation` field. The logic which uses the validator bonds in order to enforce 
limits on tokenized shares has been preserved, but the calculations for limits now looks up self delegations from 
the validator and uses that total as the validator bond amount.

Since the validator bond shares are stored in a new storage primitive specific to the lsm module, `LiquidValidator`, 
this PR uses the staking hooks to update the value any time a delegation is updated--checking whether the delegation 
was a validator self delegation and updating the value corresponding.

It's possible in a followup PR that we should remove these limits entirely. I advocate for setting this value to 
`ValidatorBondCapDisabled` which removes this limit entirely when doing the upgrade.

### **Liquid Staking Caps**
The existing parameters limiting liquid stake issuance are `global_liquid_staking_cap` and 
`validator_liquid_staking_cap`. These limit the sum of all tokenized shares and the per-validator tokenized shares 
respectively.

This PR preserves those limits and calculations for tokenization of shares which occurs through the lsm module 
itself. However, the existing implementation also altered the staking module itself to include delegations from ICAs.
The rationale behind this is that LST protocols are currently implemented as interchain applications and the 
majority of their LSTs are issued on behalf of stake which is delegated via ICAs.

As we're no longer forking the staking module, it's not practical to calculate these values--the existing hooks 
don't provide amounts. Also, the regulation of external protocols doesn't seem like something our module should be 
trying to regulate so I've removed those calculations entirely and limited these parameters to lsm module backed shares.

## Storage Primitives
Most of the existing storage primitives used in the upstream implementation are preserved. `TokenizeShareRecord`s 
are still the basic way of recording tokenized delegations. The disabling and enabling of lsm functionality per 
validator is preserved. And the `TokenizeShareRecordReward` has been migrated from the distribution keeper for 
rewards tabulation.

The new primitive introduced for module storage is the `LiquidValidator` which is simply keyed on validator operator 
address and stores the current bond shares (calculated via self delegations) and liquid shares issued for each 
validator.

The upgrade code to accomplish these migrations and populate the module storage on upgrade is not included in this PR.

### **`distribution.go`**
This contains all the code migrated from the upstream distribution keeper fork. It is used solely for withdrawing 
rewards accumulated to a tokenize share record.
### **`hooks.go`**
These hooks are all newly implemented. They mostly replace the parts of the code which were altering existing 
staking, distribution, and slashing module code. Its purpose is to update validator bond shares and the 
`LiquidValidator` records which keep track of total liquid staked shares a validator has issued.
### **`liquid_stake.go`**
Directly ported from the upstream implementation. This is mostly convenience methods for interacting with and 
implementing logic around liquid staking.
### **`tokenize_share_record.go`**
Contains most of the logic for store updates and access.

## Keeper Dependencies
In places where the previous implementation was using the forked Keepers directly I've had to replace them with 
dependencies in the new module. This means the lsm module requires the `AccountKeeper`, `BankKeeper`, 
`StakingKeeper`, and `DistributionKeeper` in order to work. You can see the required methods in `expected_keepers.go`.

In tests, I've added mocks generated via Mockery to replace these implementations.


faddat:
Progress is a beautiful thing

mpoke:
> The PR removes the concept of ValidatorBond and UnbondValidator along with the bond factor param. It removes the
deprecated label from the min_self_delegation field. The logic which uses the validator bonds in order to enforce
limits on tokenized shares has been preserved, but the calculations for limits now looks up self delegations from
the validator and uses that total as the validator bond amount.

This requires migration that sets the `MinSelfDelegation` for all the validators to zero. This results in two potential issues:

1. The `Tokenize Share Cap` will be zero. 
2. If a validator unbonds its own stake (delegated from its own address), then it will get jailed (see [here](https://github.com/cosmos/cosmos-sdk/blob/release%2Fv0.50.x/x/staking/keeper/delegation.go#L1015)). This could be fixed by either having all validators update their `MinSelfDelegation` accordingly or doing so automatically through a state migration.  

> It's possible in a followup PR that we should remove these limits entirely. I advocate for setting this value to
ValidatorBondCapDisabled which removes this limit entirely when doing the upgrade.

This will fix the first issue (as the `Tokenize Share Cap` will no longer matter). However, it will remove the [security feature introduced for liquid staking](https://github.com/cosmos/cosmos-sdk/blob/feature%2Fv0.50.x-lsm/docs/architecture/adr-061-liquid-staking.md#validator-bond), i.e., 
> As an additional security feature, validators who want to receive delegations from liquid staking providers would be required to self-bond a certain amount of tokens. The validator self-bond, or “validator-bond,” means that validators need to have “skin in the game” in order to be entrusted with delegations from liquid staking providers. This disincentivizes malicious behavior and enables the validator to negotiate its relationship with liquid staking providers.

Is there a plan to reintroduce this security feature?

mpoke:
> As we're no longer forking the staking module, it's not practical to calculate these values--the existing hooks
don't provide amounts. Also, the regulation of external protocols doesn't seem like something our module should be
trying to regulate so I've removed those calculations entirely and limited these parameters to lsm module backed shares.

IIUC, this means that the liquid staking caps will cover only tokenized shares issued by the LSM module and no longer include delegation from liquid staking providers, right? This means that another [key security feature of LSM](https://github.com/cosmos/cosmos-sdk/blob/feature%2Fv0.50.x-lsm/docs/architecture/adr-061-liquid-staking.md#limiting-liquid-staking) is removed, i.e., 
> The LSM would limit the percentage of liquid staked tokens by all liquid staking providers to 25% of the total supply of staked tokens. For example, if 100M tokens were currently staked, and if the LSM were installed today then the total liquid staked supply would be limited to a maximum of 25M tokens.
> 
> This is a key safety feature, as it would prevent liquid staking providers from collectively controlling more than ⅓ of the total staked token supply, which is the threshold at which a group of bad actors could halt block production.

Are there plans to reintroduce this security feature?

mpoke:
nit: 

> The disabling and enabling of lsm functionality per validator is preserved.

This functionality is per delegator.

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+		}
+	}
+
+	authorizations.Addresses = updatedAddresses
+	k.SetPendingTokenizeShareAuthorizations(ctx, completionTime, authorizations)
+}
+
+// Unlocks all queued tokenize share authorizations that have matured
+// (i.e. have waited the full unbonding period)
+func (k Keeper) RemoveExpiredTokenizeShareLocks(ctx context.Context, blockTime time.Time) ([]string, error) {
```
mpoke: This method is never called. In the original implementation it was called in the BeginBlock of the staking module.

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+	}
+
+	return unlockedAddresses, nil
+}
+
+// Calculates and sets the global liquid staked tokens and liquid shares by validator
+// The totals are determined by looping each delegation record and summing the stake
+// if the delegator is the lsm.
+// This function must be called in the upgrade handler which onboards LSM
+func (k Keeper) RefreshTotalLiquidStaked(ctx context.Context) error {
```
mpoke: This method is never called. Should assess if the logic is needed for migration. Otherwise, remove it to simplify code.

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// SafelyDecreaseValidatorBond decrements the validator's self bond
+// so long as it will not cause the current delegations to exceed the threshold
+// set by validator bond factor
+func (k Keeper) SafelyDecreaseValidatorBond(ctx context.Context, valAddress sdk.ValAddress, shares math.LegacyDec) error {
```
mpoke: This method is not called. If the validator bond concept was removed, there is no reason to keep this method.

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+	if err != nil {
+		return types.LiquidValidator{}, err
+	}
+
+	return validator, nil
+}
+
+// Increase validator bond shares increments the validator's self bond
+// in the event that the delegation amount on a validator bond delegation is increased
+func (k Keeper) IncreaseValidatorBondShares(ctx context.Context, valAddress sdk.ValAddress, shares math.LegacyDec) error {
```
mpoke: Not sure I understand the logic behind this function given that the validator bond was removed. Currently, it is called by `RedeemTokensForShares` when the delegator address matches the validator address, but I'm not sure of the reason for doing so. What is the invariant for `validator.ValidatorBondShares`?

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+	err = k.SetLiquidValidator(ctx, validator)
+	if err != nil {
+		return types.LiquidValidator{}, err
+	}
+
+	return validator, nil
+}
+
+// DecreaseValidatorLiquidShares decrements the liquid shares on a validator
+func (k Keeper) DecreaseValidatorLiquidShares(ctx context.Context, valAddress sdk.ValAddress,
```
mpoke: In the original implementation, this method is called in four scenarios:

- a redelegation
- an undelegation
- converting tokenized shares back to native delegations (RedeemTokensForShares)
- slashing a redelegation

Now, the method is called only from RedeemTokensForShares. What is the reason for the change in behavior?

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+}
+
+// SafelyIncreaseValidatorLiquidShares increments the liquid shares on a validator, if:
+// the validator bond factor and validator liquid staking cap will not be exceeded by this delegation
+//
+// The percentage of validator liquid shares must be less than the ValidatorLiquidStakingCap,
+// and the total liquid staked shares cannot exceed the validator bond cap
+//  1. (TotalLiquidStakedTokens / TotalStakedTokens) <= ValidatorLiquidStakingCap
+//  2. LiquidShares <= (ValidatorBondShares * ValidatorBondFactor)
+func (k Keeper) SafelyIncreaseValidatorLiquidShares(ctx context.Context, valAddress sdk.ValAddress,
```
mpoke: ditto: in the original implementation, this method is called in several scenarios, while now is called only when tokenizing shares.

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+	if exceedsCap {
+		return types.ErrGlobalLiquidStakingCapExceeded
+	}
+
+	k.SetTotalLiquidStakedTokens(ctx, k.GetTotalLiquidStakedTokens(ctx).Add(amount))
+	return nil
+}
+
+// DecreaseTotalLiquidStakedTokens decrements the total liquid staked tokens
+func (k Keeper) DecreaseTotalLiquidStakedTokens(ctx context.Context, amount math.Int) error {
```
mpoke: ditto: This method is called in several scenarios in the original implementation -- undelegate, redeem tokens, slashing, and slashing redelegations -- while now is called only when redeeming tokens and slashing a validator.

x/lsm/keeper/liquid_stake.go
```diff
@@ -0,0 +1,642 @@
...
+
+	return liquidStakePercent.GT(liquidStakingCap), nil
+}
+
+// SafelyIncreaseTotalLiquidStakedTokens increments the total liquid staked tokens
+// if the global cap is not surpassed by this delegation
+//
+// The percentage of liquid staked tokens must be less than the GlobalLiquidStakingCap:
+// (TotalLiquidStakedTokens / TotalStakedTokens) <= GlobalLiquidStakingCap
+func (k Keeper) SafelyIncreaseTotalLiquidStakedTokens(ctx context.Context, amount math.Int, sharesAlreadyBonded bool) error {
```
mpoke: This method is only called when tokenizing shares (as compared to the original implementation). Is this because delegations from LSPs are no longer considered to be liquid staked tokens? If so, what's the reason behind keeping track of this value? Is there a reason to cap the amount of tokenized shares?

x/lsm/keeper/hooks.go
```diff
@@ -0,0 +1,119 @@
...
+		return err
+	}
+	liquidVal, err := h.k.GetLiquidValidator(ctx, valAddr)
+	if err != nil {
+		return err
+	}
+	initialLiquidTokens := validator.TokensFromShares(liquidVal.LiquidShares).TruncateInt()
+	slashedLiquidTokens := fraction.Mul(sdkmath.LegacyNewDecFromInt(initialLiquidTokens))
+
+	decrease := slashedLiquidTokens.TruncateInt()
```
mpoke: The way the `decrease` value is calculated here is different from the [original implementation](https://github.com/cosmos/cosmos-sdk/blob/feature%2Fv0.50.x-lsm/x/staking/keeper/slash.go#L175). Would be useful to have a test verifying that the calculation is correct.

x/lsm/keeper/msg_server.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+	if found {
+		return nil, types.ErrRedelegationInProgress
+	}
+
+	// If this tokenization is NOT from a liquid staking provider,
+	//   confirm it does not exceed the global and validator liquid staking cap
+	// If the tokenization is from a liquid staking provider,
+	//   the shares are already considered liquid and there's no need to increment the totals
+	if !k.DelegatorIsLiquidStaker(delegatorAddress) {
```
mpoke: This is no longer accurate given that the caps no longer include delegations from LSPs.

x/lsm/keeper/msg_server.go
```diff
@@ -0,0 +1,535 @@
...
+
+	// prevent redemption that returns a 0 amount
+	if tokens.IsZero() {
+		return nil, types.ErrTinyRedemptionAmount
+	}
+
+	// If this redemption is NOT from a liquid staking provider, decrement the total liquid staked
+	// If the redemption was from a liquid staking provider, the shares are still considered
+	// liquid, even in their non-tokenized form (since they are owned by a liquid staking provider)
+	if !k.DelegatorIsLiquidStaker(delegatorAddress) {
```
mpoke: ditto: This is no longer accurate given that the caps no longer include delegations from LSPs.

x/lsm/keeper/msg_server.go
```diff
@@ -0,0 +1,535 @@
...
+	// Check if the delegator has disabled tokenization
+	lockStatus, unlockTime := k.GetTokenizeSharesLock(ctx, delegatorAddress)
+	if lockStatus == types.TOKENIZE_SHARE_LOCK_STATUS_LOCKED {
+		return nil, types.ErrTokenizeSharesDisabledForAccount
+	}
+	if lockStatus == types.TOKENIZE_SHARE_LOCK_STATUS_LOCK_EXPIRING {
+		return nil, types.ErrTokenizeSharesDisabledForAccount.Wrapf("tokenization will be allowed at %s", unlockTime)
+	}
+
+	// ValidatorBond delegation is not allowed for tokenize share
```
mpoke: nit: fix comment

PR #3461, opened 2024-12-22
fix: typos in documentation files

vtjl10:
Corrected `similar to contract to contract interactions` to `similar to contract interactions`


vtjl10:
@zmanian @jackzampolin Could you check this PR please?

PR #3493, opened 2025-01-25
test: add fee test with wrong denom order

DeVikingMark:
Add test to verify that transactions with fees in non-standard denomination order are processed correctly. 
This ensures that the chain is resilient to different orderings of fee denominations in transactions.

The test creates a transaction with fees ordered as "stake" then "atom" and verifies successful processing.


tests/e2e/e2e_test.go
```diff
@@ -86,7 +91,40 @@ func (s *IntegrationTestSuite) TestSlashing() {
 	s.testSlashing(chainAPI)
 }
 
-// todo add fee test with wrong denom order
+func (s *IntegrationTestSuite) TestFeeWithWrongDenomOrder() {
```
Eric-Warehime: Looks good. Can you move this to be a function in `e2e_bank_test.go` and invoke it from within the `TestBank` function?

tests/e2e/e2e_test.go
```diff
@@ -86,7 +91,40 @@ func (s *IntegrationTestSuite) TestSlashing() {
...
+	s.Require().NoError(err)
+	
+	txBuilder.SetFeeAmount(fees)
+	txBuilder.SetGasLimit(200000)
+	
+	// The transaction should still be accepted and processed correctly
+	// despite the fees being in a different order
+	tx, err := s.chainA.SignAndBroadcastTx(val1.keyInfo, txBuilder)
+	s.Require().NoError(err)
+	s.Require().Equal(uint32(0), tx.Code)
```
Eric-Warehime: Would be good to use a `require.Eventually` block like in https://github.com/cosmos/gaia/blob/accd9978e2f9fbe8ae296dd40d24e7c4617b7f9c/tests/e2e/e2e_bank_test.go#L40 to actually validate the send worked.

DeVikingMark:
Hey @Eric-Warehime, I've addressed your review comments:
- Moved the test into `e2e_bank_test.go` where it belongs
- Added it to `TestBank` function
- Made the test more robust with `require.Eventually` to properly verify the tx goes through
- Cleaned up by removing the old test from `e2e_test.go`

Eric-Warehime:
> Hey @Eric-Warehime, I've addressed your review comments:
> 
> * Moved the test into `e2e_bank_test.go` where it belongs
> * Added it to `TestBank` function
> * Made the test more robust with `require.Eventually` to properly verify the tx goes through
> * Cleaned up by removing the old test from `e2e_test.go`

Your test is failing and not passing lint.

Eric-Warehime:
```
=== RUN   TestIntegrationTestSuite/TestBank/test_fee_with_wrong_denom_order
    e2e_bank_test.go:287: 
        	Error Trace:	/home/runner/work/gaia/gaia/tests/e2e/e2e_bank_test.go:287
        	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.10.0/suite/suite.go:115
        	Error:      	Condition never satisfied
        	Test:       	TestIntegrationTestSuite/TestBank/test_fee_with_wrong_denom_order
        	Messages:   	Transaction with wrong denom order should succeed

DeVikingMark:
@Eric-Warehime I have some problems running tests on Windows, but now should be fine. Could you please run CI to check if everything is correct now?

PR #3578, opened 2025-02-28
fix!: upgrade to include tendermint client to allowed clients

vladjdk:
## Description

Fixes tendermint client not being in ClientKeeper AllowList

---

### Author Checklist

I have...

* [x] Provided a link to the relevant issue or specification
* [ ] Added a changelog entry in `.changelog` (for details, see [contributing guidelines](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#changelog))
* [x] Included comments for [documenting Go code](https://blog.golang.org/godoc)
* [x] Updated the relevant documentation or specification

### Reviewers Checklist

I have...

### DIFF
--- a/app/upgrades/v23/upgrades.go
+++ b/app/upgrades/v23/upgrades.go
@@ -2,6 +2,8 @@ package v23
 
 import (
 	"context"
+	"github.com/cosmos/ibc-go/v10/modules/core/02-client/types"
+	ibctmtypes "github.com/cosmos/ibc-go/v10/modules/light-clients/07-tendermint"
 
 	ibcwasmtypes "github.com/cosmos/ibc-go/modules/light-clients/08-wasm/types"
 	clientkeeper "github.com/cosmos/ibc-go/v10/modules/core/02-client/keeper"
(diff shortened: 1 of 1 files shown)


app/upgrades/v23/upgrades.go
```diff
@@ -30,17 +30,12 @@ func CreateUpgradeHandler(
 			return vm, errorsmod.Wrapf(err, "running module migrations")
 		}
 
-		// Add the Wasm client type to the allowed clients
-		Add08WasmToAllowedClients(ctx, *keepers.IBCKeeper.ClientKeeper)
+		// Set IBC Client AllowedClients
+		params := keepers.IBCKeeper.ClientKeeper.GetParams(ctx)
+		params.AllowedClients = []string{ibctmtypes.ModuleName, ibcwasmtypes.ModuleName}
+		keepers.IBCKeeper.ClientKeeper.SetParams(ctx, params)
```
Eric-Warehime, line 36: Why not set this to Allow All?
````
params.AllowedClients = []string{clienttypes.AllowAllClients}

vladjdk: The chain default is ["*"] but current mainnet config is set to only ["07-tendermint"]. I don't want to change it off of the params mainnet already has

Eric-Warehime: Yeah, my question is basically, is there a risk to allowing all versus specifically calling out the wasm and tm client types only.

gjermundgaraba: Shouldn't really be. Most chains run with "*" I believe, but having them like this is the more conservative and controlled option. I would perhaps consider having the list of defined light clients somewhere else so you could have this set up in genesis or tests too in that case.

vladjdk: I can't think of a risk, given the fact that the only way to introduce a new client type is via a new binary.

However, the `gaiad query ibc client params` query is clearer when it shows the type of clients that are allowed in case someone needs that information.

Either way, I think in general less permissive params are better in case there is a risk that we're not aware of

wllmshao: what do we think of making this conditional?
if the existing allowed clients is *, do nothing
otherwise append wasm to allowed clients?

Eric-Warehime: Yeah I'm fine with this then. We'll probably want to refactor this out into a constant somewhere for testing as Gjermund said, but we can do that as the need arises.

Eric-Warehime (approved):

app/upgrades/v23/upgrades.go
Eric-Warehime: > what do we think of making this conditional?
if the existing allowed clients is *, do nothing
otherwise append wasm to allowed clients?

This would result in some networks running this code to have different values than others. Which I believe is not desirable.

wllmshao (approved):

Issue #3646, opened 2025-03-14
Consistency of liquid share tracking


Eric-Warehime:
- [`TokenizeShares`](https://github.com/cosmos/gaia/blob/f46b6b058caf9b9b2d8780e59e5f7ad30996f5b3/x/liquid/keeper/msg_server.go#L132-L138)
- [`RedeemTokensForShares`](https://github.com/cosmos/gaia/blob/f46b6b058caf9b9b2d8780e59e5f7ad30996f5b3/x/liquid/keeper/msg_server.go#L289-L297)
In the `TokenizeShares` function, `SafelyIncreaseTotalLiquidStakedTokens` and `SafelyIncreaseValidatorLiquidShares` are called for all delegations.
However, in `RedeemTokensForShares`, `DecreaseTotalLiquidStakedTokens` and `DecreaseValidatorLiquidShares` are only called when the delegator is an ICA.
For consistency, I believe `RedeemTokensForShares` should also call `DecreaseTotalLiquidStakedTokens` and `DecreaseValidatorLiquidShares` for all delegations.

PR #3668, opened 2025-03-25
feat: Uniformly treat liquid tokens

Eric-Warehime:
Fixes the handling of liquid staking token accounting by removing a condition in which we fail to decrement the total liquid staked tokens in the case when the delegator is an ICA or module account.

Adds e2e testing for both the global and validator specific tokenization limits.

Closes: #3646


tests/e2e/e2e_exec_test.go
```diff
@@ -809,6 +809,36 @@ func (s *IntegrationTestSuite) executeValidatorBond(c *chain, valIdx int, valOpe
...
+		fmt.Sprintf("--%s=%s", flags.FlagGasPrices, delegateFees),
+		fmt.Sprintf("--%s=%d", flags.FlagGas, 1000000),
+		"--keyring-backend=test",
+		fmt.Sprintf("--%s=%s", flags.FlagHome, home),
+		"--output=json",
+		"-y",
+	}
+
+	s.executeGaiaTxCommand(ctx, c, gaiaCommand, valIdx, s.expectErrExecValidation(c, valIdx, true))
+	s.T().Logf("%s expected failure on execution of tokenize share tx from %s", delegatorAddr, valOperAddress)
```
vladjdk, line 839: Feels kinda weird to have the same logic duplicated for a failure case. Maybe we can deduplicate and refactor out the success messages from every execute message. Imo they should be agnostic to whether they succeed or fail

vladjdk (approved):

tests/e2e/e2e_exec_test.go
Eric-Warehime: Yeah I agree, it's not super clean. Let me take a look at simplifying this when I merge in the e2e test refactors from main.

PR #3695, opened 2025-04-08
fix: IBC v1 transfer stack (ics4wrapper)

### LINKED ISSUE #3694: [Bug]: IBC transfer stack is set up incorrectly
The transfer stack is currently not set up as is described. This is primarily due to a pointer being set to a new variable and modified incorrectly:
https://github.com/cosmos/gaia/blob/25707a9914a520bcb2e6010803b37ac8ea75c480/app/keepers/keepers.go#L530

The solution is an analysis of the stack and wiring it up in the expected order using a separate stack variable.

gjermundgaraba:
Fixes the transfer stack to the correct order.

Closes: #3694


vladjdk:
needs a linting fix

vladjdk (approved):

gjermundgaraba:
Need to investigate this error

Eric-Warehime:
Test failure is happening because rate limits are not being enforced. It seems like the changes made here to the transfer stack break rate limiting, but I'm not sure why yet.

PR #3713, opened 2025-04-16
feat: lsm migration

aljo242:
Reuse some logic from the `genesis` initialization
Go back to using SDK fork so we can get the types we need

Please go the `Preview` tab and select the appropriate sub-template:

* [Production code](?expand=1&template=production.md) - for types `fix`, `feat`, and `refactor`.
* [Docs](?expand=1&template=docs.md) - for documentation changes.
* [Others](?expand=1&template=others.md) - for changes that do not affect production code.

### DIFF
--- a/app/upgrades/v24/upgrades.go
+++ b/app/upgrades/v24/upgrades.go
@@ -2,11 +2,9 @@ package v24
 
 import (
 	"context"
-	"fmt"
-
 	errorsmod "cosmossdk.io/errors"
 	upgradetypes "cosmossdk.io/x/upgrade/types"
-
+	"fmt"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
@@ -112,14 +110,15 @@ func migrateTotalLiquidStakedTokens(ctx sdk.Context, sk *stakingkeeper.Keeper, l
 
 func migrateTokenizeShareLocks(ctx sdk.Context, sk *stakingkeeper.Keeper, lsmk *liquidkeeper.Keeper) error {
 	tokenizeShareLocks := sk.GetAllTokenizeSharesLocks(ctx)
-	for _, lock := range tokenizeShareLocks {
-		accAddr, err := sdk.AccAddressFromBech32(lock.Address)
-		if err != nil {
-			return err
+	converted := make([]liquidtypes.TokenizeShareLock, len(tokenizeShareLocks))
+	for i, tokenizeShareLock := range tokenizeShareLocks {
+		converted[i] = liquidtypes.TokenizeShareLock{
+			Address:        tokenizeShareLock.Address,
+			Status:         tokenizeShareLock.Status,
+			CompletionTime: tokenizeShareLock.CompletionTime,
 		}
-		lsmk.AddTokenizeSharesLock(ctx, accAddr)
-		lsmk.SetTokenizeSharesUnlockTime(ctx, accAddr, lock.CompletionTime)
 	}
+	lsmk.SetTokenizeShareLocks(ctx, converted)
 
 	return nil
 }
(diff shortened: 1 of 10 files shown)


x/liquid/keeper/genesis.go
```diff
@@ -33,7 +33,14 @@ func (k Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) {
...
 	// If it is in status LOCK_EXPIRING, the unlocking must also be queued
-	for _, tokenizeShareLock := range data.TokenizeShareLocks {
+	k.SetTotalLiquidStakedTokens(ctx, data.TotalLiquidStakedTokens)
+}
+
+func (k Keeper) SetTokenizeShareLocks(ctx sdk.Context, tokenizeShareLocks []types.TokenizeShareLock) {
+	// Set the tokenize shares locks for accounts that have disabled tokenizing shares
+	// The lock can either be in status LOCKED or LOCK_EXPIRING
+	// If it is in status LOCK_EXPIRING, the unlocking must also be queued
+	for _, tokenizeShareLock := range tokenizeShareLocks {
```
Eric-Warehime: This doesn't look right to me.

The `SetTotalLiquidStakedTokens` call in `InitGenesis` is repeated and now the tokenize share locks are not set in init genesis.

app/upgrades/v24/upgrades.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+func migrateLastTokenizeShareRecordID(ctx sdk.Context, sk *stakingkeeper.Keeper, lsmk *liquidkeeper.Keeper) {
+	lastTokenizeShareRecordID := sk.GetLastTokenizeShareRecordID(ctx)
+	lsmk.SetLastTokenizeShareRecordID(ctx, lastTokenizeShareRecordID)
+}
+
+func migrateTotalLiquidStakedTokens(ctx sdk.Context, sk *stakingkeeper.Keeper, lsmk *liquidkeeper.Keeper) {
+	totalLiquidStaked := sk.GetTotalLiquidStakedTokens(ctx)
```
Eric-Warehime: I'm not sure if those will be the same. The existing LSM will track stake from an ICA as liquid staked, but the intention is to just track the total stake tokenized via the x/liquid module going forward.

go.mod
```diff
@@ -275,15 +275,20 @@ require (
...
 
 replace (
+	// Use version with lsm changes
+	cosmossdk.io/api => github.com/informalsystems/cosmos-sdk/api v0.7.5-lsm
+
 	// Use cosmos keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
 
+	// Use special SDK v0.50.x release with support for both ICS and LSM
+	github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.50.11-lsm
```
arlai-mk: Curious if keeping the forked SDK here means LSM methods (e.g. `TokenizeShares`) from the forked `x/staking` module are still active after the migration?
If yes, there could be a risk that some clients continue calling those instead of the methods from the new `x/liquid` module, leading to inconsistencies between both modules. I understand that in this case, LSM shares created under the `x/staking` module would then "disappear" once reverting to a non-forked version of the SDK.

**Side question:** Could there be a risk if a future upgrade to the canonical (SDK) `x/staking` module ends up using store keys in the `0x81`–`0x87` range, which are currently used for LSM-related state? I'm not too familiar with how store key reuse is handled, but if that might cause issues, would it make sense to add a comment in the canonical `x/staking` code to avoid reusing those keys?

aljo242: Yes, you are right about this.  We need the LSM fork to use the actual types and  methods, but we also need some kind of way to "block" the usage of these LSM methods from `x/staking` until we actually can do the de-fork

aljo242: And yes - I think we should make an issue for the SDK to comment to avoid those keys. I can handle this

x/liquid/keeper/genesis.go
aljo242: i called the wrong function oops

Eric-Warehime (approved):

PR #3715, opened 2025-04-27
docs: CHANGELOG.md updated 

gap-editor:


gap-editor:
@Eric-Warehime hi. could you explain is smth wrong here? maybe i can fix it. thank you!

PR #3723, opened 2025-05-02
feat!: Remove expedited gov prop restriction

### LINKED ISSUE #3722: Remove Expedited Prop Restrictions
Remove the restrictions on only being able to open expedited gov props for software upgrades and cancellations.

Eric-Warehime:
## Description

Removes the restriction on expedited gov props. This opens up expedited gov props to any message type. 

The rationale for this is that restrictions on this are completely arbitrary. The deposit and quorum for a proposal are used to gate against spam and if people want to burn a bunch of atom on opening up gov props that will receive no w/ veto, then we should welcome their donations.

Closes: #3722


vladjdk (approved):

vladjdk:
lgtm, just need to fix some lints

PR #3731, opened 2025-05-08
deps: Resolve dependabots

Eric-Warehime:
Mass resolution of dependabot updates ahead of the v24 RC.

Relies on #3732


vladjdk (approved):
lgtm

PR #3738, opened 2025-05-09
Use all-caps DEPENDENCIES header in Dependabot changelog action

### LINKED ISSUE #3734: [Docs]: Update dependabot changelog writer config for all caps
### Summary

Comment from a dependabot PR:

> Hi, maintainer of the [dependabot-changelog-writer](https://github.com/CramBL/dependabot-changelog-writer) action here.

> I noticed that you are using all caps for your changelog sections, but the action default for the section header is Dependencies. To match with your changelog layout you would need to set section-header: "DEPENDENCIES", then the entries will show up under the right section.

gap-editor:
Closes #3734

Configure `section-header: "DEPENDENCIES"` in the Dependabot changelog action so entries land under our all-caps DEPENDENCIES heading.


gap-editor:
@Eric-Warehime

PR #3769, opened 2025-05-30
chore: Temporarily disable linux builds in goreleaser

Eric-Warehime:
The musl.cc download ban from GH actions breaks the linxu build so for now we'll disable it and revert to the previous method in which linux builds are not done via GH actions and are uploaded to the release separately.


vladjdk (approved):

puneet2019:
is this fixed and can it be reenabled?

PR #3772, opened 2025-06-02
chore: Remove sonarcloud

Eric-Warehime:
Sonarcloud fails on every test run right now and isn't generally useful to us at this moment.

We should replace w/ codecov soon.


vladjdk (approved):
cyaaaaaa won't be missed :)

PR #3780, opened 2025-06-12
docs: update broken link

Keemosty12:
#### What has changed
- Updated the link for the list of usable endpoints in `docs/docs/governance/submitting.md` to point directly to the correct `README.md` file.

#### Why is this change necessary
- The previous link pointed to a directory section, which could be confusing. The new link points directly to the relevant `README.md` file, making it easier for users to find the information they need.

#### Additional notes
- No changes to production code.
- Only documentation has been updated for clarity and accuracy.
- before:
before
- after
after


aljo242 (approved):

Cordtus:
The chain-id is also incorrect here - that one was shut down some time ago.

PR #3788, opened 2025-06-14
feat: add preinstalls

gjermundgaraba:
Adds preinstalls to the local node.

Closes HUB-1171

The ones added here are:
- deterministic-deployment-proxy
- multicall3
- permit2
- safe-singleton-factory (which is the same bytecode as deterministic-deployment-proxy, btw)

I did not add the ENS contracts, because there are like a million of them, and I don't wanna. 

Now that I know the procedure, I can add this as either some genesis command, or just add this to a little helper script to modify genesis to add the preinstalls. Just let me know what you want.


Eric-Warehime:
>I can add this as either some genesis command, or just add this to a little helper script to modify genesis to add the preinstalls.

@gjermundgaraba It seems like some command in the EVM repo which exposes something like `tx evm add-preinstall deployedBytecode deploymentAddress`.
Ideally we can add that in a way which allows people to set these things in upgrade handlers as well as during genesis creation.

gjermundgaraba:
That makes sense. I did it programatically in the first iteration to test, so this seems doable. For the command, would you want it to deploy them to a running network for now, guarded by an authority or something, to enable on devnet?

PR #3805, opened 2025-06-18
feat: open telemetry

technicallyty:
Adds opt-out open telemetry support for gaia validator nodes.

Notable changes:
- adds an open telemetry client that forwards prometheus metrics to an otel collector. 
- forces the node to use SDK's telemetry. this cannot be turned off.
- adds a new module `x/telemetry` that reports to the otel client whether the running node is a validator or not. (note i opted to optimize this a bit by only running the check every 20 blocks. feel free to suggest a different value or scrapping it altogether)

if you want to view metrics yourself with a localnode:

**BEFORE YOU START CHANGE THE CONFIG IN telemetry/config.go**

rename LocalOtelConfig to DefaultOtelConfig, and comment out the original DefaultOtelConfig. This will make the node use the local infra.

```
make start-telemetry-server
make start-localnet-ci
```

then open grafana: `localhost:3000` (user: admin, pass: admin).


x/amiavalidator/module.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (m Module) RegisterGRPCGatewayRoutes(c client.Context, mux *runtime.ServeMux) {}
+
+func (m Module) IsOnePerModuleType() {}
+
+func (m Module) IsAppModule() {}
+
+func (m Module) PreBlock(ctx context.Context) (appmodule.ResponsePreBlock, error) {
+	if m.oc.Enabled() {
+		// TODO(technicallyty): maybe we can just update it every 20 blocks, so we don't have to query the val every block?
```
Eric-Warehime: Agree w/ this comment, but it looks like you've already implemented that below so let's just remove the comment.

x/amiavalidator/module.go
```diff
@@ -0,0 +1,65 @@
...
+		if sdk.UnwrapSDKContext(ctx).BlockHeight()%20 == 0 {
```
Eric-Warehime: Let's make 20  a const

go.mod
```diff
@@ -45,12 +46,27 @@ require (
...
 	github.com/stretchr/testify v1.10.0
 	github.com/vektra/mockery/v2 v2.53.4
+	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp v1.36.0
 	google.golang.org/genproto/googleapis/api v0.0.0-20250528174236-200df99c418a
 	google.golang.org/grpc v1.73.0
 	google.golang.org/protobuf v1.36.6
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require (
```
Eric-Warehime: nit: I don't like having a bunch of separate require blocks.

telemetry/config.go
```diff
@@ -0,0 +1,48 @@
...
+	CollectorEndpoint       string        `mapstructure:"collector-endpoint"`
+	CollectorMetricsURLPath string        `mapstructure:"collector-metrics-url-path"`
+	User                    string        `mapstructure:"user"`
+	Token                   string        `mapstructure:"token"`
+	PushInterval            time.Duration `mapstructure:"push-interval"`
+}
+
+var DefaultOtelConfig = OtelConfig{
+	Disable:                 false,
+	CollectorEndpoint:       "localhost:4318",
```
Eric-Warehime, line 27: This will need to change before we cut a release correct?

telemetry/otel.go
Eric-Warehime, line 1: I think some of the stuff in here could benefit from some unit tests. Mostly the histogram and summary logic where we're taking a big data set and splitting it into multiple entries w/ different attributes.

x/amiavalidator/module.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (m Module) RegisterInterfaces(registry types.InterfaceRegistry) {}
+
+func (m Module) RegisterGRPCGatewayRoutes(c client.Context, mux *runtime.ServeMux) {}
+
+func (m Module) IsOnePerModuleType() {}
+
+func (m Module) IsAppModule() {}
+
+func (m Module) PreBlock(ctx context.Context) (appmodule.ResponsePreBlock, error) {
```
Eric-Warehime: As trivial as this function is, I still think we should write a unit test for it.

cmd/gaiad/cmd/root.go
```diff
@@ -137,7 +138,20 @@ func NewRootCmd() *cobra.Command {
...
-			return server.InterceptConfigsPreRunHandler(cmd, customAppTemplate, customAppConfig, customCometConfig)
+			err = server.InterceptConfigsPreRunHandler(cmd, customAppTemplate, customAppConfig, customCometConfig)
+			if err != nil {
+				return err
+			}
+
+			// we force telemetry to be enabled here, otherwise open telemetry won't emit SDK metrics.
+			serverCtx := server.GetServerContextFromCmd(cmd)
+			serverCtx.Viper.Set("telemetry.enabled", true)
+			serverCtx.Viper.Set("telemetry.prometheus-retention-time", 60)
```
Eric-Warehime: Hard for me to tell if this is the case right now, but we should only force telemetry to be enabled if the opentel is not disabled. People should still be able to disable both of them if they choose to.

app/app.go
```diff
@@ -567,6 +604,48 @@ func (app *GaiaApp) AutoCliOpts() autocli.AppOptions {
...
+		P2P:             &tmcfg.P2PConfig{},
+		Mempool:         &tmcfg.MempoolConfig{},
+		StateSync:       &tmcfg.StateSyncConfig{},
+		BlockSync:       &tmcfg.BlockSyncConfig{},
+		Consensus:       &tmcfg.ConsensusConfig{},
+		Storage:         &tmcfg.StorageConfig{},
+		TxIndex:         &tmcfg.TxIndexConfig{},
+		Instrumentation: &tmcfg.InstrumentationConfig{},
+	}
+	cfg.SetRoot(homePath)
```
Eric-Warehime, line 620: Why are you setting all of these fields and then just deserializing from file anyways after this?

x/amiavalidator/module.go
technicallyty: [remove todo comment](https://github.com/cosmos/gaia/pull/3805/commits/ba225fc25522b042f399b636d72951d2e4faf4f0)

telemetry/config.go
technicallyty: yes

app/app.go
```diff
@@ -221,6 +248,7 @@ func NewGaiaApp(
 	app.mm.SetOrderPreBlockers(
 		upgradetypes.ModuleName,
 		authtypes.ModuleName,
+		amiavalidator.ModuleName,
```
aljo242: nit: can we just call this module `x/telemetry`

telemetry/config.go
```diff
@@ -0,0 +1,48 @@
...
+type OtelConfig struct {
+	Disable                 bool          `mapstructure:"disable"`
+	CollectorEndpoint       string        `mapstructure:"collector-endpoint"`
+	CollectorMetricsURLPath string        `mapstructure:"collector-metrics-url-path"`
+	User                    string        `mapstructure:"user"`
+	Token                   string        `mapstructure:"token"`
+	PushInterval            time.Duration `mapstructure:"push-interval"`
+}
+
+var DefaultOtelConfig = OtelConfig{
```
aljo242, line 14: Since most nodes will not be "updating" their configs using the template as they are just existing nodes that are updating their binary, will these default config values still work?

aljo242 (commented):
overall lgtm - agreed with @Eric-Warehime on needing more unit testing

telemetry/config.go
technicallyty: yep, this config is injected in `root.go` and is only updated if they set the values themselves in the toml

cmd/gaiad/cmd/root.go
technicallyty: updated so it only runs if otel is not explicitly disabled

x/amiavalidator/module.go
technicallyty: added

telemetry/otel.go
technicallyty: added

app/app.go
technicallyty: we get a nil deref if i dont do this

telemetry/otel.go
```diff
@@ -108,6 +108,10 @@ func (o *OtelClient) SetValidatorStatus(isVal bool) {
 	o.vi.IsValidator = isVal
 }
 
+func (o *OtelClient) IsValidator() bool {
+	return o.vi.IsValidator
+}
+
 func (o *OtelClient) GetValAddr() crypto.Address {
 	return o.vi.Address
```
aljo242, line 133: I feel like this should return an error if we are not a validator ?

telemetry/otel.go
```diff
@@ -38,19 +38,38 @@ type (
...
 		vi:  vi,
 	}
+
+	for _, opt := range opts {
+		opt(client)
+	}
+
+	if client.gatherer == nil {
+		client.gatherer = prometheus.DefaultGatherer
+	}
```
aljo242: I think it is cleaner to just set the default value above like: 

```
	client := &OtelClient{
		cfg: otelConfig,
		vi:  vi,
		gatherer: prometheus.DefaultGatherer,
	}
	```

File: telemetry/otel_test.go [hunk 12]
```diff
@@ -1 +1,479 @@
...
+	reg := prometheus.NewRegistry()
+	attrs := createTestAttributes()
+	// Create and register a test gauge metric
+	name := "test_metric_gauge"
+	help := "A test gauge metric"
+	testGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: name,
+		Help: help,
+	})
+	gaugeValue := 3.1415
```
aljo242, line 66: king

aljo242:
LGTM - just left a few more tiny nits and needs a lint fix

telemetry/otel.go
technicallyty: i don't think that really benefits the usage much. the address is always set because sdk/comet will always provision a node key. then its up to `x/telemetry` to use this address to determine whether we are bonded or not. and you can still have a val address without being bonded.

telemetry/otel.go
technicallyty: right - i usually do this too... brainfart 🙂

technicallyty:
address your comments in my latest commits @aljo242

technicallyty:
linter is yapping about deprecated code. don't think i will fix it here

aljo242:
@technicallyty lgtm lets ship it - before merge can you just add `//nolint` for the lint issues with  TODOs to resolve?

aljo242 (approved):

Eric-Warehime (approved):

aljo242:
@technicallyty we should add a changelog sorry

PR #3970, opened 2026-02-20
test(ante): remove unused fields from mockMessenger

Ocheretovich:
Remove unused fields from mockMessenger in WasmGovVoteDecorator tests. These fields were not referenced in tests and can be safely removed


dasanchez (approved):
Thanks @Ocheretovich !

PR #3998, opened 2026-03-12
!feat: Remove Interchain Security

dasanchez:
## Description

* Remove all wiring related to Interchain Security.
  * Block the consumer rewards pool from receiving any additional tokens.
* In the upgrade handler:
  * Set the `max_validators` param in the staking module to match the `max_provider_consensus_validators` param in the provider module.
  * Trim the B-group validator set to match: validators that were bonded in staking but ranked below `max_provider_consensus_validators` (i.e. excluded from the CometBFT consensus set by ICS, aka the "B group") are evicted from `LastValidatorPower`. Any of these still in `Bonded` status are moved to `Unbonding` via `BeginUnbondingValidator`, and their tokens are transferred from the bonded to the not-bonded pool. They are not jailed or slashed, this is the same transition normal staking-driven demotion below `max_validators` would cause. This step is required because otherwise `GetLastValidators` (called from `TrackHistoricalInfo` in `BeginBlocker`) would panic once `max_validators` is lowered below the existing number of `LastValidatorPower` entries.
  * Transfer the balances from the consumer rewards pool to the community pool account.
  * Close all `provider` port IBC channels.

---

### Author Checklist

I have...

* [x] Included the necessary unit and integration [tests](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#testing)
* [x] Added a changelog entry in `.changelog` (for details, see [contributing guidelines](https://github.com/cosmos/gaia/blob/main/CONTRIBUTING.md#changelog))

### Reviewers Checklist

I have...

* [x] confirmed `!` in the type prefix if API or client breaking change


app/upgrades/v28_0_0/upgrades.go
```diff
@@ -20,14 +29,213 @@ func CreateUpgradeHandler(
...
+		if err := proto.Unmarshal(paramsBz, &providerParams); err != nil {
+			return vm, fmt.Errorf("failed to unmarshal provider params: %w", err)
+		}
+		maxVals := providerParams.MaxProviderConsensusValidators
+		ctx.Logger().Info("Read provider max_provider_consensus_validators", "value", maxVals)
+
+		// Validate maxVals is a positive value within uint32 range.
+		// Rejecting zero prevents the handler from setting max_validators=0,
+		// which would unbond every validator and halt the chain.
+		if maxVals <= 0 || maxVals > math.MaxUint32 {
```
farkhod-clk: Hi @dasanchez, I've been curious on the process of sanity check of this, a schema drift that still decodes to some small positive `int64` (ex: a wire-type collision landing on an adjacent field) would silently pass and set `max_validators` to a wrong-but-plausible number. 

Not blocking, but worth a tighter bound or an explicit cross-check against `providertypes.Params{}.MaxProviderConsensusValidators` field tag in CI so upstream drift breaks the build loudly instead of at upgrade time.

CHANGELOG.md
farkhod-clk, line 1: @dasanchez 
`x/legacy/ics` stubs are honestly documented as **"Unmarshal is a no-op that discards bytes"** — after the upgrade, querying an old `MsgConsumerAddition/ConsumerAdditionProposal` etc. will return an empty stub instead of the real historical fields (`chain_id`, `spawn_time`, etc.). This is a reasonable tradeoff to avoid decode errors, but it's a query-API behavior change for indexers/explorers that isn't mentioned in the `CHANGELOG.md` entry (**"Remove Interchain Security provider module"** is the only line). Worth a changelog note, what do you think about?

app/upgrades/v28_0_0/constants.go
farkhod-clk, line 1: @dasanchez following up our conversation regarding the old provider KV store which is never pruned — `StoreUpgrades.Deleted` is missing. 
`constants.go`'s upgrade literal only sets `UpgradeName/CreateUpgradeHandler`
it never sets [StoreUpgrades](https://github.com/cosmos/gaia/blob/ad3637a29c0bd8e94318368269a0ecb7037f14e4/app/app.go#L517). (`setupUpgradeStoreLoaders`) shows this is exactly the mechanism the codebase uses to prune a removed module's store at the upgrade height (`upgradetypes.UpgradeStoreLoader`).

If deleted while still registered, the store may become permanently `unprunable` (the store-loader mechanism requires the key to exist in the binary at the moment of deletion).

CHANGELOG.md
dasanchez: Thanks, agreed! The changelog now has a note about this.

app/upgrades/v28_0_0/constants.go
dasanchez: @farkhod-clk the `provider` kv store is still registered in GenerateKeys so we can read the provider params and pending VSC packets in the upgrade handler: https://github.com/cosmos/gaia/blob/feat/remove-ics/app/keepers/keys.go#L58
This `provider` hard-coded string in key.go would be removed in a later upgrade. The tentative sequence is:
1. v28.0.0: remove ICS wiring, leave `provider` in keys.go
2. v28.1.0: delete store, leave `provider` in keys.go (see below)
3. v28.2.0 / v29.0.0: remove `provider` from keys.go

`upgrades/v28_1_0/constants.go` would have this:
```
var Upgrade = upgrades.Upgrade{
	UpgradeName:          UpgradeName,
	CreateUpgradeHandler: CreateUpgradeHandler,
	StoreUpgrades: store.StoreUpgrades{
		Deleted: []string{
			providerStoreKey,
		},
	},
```

farkhod-clk: @dasanchez looks good,  in this way It doesn't leave the store permanently unprunable, the key stays registered exactly through the release where the delete happens, per the rule as you've already cited.

app/upgrades/v28_0_0/upgrades.go
dasanchez: Thanks! The accepted range for `maxVals` is now [4,225].

dasanchez:
Interchain test run: https://github.com/cosmos/gaia/actions/runs/30565648374

danbryan:
@greptileai please review

farkhod-clk:
> Interchain test run: https://github.com/cosmos/gaia/actions/runs/30565648374

Hi @dasanchez, I see like above action's `test-report` has failed with
```
The specified go version file at: tests/interchain/go.mod does not exist
```

And I can also see that, you've already proposed solution in #4106 which resulted in successful run of Git action: https://github.com/cosmos/gaia/actions/runs/30571949442 from branch: `ci/fix-interchaintest-report`

Do you think it would be also better to merge above PR and re-run the `Interchain Test`?  I personally, don't see any benefit of it , cause it just reports the output~

dasanchez:
@farkhod-clk 
I merged https://github.com/cosmos/gaia/pull/4106 and re-ran the Interchain Test workflow, all good:
https://github.com/cosmos/gaia/actions/runs/30913001051

app/upgrades/v28_0_0/upgrades.go
```diff
@@ -20,14 +28,213 @@ func CreateUpgradeHandler(
...
+		}
+		maxVals := providerParams.MaxProviderConsensusValidators
+		ctx.Logger().Info("Read provider max_provider_consensus_validators", "value", maxVals)
+
+		// Validate maxVals is a positive value within uint32 range.
+		// Rejecting zero prevents the handler from setting max_validators=0,
+		// which would unbond every validator and halt the chain.
+		if maxVals <= 3 || maxVals > 225 {
+			return vm, fmt.Errorf("invalid max_provider_consensus_validators value: %d (must be between 4 and 225)", maxVals)
+		}
```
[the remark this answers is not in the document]
farkhod-clk: IMO, the bot is factually right about the code, but very likely wrong that it's a bug and real-world mainnet risk is effectively nil right now, because Hub's live `max_provider_consensus_validators` is currently `180` (the size of the active set), nowhere near 1–3. So under today's on-chain state, this branch will not fire during the actual `v28` upgrade. Given the mainnet value is `180` and no one is proposing to drop it into single digits before the upgrade, I think it's not a mainnet-upgrade blocker.

@dasanchez What do you think about this?

farkhod-clk (approved):
LGTM

jeonghwan-clk (approved):
LGTM!

PR #4141, opened 2026-09-18
chore: inherit org security policy

aljo242:
Removes this repo's `SECURITY.md` so it inherits the org default from `cosmos/.github`, which points at `cosmos/security`.

- Sends reporters to `hackerone.com/cosmos`. The live program is Immunefi.
- References `security@interchain.io`.

One policy instead of a copy per repo, so it can't go stale again.


farkhod-clk (commented):
The inherited policy at [cosmos/security](https://github.com/cosmos/security/blob/main/SECURITY.md) scopes reports to the SDK "release-family" list (CometBFT, SDK, Cosmos EVM, ibc-go, etc.), which doesn't include gaia — even though gaia is listed as in-scope on [Immunefi](https://immunefi.com/bug-bounty/cosmos/scope/). Won't this make researchers think gaia is out of scope for reporting? Should the org policy be updated to include gaia before we drop gaia's own [SECURITY.md](https://github.com/cosmos/gaia/blob/main/SECURITY.md) here?

